PR #61, opened 2024-04-02
Add HTS Field 'StorageType' while ensuring backward-compatibility.

HotSushi:
## Summary
This PR introduces a new HTS field called `storageType`, which keeps track of storage used for each table (ex: "hdfs" vs "blobfs"). Its necessary to have this metadata as HTS field rather than iceberg metadata.

We need to make this change backward compatibly to ensure services do not break, and HTS is the most critical service of all.

## Changes
- [X] New Features

## Ensuring backward compatibility
We need to ensure backward compatibility at following 4 places: 
image
1. GET call from Tables to HTS
2. GET row call from HTS to MySQL
3. PUT call from Tables to HTS
4. PUT row call from HTS to MySQL

#### For 1 & 2
we ensure backward compatibility by evolving mysql schema with a default value, `ie. DEFAULT 'hdfs' NOT NULL`. For existing rows, hdfs would be returned.

#### For 3 & 4
we ensure backward compatibility by allowing `null` in the requestBody's "storageType" field, and in the PUT request null is defaulted to 'hdfs'. This way non-evolved table-service's PUT call will be automatically treated as "hdfs" storageType. 
For 4, storageType will always be set to something because the mapping of null -> 'hdfs' happens in the service layer.
 
## Deployment strategy
1. ALTER SCHEMA on MySQL instances
```
ALTER TABLE user_table_row
ADD COLUMN storage_type VARCHAR(128) DEFAULT 'hdfs' NOT NULL;
```
2. Deploy HTS service

We'll run 1 & 2 on staging first and then repeat them on all other clusters.

(eventually) 3: Deploy Table Service which fills "hdfs" storage type
(eventually) 4: Remove default value from HTS service.

## Testing Done
✅  added unit tests and existing e2e tests work 
✅ ran e2e docker tests [HTS Field "StorageType" Docker tests](https://docs.google.com/document/d/18fTgTavQO2kPTmdHqNLAIBQyQP86MLEsUsiUl0PffDk/edit?usp=sharing)
✅ ran e2e docker tests for backward compatibility scenarios [HTS Field "StorageType" Docker tests](https://docs.google.com/document/d/18fTgTavQO2kPTmdHqNLAIBQyQP86MLEsUsiUl0PffDk/edit?usp=sharing)

### DIFF
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/dto/mapper/UserTablesMapper.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/dto/mapper/UserTablesMapper.java
@@ -26,11 +26,15 @@ import org.mapstruct.Mapping;
     uses = {UserTableVersionMapper.class})
 public interface UserTablesMapper {
 
+  String DEFAULT_STORAGE = "hdfs";
+
   /**
    * From a source {@link UserTable}, prepare a {@link UserTableRow} object which is compliant with
    * the storage {@link
    * com.linkedin.openhouse.housetables.repository.impl.iceberg.UserTableHtsRepository},
    * particularly wrt its {@link Long} version requirements.
+   *
+   * <p>Also sets the default storageType as "hdfs" if not present.
    *
    * @param userTable userTable source to be mapped
    * @param existingUserTableRow if not present, set "version" as 1. If present, set "version" as
@@ -39,6 +43,7 @@ public interface UserTablesMapper {
    * @return Destination House Table Data Model for storing Table Metadata.
    */
   @Mapping(target = "version", source = "userTable", qualifiedByName = "toVersion")
+  @Mapping(target = "storageType", source = "userTable.storageType", defaultValue = DEFAULT_STORAGE)
   UserTableRow toUserTableRow(
       UserTable userTable, @Context Optional<UserTableRow> existingUserTableRow);
 
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/model/UserTable.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/model/UserTable.java
@@ -48,6 +48,13 @@ public class UserTable {
   @NotEmpty(message = "metadataLocation cannot be empty")
   private String metadataLocation;
 
+  @Schema(
+      description = "Storage type to be used for the table.",
+      example = "hdfs",
+      defaultValue = "hdfs")
+  @JsonProperty(value = "storageType")
+  private String storageType;
+
   public String toJson() {
     return new Gson().toJson(this);
   }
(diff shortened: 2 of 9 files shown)


services/housetables/src/main/java/com/linkedin/openhouse/housetables/dto/mapper/UserTablesMapper.java
```diff
@@ -26,19 +26,24 @@
...
+   *
    * @param userTable userTable source to be mapped
    * @param existingUserTableRow if not present, set "version" as 1. If present, set "version" as
    *     the last saved version as long as: existingUserTableRow's metadata location and userTable's
    *     version match.
    * @return Destination House Table Data Model for storing Table Metadata.
    */
   @Mapping(target = "version", source = "userTable", qualifiedByName = "toVersion")
+  @Mapping(target = "storageType", source = "userTable.storageType", defaultValue = DEFAULT_STORAGE)
   UserTableRow toUserTableRow(
```
HotSushi, line 47: ### null -> 'hdfs' mapping happens here!

Another option was to do this by setting default in `api/spec/model/UserTable.java` using annotations such as `@DefaultValue`. But this wouldn't work cleanly with `/hts/tables/query` query, where null has a special semantics (null means `*` in that search api, which would conflict with put where null would mean `hdfs`)

services/housetables/src/main/java/com/linkedin/openhouse/housetables/dto/mapper/UserTablesMapper.java
```diff
@@ -26,19 +26,24 @@
     uses = {UserTableVersionMapper.class})
 public interface UserTablesMapper {
 
+  String DEFAULT_STORAGE = "hdfs";
+
```
HotSushi, line 30: considered keeping this constant in `:cluster:storage` but this hardcoding is short-lived until tables service is updated. Therefore keeping it as a local constant.

sumedhsakdeo:
>> (eventually) 3: Deploy Table Service which fills "hdfs" storage type

If a table is never rewritten again, would this eventual catchup never happen? Is there plan to do a noop table update or backfill mysql, in a way that we dont need to rely on external table writers to update this property.

HotSushi:
> If a table is never rewritten again, would this eventual catchup never happen? Is there plan to do a noop table update or backfill mysql, in a way that we dont need to rely on external table writers to update this property.

There's no need to catch up. Rows inserted meanwhile will be in [a good state](https://docs.google.com/document/d/18fTgTavQO2kPTmdHqNLAIBQyQP86MLEsUsiUl0PffDk/edit#heading=h.vb5xqkao6gp8). The only reason for `3: Deploy Table Service which fills "hdfs" storage type` is so that we can remove the hardcoding and mapping of null -> 'hdfs'

sumedhsakdeo (approved):
Nice clean PR. And really thorough testing to make sure change is backward compatible.

teamurko (approved):

PR #76, opened 2024-04-18
Add cluster.yaml, StorageProperties, StorageType enum for supporting multiple storages

HotSushi:
## Summary
This PR along with series of future PR will lay a foundation for adding multiple storages.

- `cluster-test-properties.yaml` shows how the yaml would look like post multi-storage support
```
cluster:
storages:
  default-type: hdfs
  types:
    hdfs:                                       # <----- this StorageType will be stored in HTS
      scheme: hdfs
      root-path: /data/openhouse
      endpoint: hdfs://namenode:9000/
      parameters:
        stringK1: stringV1
        token-refresh-enabled: true

    s3:
      scheme: s3
      root-path: bucketName
      endpoint: http://minio:9000
      parameters:
        stringK2: stringV2

    blob:
      scheme: blobfs
      root-path:
      endpoint: 127.0.0.1:22814
      parameters:
        tenant-name: e2eTestTenant1

```
- `StorageType` imitates an enum and should contain all storages that are supported by OH.
```
StorageType {
  HDFS
  LOCAL
}
```

## Details
`StorageType` is designed in a way that allows it to be extended to add more Storages. Please see the test `StorageTypeEnumTest`

If cluster.yaml is not updated with new layout, `StorageProperties` will be null.

For more details see this doc: https://docs.google.com/document/d/1iPKol1jPzR6fq2YaoWbHHCYzBciCwfhqZqGyYhzZfoQ/edit?usp=sharing

## Changes

- [X] New Features
-- Mult-storage support

## Testing Done
✅ Added unit tests
✅ Build succeeds
✅ springboot service starts even if cluster.yaml is not updated. (Backward-compatible)

### DIFF
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageType.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageType.java
@@ -1,0 +1,34 @@
+package com.linkedin.openhouse.cluster.storage;
+
+import lombok.*;
+
+/**
+ * Enum for supported storage types.
+ *
+ * <p>New types should be added here as public static final fields, and their corresponding
+ * implementations should be added to the fromString method.
+ */
+public class StorageType {
+  public static final Type HDFS = new Type("hdfs");
+  public static final Type LOCAL = new Type("local");
+
+  @AllArgsConstructor
+  @EqualsAndHashCode
+  @Getter
+  public static class Type {
+    private String value;
+  }
+
+  public Type fromString(String type) {
+    if (type == null) {
+      return null;
+    }
+    if (type.equals(HDFS.getValue())) {
+      return HDFS;
+    } else if (type.equals(LOCAL.getValue())) {
+      return LOCAL;
+    } else {
+      return null;
+    }
+  }
+}
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageType.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageType.java
@@ -20,15 +20,12 @@ public class StorageType {
   }
 
   public Type fromString(String type) {
-    if (type == null) {
-      return null;
-    }
-    if (type.equals(HDFS.getValue())) {
+    if (HDFS.getValue().equals(type)) {
       return HDFS;
-    } else if (type.equals(LOCAL.getValue())) {
+    } else if (LOCAL.getValue().equals(type)) {
       return LOCAL;
     } else {
-      return null;
+      throw new IllegalArgumentException("Unknown storage type: " + type);
     }
   }
 }
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
@@ -1,0 +1,35 @@
+package com.linkedin.openhouse.cluster.storage.configs;
+
+import com.linkedin.openhouse.cluster.configs.YamlPropertySourceFactory;
+import java.util.HashMap;
+import java.util.Map;
+import lombok.*;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.PropertySource;
+
+@Configuration
+@ConfigurationProperties(prefix = "cluster.storages")
+@PropertySource(
+    name = "clusterStorage",
+    value = "file:${OPENHOUSE_CLUSTER_CONFIG_PATH:/var/config/cluster.yaml}",
+    factory = YamlPropertySourceFactory.class,
+    ignoreResourceNotFound = true)
+@Getter
+@Setter
+public class StorageProperties {
+  private String defaultType;
+  private Map<String, StorageTypeProperties> types = new HashMap<>();
+
+  @Getter
+  @Setter
+  @AllArgsConstructor
+  @NoArgsConstructor
+  @Builder(toBuilder = true)
+  public static class StorageTypeProperties {
+    private String scheme;
+    private String rootPath;
+    private String endpoint;
+    @Builder.Default private Map<String, String> parameters = new HashMap<>();
+  }
+}
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
@@ -19,7 +19,7 @@ import org.springframework.context.annotation.PropertySource;
 @Setter
 public class StorageProperties {
   private String defaultType;
-  private Map<String, StorageTypeProperties> types = new HashMap<>();
+  private Map<String, StorageTypeProperties> types;
 
   @Getter
   @Setter
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
@@ -10,8 +10,8 @@ import org.springframework.context.annotation.PropertySource;
 
 /**
  * This class represents the storage properties for the cluster. It includes the default storage
- * type and a map of different storage types. Each storage type has its own properties such as
- * scheme, root path, endpoint, and parameters. For list of supported storage types, see {@link
+ * type and a map of different storage types. Each storage type has its own properties such as root
+ * path, endpoint, and parameters. For list of supported storage types, see {@link
  * com.linkedin.openhouse.cluster.storage.StorageType}.
  */
 @Configuration
@@ -33,7 +33,6 @@ public class StorageProperties {
   @NoArgsConstructor
   @Builder(toBuilder = true)
   public static class StorageTypeProperties {
-    private String scheme;
     private String rootPath;
     private String endpoint;
     @Builder.Default private Map<String, String> parameters = new HashMap<>();
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
@@ -8,6 +8,12 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.PropertySource;
 
+/**
+ * This class represents the storage properties for the cluster. It includes the default storage
+ * type and a map of different storage types. Each storage type has its own properties such as
+ * scheme, root path, endpoint, and parameters. For list of supported storage types, see {@link
+ * com.linkedin.openhouse.cluster.storage.StorageType}.
+ */
 @Configuration
 @ConfigurationProperties(prefix = "cluster.storages")
 @PropertySource(
--- a/services/tables/src/test/java/com/linkedin/openhouse/tables/mock/storage/StoragePropertiesConfigTest.java
+++ b/services/tables/src/test/java/com/linkedin/openhouse/tables/mock/storage/StoragePropertiesConfigTest.java
@@ -16,11 +16,11 @@ public class StoragePropertiesConfigTest {
 
   private static final String DEFAULT_TYPE = "hdfs";
 
-  private static final String DEFAULT_SCHEME = "hdfs";
+  private static final String DEFAULT_ENDPOINT = "hdfs://localhost:9000";
 
   private static final String ANOTHER_TYPE = "objectstore";
 
-  private static final String ANOTHER_SCHEME = "objectstorescheme";
+  private static final String ANOTHER_ENDPOINT = "http://localhost:9000";
   private static final String NON_EXISTING_TYPE = "non-existing-type";
 
   @Test
@@ -31,13 +31,13 @@ public class StoragePropertiesConfigTest {
   @Test
   public void testStorageTypeScheme() {
     Assertions.assertEquals(
-        DEFAULT_SCHEME, storageProperties.getTypes().get(DEFAULT_TYPE).getScheme());
+        DEFAULT_ENDPOINT, storageProperties.getTypes().get(DEFAULT_TYPE).getEndpoint());
   }
 
   @Test
   public void testStorageTypeLookup() {
     Assertions.assertEquals(
-        ANOTHER_SCHEME, storageProperties.getTypes().get(ANOTHER_TYPE).getScheme());
+        ANOTHER_ENDPOINT, storageProperties.getTypes().get(ANOTHER_TYPE).getEndpoint());
   }
 
   @Test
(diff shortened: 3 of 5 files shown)


services/tables/src/test/resources/cluster-test-properties.yaml
```diff
@@ -2,6 +2,21 @@ cluster:
...
+        rootpath: "/tmp/unittest"
+        endpoint: "hdfs://localhost:9000"
+        parameters:
+          key1: value1
+      s3:
+        scheme: "s3"
+        rootpath: "tmpbucket"
+        endpoint: "http://localhost:9000"
+        parameters:
+          key2: value2
```
autumnust, line 19: why aren't we not having a `local` segment here instead?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageType.java
```diff
@@ -0,0 +1,34 @@
...
+  public Type fromString(String type) {
+    if (type == null) {
+      return null;
+    }
+    if (type.equals(HDFS.getValue())) {
+      return HDFS;
+    } else if (type.equals(LOCAL.getValue())) {
+      return LOCAL;
+    } else {
+      return null;
```
autumnust, line 31: why is this not throwing exception instead? I assume downstream will have hard time dealing with null like this.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.PropertySource;
+
+@Configuration
+@ConfigurationProperties(prefix = "cluster.storages")
+@PropertySource(
+    name = "clusterStorage",
+    value = "file:${OPENHOUSE_CLUSTER_CONFIG_PATH:/var/config/cluster.yaml}",
+    factory = YamlPropertySourceFactory.class,
+    ignoreResourceNotFound = true)
```
autumnust, line 17: is this a safe config?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
```diff
@@ -0,0 +1,35 @@
...
+@Getter
+@Setter
+public class StorageProperties {
+  private String defaultType;
+  private Map<String, StorageTypeProperties> types = new HashMap<>();
```
autumnust, line 22: it is actually not common to instantiate field variable outside of ctor especially if it is not a static. curious why did we decide to do so, instead of using build or ctor?

autumnust (changes requested):

sumedhsakdeo:
+ @jainlavina

services/tables/src/test/resources/cluster-test-properties.yaml
HotSushi: This `cluster.yaml` is just for testing. 

As for introducing a local or not, I wanted to keep it optional for backward-compatibility with tables-test-fixtures. ie. If no storages are provided, Local would be used by default.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageType.java
HotSushi: ack! this should behave exactly like enum. And enum throws exception.. will address!

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
HotSushi: what aspect of safe were you thinking of? its safe in the sense, if properties are not provided, spring-boot will not fail and storageProperties will be null.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
HotSushi: following [previous code practice](https://github.com/linkedin/openhouse/blob/main/services/jobs/src/main/java/com/linkedin/openhouse/jobs/config/JobsProperties.java), its not necessary, will address.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
```diff
@@ -0,0 +1,35 @@
...
+@Configuration
+@ConfigurationProperties(prefix = "cluster.storages")
+@PropertySource(
+    name = "clusterStorage",
+    value = "file:${OPENHOUSE_CLUSTER_CONFIG_PATH:/var/config/cluster.yaml}",
+    factory = YamlPropertySourceFactory.class,
+    ignoreResourceNotFound = true)
+@Getter
+@Setter
+public class StorageProperties {
```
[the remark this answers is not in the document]
HotSushi: Added

services/tables/src/test/resources/cluster-test-properties.yaml
```diff
@@ -2,6 +2,21 @@ cluster:
   name: "TestCluster"
   storage:
     root-path: "/tmp/unittest"
+  storages:
+    default-type: "hdfs"
+    types:
+      hdfs:
```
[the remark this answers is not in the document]
HotSushi: done

services/tables/src/test/java/com/linkedin/openhouse/tables/mock/storage/StoragePropertiesConfigTest.java
```diff
@@ -0,0 +1,53 @@
...
+  private static final String ANOTHER_TYPE = "s3";
+  private static final String NON_EXISTING_TYPE = "non-existing-type";
+
+  @Test
+  public void testDefaultType() {
+    Assertions.assertEquals(DEFAULT_TYPE, storageProperties.getDefaultType());
+  }
+
+  @Test
+  public void testStorageTypeScheme() {
```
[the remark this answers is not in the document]
HotSushi: good point, Addressed!

services/tables/src/test/java/com/linkedin/openhouse/tables/mock/storage/StoragePropertiesConfigTest.java
```diff
@@ -0,0 +1,53 @@
...
+    Assertions.assertEquals(
+        DEFAULT_TYPE, storageProperties.getTypes().get(DEFAULT_TYPE).getScheme());
```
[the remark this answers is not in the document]
HotSushi: addressed

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageType.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Enum for supported storage types.
+ *
+ * <p>New types should be added here as public static final fields, and their corresponding
+ * implementations should be added to the fromString method.
+ */
+public class StorageType {
+  public static final Type HDFS = new Type("hdfs");
+  public static final Type LOCAL = new Type("local");
```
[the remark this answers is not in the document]
HotSushi: single node cluster with storage on local host, 

this has been the behavior for OH server if no storage configuration is provided. Local also gets used in `:table-test-fixtures` the test framework OH provides.

jainlavina (dismissed):

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
autumnust: that's exactly the thing I am thinking. Should we allow application to boot if storageProperties is null ?

autumnust (changes requested):
just having one pending question on the default behavior for the application when properties is missing

services/tables/src/test/resources/cluster-test-properties.yaml
```diff
@@ -2,6 +2,21 @@ cluster:
...
+        scheme: "hdfs"
```
sumedhsakdeo, line 9:
```suggestion
        type: "hdfs"
```
scheme has a very specific meaning. cloud blob stores like s3 can support an endpoint like 
`s3:///` where scheme is s3, and `https://`

I think we should rename scheme to type is better suited for this. I am sure same with GCS Wdyt?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
```diff
@@ -0,0 +1,41 @@
...
+  private String defaultType;
+  private Map<String, StorageTypeProperties> types;
```
sumedhsakdeo, line 28: We need a validation method here, given there could be misconfigurations or typo. We could start with some basic validations, for example, we could make sure if the storage type is hdfs, then the endpoint startsWith("hdfs:///"), and so on and so forth.

services/tables/src/test/resources/cluster-test-properties.yaml
```diff
@@ -2,6 +2,21 @@ cluster:
...
+        rootpath: "/tmp/unittest"
+        endpoint: "hdfs://localhost:9000"
+        parameters:
+          key1: value1
+      objectstore:
+        scheme: "objectstorescheme"
+        rootpath: "tmpbucket"
+        endpoint: "http://localhost:9000"
+        parameters:
+          key2: value2
```
sumedhsakdeo, line 19:
```suggestion
          token: xyz
```
Maybe we can add some keys we expect.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
HotSushi: Got it! I'd like to follow previous pattern of `if not provided null/default` and let the beans being initialized fail with useful error message.

services/tables/src/test/resources/cluster-test-properties.yaml
HotSushi: this property was targeting exactly that, the scheme. The goal is to prefix tablelocation with this scheme instead of using `fs.makeQualified()`, if s3 allows multiple schemes, the OH server will select one of them. let me know if we should adapt this. cc: @jainlavina  

Type information is already collected, in `StorageProperties.getTypes().getKeySet()`

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
HotSushi: Good point! we'll add it when these properties get used to construct beans. The goal of this property is simply to be a POJO for yaml confs. If they are misconfigured, we can make a decision when they get used.

services/tables/src/test/resources/cluster-test-properties.yaml
sumedhsakdeo: That's true, we have the type information available `StorageProperties.getTypes().getKeySet()`

One more question, what is the value on scheme here. If you look at s3 itself, there are various ways in which users could author this:

- https://s3.amazonaws.com/[bucket-name]/[object-key]
- - scheme: ?
- - endpoint: https://s3.amazonaws.com/
- - rootpath: bucket-name

- https://[bucket-name].s3.amazonaws.com/[object-key]
- - scheme: ?
- - endpoint: https://[bucket-name].s3.amazonaws.com/
- - rootpath: ''

- https://[bucket-name].s3-[region].amazonaws.com/[object-key]
- - scheme: ?
- - endpoint: https://[bucket-name].s3-[region].amazonaws.com
- - rootpath: ''

- https://[access-point-name]-[account-id].s3-accesspoint.[region].amazonaws.com/[object-key]
- - scheme: ?
- - endpoint: https://[access-point-name]-[account-id].s3-accesspoint.[region].amazonaws.com/
- - rootpath: ''

- s3://[bucket-name]/[object-key]
- - scheme: ?
- - endpoint: s3://
- - rootpath: bucket-name

In Hadoop Filesystem client, used against HDFS
- hdfs:///<cluster-name>/[dir-name]/[object-key]
- - scheme: ?
- - endpoint: hdfs:///<cluster-name>
- - rootpath: bucket-name

In Hadoop Filesystem client, used against Local File System
- file:///[dir-name]/[object-key]
- - scheme: ?
- - endpoint: file:///
- - rootpath: dir-name

In Hadoop Filesystem client, used against S3
- s3a:///[bucket-name]/[object-key]
- - scheme: ?
- - endpoint: s3a:///
- - rootpath: bucket-name

;tldr it is seems like endpoint could encompass scheme in pretty much all the cases. Maybe we don't need scheme, because endpoint + rootpath is sufficient?

HotSushi: great input Sumedh! and thanks for putting together various scenarios, very helpful.

I have an example where endpoint + rootpath is not sufficient and [thats for minio](https://tabular.io/blog/minio/):
there, warehouse link is: `s3://warehouse` and endpoint is `http://minio:9000`.

how do you feel about replacing scheme with `schemePrefix`, and we'll construct tableLocation by blindly adding schemePrefix+locationPath. It'll give us flexibility. 

Few Examples:

1. https://[bucket-name].s3-[region].amazonaws.com/[object-key]
schemePrefix=https://bucket-name.s3-region.amazonaws.com/
locationPath=object-key

2. s3://[bucket-name]/[object-key]
schemePrefix=s3://bucket-name/
locationPath=object-key
or 
schemePrefix=s3://
locationPath=bucket-name/object-key

3. hdfs:///[dir-name]/[object-key]
schemePrefix=hdfs:///
locationPath=bucket-name/object-key

sumedhsakdeo: >> I have an example where endpoint + rootpath is not sufficient and [thats for minio](https://tabular.io/blog/minio/):
there, warehouse link is: s3://warehouse and endpoint is http://minio:9000.

Given we will be using S3 client with MinIO,  the scheme part seems redundant and specific to how the demo catalog decided to configures it.
```
s3_client = boto3.client('s3',
                         endpoint_url=minio_endpoint,
                         aws_access_key_id=access_key,
                         aws_secret_access_key=secret_key)
s3_client.create_bucket(Bucket='my-new-bucket')
```

In our case, we would use following for minIO
```
rootPath=warehouse
endpoint=http://minio:9000
```

>> re: schemePrefix / locationPath

Let's us the terminology "endpoint", to be consistent https://docs.aws.amazon.com/general/latest/gr/s3.html 
What you have is clear, but just because this is part of cluster spec, I was thinking of start with minimum knobs. We start with endpoint and rootpath, and add scheme when we really need it.

HotSushi: Makes sense! ty resolved

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/configs/StorageProperties.java
```diff
@@ -0,0 +1,41 @@
...
+
+  @Getter
+  @Setter
+  @AllArgsConstructor
+  @NoArgsConstructor
+  @Builder(toBuilder = true)
+  public static class StorageTypeProperties {
+    private String scheme;
```
sumedhsakdeo, line 36: ==>

sumedhsakdeo (approved):

autumnust (approved):

PR #82, opened 2024-04-22
Add interfaces for StorageClient, Storage and its Implementation for Local

HotSushi:
## Summary
This PR sets foundation for `Storage` interfaces and provides a sample implementation called `LocalStorage`

two primary interfaces are:
```
Storage {
.. high level functionality expected from the storage
 }

StorageClient<T> {
 .. lower level client functionality such as token refresh.
}
```

This PR also adds `LocalStorage`,`LocalStorageClient` to show how new implementations should be added.

Methods in `Storage`,`StorageClient` are minimal right now, but they'll be extended as we do refactoring.

For more details [please see the document here](https://docs.google.com/document/d/1iPKol1jPzR6fq2YaoWbHHCYzBciCwfhqZqGyYhzZfoQ/edit?usp=sharing).

## Whats with `@Lazy`?
Yaml properties are not available when the `@Components` are initialized, they get configured after the components are initialized, therefore we cannot use handy tools like `@ConditionalOnProperty`/ `@ConditionalOnBean`. `@Lazy` is next best thing, it makes the bean initialize lazily and this is fantastic for heavy-lifting fsclients. This is also similar with our previous behavior: [HdfsStorageProvider](https://github.com/linkedin/openhouse/blob/main/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/filesystem/HdfsStorageProvider.java#L51)

## Testing Done
- [✅] Manually Tested on local docker setup. Please include commands ran, and their output.
- [✅] Added new tests for the changes made.
- [✅] Updated existing tests to reflect the changes made.


cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private static final StorageType.Type TYPE = StorageType.LOCAL;
+
+  @Autowired private StorageProperties storageProperties;
+
+  // Lazy initialization of the LocalStorageClient
+  @Autowired @Lazy private LocalStorageClient localStorageClient;
+
+  /**
+   * Check if the local storage is active.
```
HotSushi, line 29: This will help us configure localStorage if no storages are provided in the cluster.yaml

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorageClient.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Initialize the LocalStorageClient when the bean is accessed for the first time.
+   *
+   * @throws IOException if an I/O error occurs
+   */
+  @PostConstruct
+  private synchronized void init() throws IOException {
+    log.info("Initialized storage client for type: " + StorageType.LOCAL.getValue());
+    this.fs = FileSystem.get(new Configuration());
```
ctrezzo, line 32: You might want to do a check here to verify that the FileSystem instance returned is an instanceOf LocalFileSystem.

Depending on what is in the classpath, this call could return a FileSystem impl that is not local. It might be good to guarantee that this is the local filesystem for this class since it is called LocalStorageClient.

Another option that might even be stronger is to change the type parameter for the class to be LocalFileSystem instead of FileSystem.

HotSushi: great feedback! we don't need to initialize LocalStorage  based on classpath/configuration. Changed to instantiate purely `LocalFileSystem`

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorageClient.java
```diff
@@ -0,0 +1,34 @@
...
+@Component
+public class LocalStorageClient implements StorageClient<FileSystem> {
+
+  private FileSystem fs;
+
+  /** Initialize the LocalStorageClient when the bean is accessed for the first time. */
+  @PostConstruct
+  private synchronized void init() {
+    log.info("Initialized storage client for type: " + StorageType.LOCAL.getValue());
+    this.fs = new LocalFileSystem();
```
sumedhsakdeo, line 27: What about initializing this filesystem object, with 

https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/fs/LocalFileSystem.html#initialize-java.net.URI-org.apache.hadoop.conf.Configuration-

with endpoint `file:///` and rootPath '<path_to_root_dir>`?

And validate things like endpoint startsWith file:/// and path_to_root_dir exists?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
```diff
@@ -0,0 +1,45 @@
...
+ * system.
+ */
+public interface Storage {
+
+  /**
+   * Check if the storage is active.
+   *
+   * @return true if the storage is active, false otherwise
+   */
+  boolean isActive();
```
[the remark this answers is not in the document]
HotSushi: Storage is considered active if cluster-yaml has property defined for it.
for example in the end-state if we have cluster-yaml **only** as:
```
storages:
   hdfs: 
           endpoint:
           ..
```
then `HDFSStorage.isActive()=true` and `S3Storage.isActive=false`

it will be used by StorageManager to supply a Storage. Lemme know if its not clear, can raise a subsequent PR.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+@Component
+public class LocalStorage implements Storage {
+
+  private static final StorageType.Type TYPE = StorageType.LOCAL;
+
+  @Autowired private StorageProperties storageProperties;
+
+  // Lazy initialization of the LocalStorageClient
+  @Autowired @Lazy private LocalStorageClient localStorageClient;
```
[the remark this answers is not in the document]
HotSushi: `LocalStorageClient` will have a copy of storageProperties and it can use it to initialize appropriately. We should get to see it in: https://jira01.corp.linkedin.com:8443/browse/BDP-23325.

As for, where's rootPath/endpoint being used. I still haven't exposed them yet until we get to refactoring the codebase. I am debating whether to do: `Storage::allocateTableStorage()` or `Storage::getRootPath`.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorageClient.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * The LocalStorageClient class is an implementation of the StorageClient interface for local
+ * storage. It uses an Apache Hadoop FileSystem to interact with the local file system.
+ */
+@Slf4j
+@Lazy
+@Component
+public class LocalStorageClient implements StorageClient<FileSystem> {
```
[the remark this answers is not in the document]
HotSushi: there's some refactoring for fsStorageProvider that we need to do before we can run local docker recipe, [ex](https://jira01.corp.linkedin.com:8443/browse/BDP-23327)

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
HotSushi: yes! that was exactly the purpose. Having table service bootup without a cluster.yaml with certain default settings. `:tables-test-fixtures` also boots up without cluster.yaml

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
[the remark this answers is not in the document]
HotSushi: see the usage here: https://github.com/HotSushi/openhouse/pull/1/files#diff-3eab08eaf8a381dc4108e6db8d27ff2f34a126442a904758bb0f5ca9211c54bdR74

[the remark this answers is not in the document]
sumedhsakdeo: 👍

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
sumedhsakdeo: 👍

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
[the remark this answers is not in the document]
jainlavina: I had the same question. May be adda. comment describing what active means?
Do you think active means that storage is configured? May be isConfigured() is a better name?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Get the properties of the storage.
+   *
+   * @return a map of properties of the storage
+   */
+  Map<String, String> getProperties();
+
+  /**
+   * Get the type of the storage.
```
jainlavina, line 33: dd examples to demonstrate what type may look like and point to the code that lists all available types.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageClient.java
```diff
@@ -0,0 +1,22 @@
...
+ * {@link org.apache.hadoop.fs.FileSystem} to interact with the local file system.
+ *
+ * @param <T> the type of the native client of the storage system
+ */
+public interface StorageClient<T> {
+
+  /**
+   * Get the native client of the storage system.
+   *
+   * @return the native client of the storage system
```
jainlavina, line 19: Could it ever return a different client? Is "native" obvious and should we remove it and call it just getClient()?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Component;
+
+/**
+ * The LocalStorage class is an implementation of the Storage interface for local storage. It uses a
+ * LocalStorageClient to interact with the local file system.
```
jainlavina, line 16: LocalStorageClient in turn uses the hadoop FileSystem? Should we mention it here?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
```diff
@@ -0,0 +1,70 @@
...
+   * Check if the local storage is active.
+   *
+   * <p>The local storage is considered active if the default type is not set or no types are
+   * provided or specific "local" type is provided.
+   *
+   * @return true if the local storage is active, false otherwise
+   */
+  @Override
+  public boolean isActive() {
+    if (storageProperties.getDefaultType() == null) {
```
jainlavina, line 38: If a default type is not set but storageType is explicitly configured to be a type that is different from local, for example hdfs, in that case storage type should be set to hdfs and not local, right?

Default type should be used only if explicit type is not configured.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * The LocalStorage class is an implementation of the Storage interface for local storage. It uses a
+ * LocalStorageClient to interact with the local file system.
+ */
+@Component
+public class LocalStorage implements Storage {
+
+  private static final StorageType.Type TYPE = StorageType.LOCAL;
```
jainlavina, line 21: name it LOCALTYPE or LOCAL_STORAGE_TYPE?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
```diff
@@ -0,0 +1,70 @@
...
+   * @return true if the local storage is active, false otherwise
+   */
+  @Override
+  public boolean isActive() {
+    if (storageProperties.getDefaultType() == null) {
+      return true;
+    } else if (storageProperties.getTypes() == null || storageProperties.getTypes().isEmpty()) {
+      return true;
+    } else {
+      return storageProperties.getTypes().containsKey(TYPE.getValue());
```
jainlavina, line 43: If you name this LOCALTYPE or something like that this code will be easier to read and understand without looking at what TYPE is set to.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  /**
+   * Get the properties of the local storage.
+   *
+   * @return a copy map of properties of the local storage
```
jainlavina, line 50: Did you mean to say "a map of" or "a copy of map of"?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
```diff
@@ -0,0 +1,70 @@
...
+   */
+  @Override
```
jainlavina, line 52: This code will likely be common to all storage types? Should it be in a base class then? It is ok to add it here for now and when we add the second storage type, we can move code to avoid duplication.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorageClient.java
jainlavina: +1. This should be initialized with the hadoop configuration.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
HotSushi: Added

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
HotSushi: Yes, we could pursue it in the next PR!

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
HotSushi: yes correct! default type will only be configured if explicit type is not configured. I've added more validations+tests for this.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageClient.java
HotSushi: I see "native" adding value when we see the complete usage pattern:
```
Storage storage = localStorage;
storage.getClient().getNativeClient()
```
as opposed to 
```
storage.getClient().getClient()
```

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorageClient.java
HotSushi: Agree, added. PTAL.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
jainlavina: probably didn't explain clearly.

If the defaultType == null and type == "hdfs" then storageType will be hdfs and not local, correct?
In that case retuning that local storage is active if storageType == null is incorrect.

HotSushi: if `defaultType==null` and `types.hdfs != null` then storageType will not be configured. we should see a boot time exception. I have it lined up in the next PR: https://github.com/HotSushi/openhouse/pull/1/files#diff-3eab08eaf8a381dc4108e6db8d27ff2f34a126442a904758bb0f5ca9211c54bdR34

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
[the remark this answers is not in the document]
HotSushi: yes! isConfigured is definitily a better name, will address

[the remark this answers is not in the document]
HotSushi: Addressed

sumedhsakdeo (approved):

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
jainlavina: Looks like you are assuming an invariant that if type is configured then default should also be set? Is that right?

May be list down the invariants on the config somewhere, enforce them as you said and assert them here?

PR #90, opened 2024-04-25
Introduce StorageManager and an Hdfs Implementation for Storage

HotSushi:
## Summary
Laying foundations for storage part 3: `StorageManager` and an implementation for Hdfs: `HdfsStorage`.

StorageManager interface looks like
```
StorageManager {
  Storage getDefaultStorage()
  Storage getStorage(Type)
}
```

It runs through all configured Storages and returns the appropriate storage for default or for particular type.

`HdfsStorage` is yet another implementation for `Storage` similar to `LocalStorage`. 
We've abstracted the common code for all storage-impls into **BaseStorage**

## Changes
- [X] New Features
Other related PRs:
https://github.com/linkedin/openhouse/pull/76
https://github.com/linkedin/openhouse/pull/82

## Testing Done

### DIFF
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
@@ -1,0 +1,44 @@
+package com.linkedin.openhouse.cluster.storage;
+
+import com.linkedin.openhouse.cluster.storage.configs.StorageProperties;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * The BaseStorage class is an abstract class that implements the Storage interface. It provides
+ * common functionality for all storage implementations.
+ */
+public abstract class BaseStorage implements Storage {
+
+  @Autowired private StorageProperties storageProperties;
+
+  /**
+   * Check if the storage is configured.
+   *
+   * <p>The storage is considered configured if type is defined in the storage properties.
+   *
+   * @return true if the storage is configured, false otherwise
+   */
+  @Override
+  public boolean isConfigured() {
+    return Optional.ofNullable(storageProperties.getTypes())
+        .map(types -> types.containsKey(getType().getValue()))
+        .orElse(false);
+  }
+
+  /**
+   * Get the properties of the storage.
+   *
+   * @return a copy of map of properties of the storage
+   */
+  @Override
+  public Map<String, String> getProperties() {
+    return Optional.ofNullable(storageProperties.getTypes())
+        .map(types -> types.get(getType().getValue()))
+        .map(StorageProperties.StorageTypeProperties::getParameters)
+        .map(HashMap::new)
+        .orElseGet(HashMap::new);
+  }
+}
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
@@ -1,0 +1,96 @@
+package com.linkedin.openhouse.cluster.storage;
+
+import static com.linkedin.openhouse.cluster.storage.StorageType.LOCAL;
+
+import com.google.common.base.Preconditions;
+import com.linkedin.openhouse.cluster.storage.configs.StorageProperties;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import javax.annotation.PostConstruct;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
+import org.springframework.util.StringUtils;
+
+/**
+ * The StorageManager class is responsible for managing the storage types and providing the
+ * appropriate storage implementation based on the configuration.
+ */
+@Component
+public class StorageManager {
+
+  @Autowired StorageProperties storageProperties;
+
+  @Autowired StorageType storageType;
+
+  @Autowired List<Storage> storages;
+
+  /**
+   * Validate the storage properties.
+   *
+   * <p>It validates storage properties as follows:
+   *
+   * <p>1. If any storage type is configured, then default type must be set. Alternatively, if a
+   * default type is not set, then storage types should also be null or empty. **valid**
+   *
+   * <p>2. If default-type is set, then the value of the default type must exist in the configured
+   * storage types. **valid**
+   *
+   * <p>all other configurations are **invalid**
+   */
+  @PostConstruct
+  public void validateProperties() {
+    String clusterYamlError = "Cluster yaml is incorrectly configured: ";
+    if (StringUtils.hasText(storageProperties.getDefaultType())) {
+      // default-type is configured, types should contain the default-type
+      Preconditions.checkArgument(
+          !CollectionUtils.isEmpty(storageProperties.getTypes())
+              && storageProperties.getTypes().containsKey(storageProperties.getDefaultType()),
+          clusterYamlError
+              + "storage types should contain the default-type: "
+              + storageProperties.getDefaultType());
+    } else {
+      // default-type is not configured, types should be null or empty
+      Preconditions.checkArgument(
+          CollectionUtils.isEmpty(storageProperties.getTypes()),
+          clusterYamlError + "default-type must be set if storage types are configured");
+    }
+    try {
+      Optional.ofNullable(storageProperties.getDefaultType()).ifPresent(storageType::fromString);
+      Optional.ofNullable(storageProperties.getTypes())
+          .map(Map::keySet)
+          .ifPresent(keyset -> keyset.forEach(key -> storageType.fromString(key)));
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException(clusterYamlError + e.getMessage());
+    }
+  }
+
+  /**
+   * Get the default storage.
+   *
+   * @return the default storage
+   */
+  public Storage getDefaultStorage() {
+    if (!StringUtils.hasText(storageProperties.getDefaultType())) {
+      return getStorage(LOCAL);
+    }
+    return getStorage(storageType.fromString(storageProperties.getDefaultType()));
+  }
+
+  /**
+   * Get the storage based on the storage type.
+   *
+   * @param storageType the storage type
+   * @return the storage
+   */
+  public Storage getStorage(StorageType.Type storageType) {
+    for (Storage storage : storages) {
+      if (storage.getType().equals(storageType) && storage.isConfigured()) {
+        return storage;
+      }
+    }
+    throw new IllegalArgumentException(
+        "No configured storage found for type: " + storageType.getValue());
+  }
+}
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorage.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorage.java
@@ -1,0 +1,39 @@
+package com.linkedin.openhouse.cluster.storage.hdfs;
+
+import com.linkedin.openhouse.cluster.storage.BaseStorage;
+import com.linkedin.openhouse.cluster.storage.StorageClient;
+import com.linkedin.openhouse.cluster.storage.StorageType;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Component;
+
+/**
+ * The HdfsStorage class is an implementation of the Storage interface for HDFS storage. It uses a
+ * HdfsStorageClient to interact with the HDFS file system. The HdfsStorageClient uses the {@link
+ * org.apache.hadoop.fs.FileSystem} class to interact with the HDFS file system.
+ */
+@Component
+public class HdfsStorage extends BaseStorage {
+
+  @Autowired @Lazy private HdfsStorageClient hdfsStorageClient;
+
+  /**
+   * Get the type of the HDFS storage.
+   *
+   * @return the type of the HDFS storage
+   */
+  @Override
+  public StorageType.Type getType() {
+    return StorageType.HDFS;
+  }
+
+  /**
+   * Get the HDFS storage client.
+   *
+   * @return the HDFS storage client
+   */
+  @Override
+  public StorageClient<?> getClient() {
+    return hdfsStorageClient;
+  }
+}
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorageClient.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorageClient.java
@@ -1,0 +1,66 @@
+package com.linkedin.openhouse.cluster.storage.hdfs;
+
+import com.google.common.base.Preconditions;
+import com.linkedin.openhouse.cluster.storage.StorageClient;
+import com.linkedin.openhouse.cluster.storage.StorageType;
+import com.linkedin.openhouse.cluster.storage.configs.StorageProperties;
+import java.io.IOException;
+import javax.annotation.PostConstruct;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.hadoop.fs.FileSystem;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
+
+/**
+ * The HdfsStorageClient class is an implementation of the StorageClient interface for HDFS Storage.
+ * It uses the {@link FileSystem} class to interact with the HDFS file system.
+ */
+@Slf4j
+@Lazy
+@Component
+public class HdfsStorageClient implements StorageClient<FileSystem> {
+
+  private FileSystem fs;
+
+  @Autowired private StorageProperties storageProperties;
+
+  private static final StorageType.Type HDFS_TYPE = StorageType.HDFS;
+
+  /** Initialize the HdfsStorageClient when the bean is accessed for the first time. */
+  @PostConstruct
+  public synchronized void init() throws IOException {
+    validateProperties();
+    StorageProperties.StorageTypeProperties hdfsStorageProperties =
+        storageProperties.getTypes().get(HDFS_TYPE.getValue());
+    org.apache.hadoop.conf.Configuration configuration = new org.apache.hadoop.conf.Configuration();
+    configuration.set("fs.defaultFS", hdfsStorageProperties.getEndpoint());
+    fs = FileSystem.get(configuration);
+  }
+
+  /** Validate the storage properties. */
+  private void validateProperties() {
+    log.info("Initializing storage client for type: " + HDFS_TYPE);
+    Preconditions.checkArgument(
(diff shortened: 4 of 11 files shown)


cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
```diff
@@ -0,0 +1,93 @@
...
+public class StorageManager {
+
+  @Autowired StorageProperties storageProperties;
+
+  @Autowired StorageType storageType;
+
+  @Autowired List<Storage> storagesBeans;
+
+  /**
+   * Validate the storage properties.
```
HotSushi, line 28: @jainlavina  Invariants on the config and their enforcement

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * The BaseStorage class is an abstract class that implements the Storage interface. It provides
+ * common functionality for all storage implementations.
+ */
+public abstract class BaseStorage implements Storage {
```
HotSushi, line 13: Abstraction for common Impl code for all storages

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorageClient.java
```diff
@@ -0,0 +1,59 @@
...
+        "Storage properties doesn't contain type: " + HDFS_TYPE.getValue());
+    Preconditions.checkArgument(
+        hdfsStorageProperty.getEndpoint() != null,
+        "Storage properties doesn't contain endpoint for: " + HDFS_TYPE.getValue());
+    Preconditions.checkArgument(
+        hdfsStorageProperty.getRootPath() != null,
+        "Storage properties doesn't contain rootpath for: " + HDFS_TYPE.getValue());
+    org.apache.hadoop.conf.Configuration configuration = new org.apache.hadoop.conf.Configuration();
+    configuration.set("fs.defaultFS", hdfsStorageProperty.getEndpoint());
+    fs = FileSystem.get(configuration);
```
sumedhsakdeo, line 52: @abhisheknath2011 how is our hdfs conf loaded today?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
```diff
@@ -0,0 +1,93 @@
...
+   * <p>2. default-type for storage is set and storage types are provided ,and it contains
+   * default-type **valid**
+   *
+   * <p>all other configurations are **invalid**
+   */
+  @PostConstruct
+  public void validateProperties() {
+    String clusterYamlError = "Cluster yaml is incorrectly configured: ";
+    Preconditions.checkArgument(
+        !(storageProperties.getDefaultType() == null
```
jainlavina, line 44: Would this be better:

if (storageProperties.getDefaultType() == null || storageProperties.getDefaultType().isEmpty()) {
  // Default type not configured. Storage types should also be null or empty.
  Preconditions.checkArgument(storageProperties.getTypes() == null || storageProperties.getTypes().isEmpty(), "");
} else {
  // The default type must be one of the configured storage types.
  Preconditions.checkArgument(storageProperties.getTypes() != null && !storageProperties.getTypes.isEmpty() && storageProperties.getTypes().containsKey(storageProperties.getDefaultType()), "");
}

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+  /**
+   * Get the default storage.
+   *
+   * @return the default storage
+   */
+  public Storage getDefaultStorage() {
+    if (storageProperties.getDefaultType() == null) {
```
jainlavina, line 72: What about empty?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorageClient.java
abhisheknath2011: We bake in Hadoop/Hdfs conf as part of docker image and that gets mounted to the services pod as part of deployment. This should work provided we inject hadoop dependencies and set appropriate hadoop classpath during deployment.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorageClient.java
```diff
@@ -0,0 +1,59 @@
...
+        hdfsStorageProperty != null,
+        "Storage properties doesn't contain type: " + HDFS_TYPE.getValue());
+    Preconditions.checkArgument(
+        hdfsStorageProperty.getEndpoint() != null,
+        "Storage properties doesn't contain endpoint for: " + HDFS_TYPE.getValue());
+    Preconditions.checkArgument(
+        hdfsStorageProperty.getRootPath() != null,
+        "Storage properties doesn't contain rootpath for: " + HDFS_TYPE.getValue());
+    org.apache.hadoop.conf.Configuration configuration = new org.apache.hadoop.conf.Configuration();
+    configuration.set("fs.defaultFS", hdfsStorageProperty.getEndpoint());
```
abhisheknath2011, line 51: How about adding additional resources such as hadoop core site and hdfs site path as part of configuration resources?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
abhisheknath2011, line 44: What is the significance of default type? I see that StorageProperties has default type field and is this going to be populated from cluster.yaml. If that is the case can we enforce to have always a default type (either through code/config) and simplify the validation below?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
```diff
@@ -0,0 +1,93 @@
...
+      return getStorage(LOCAL);
```
abhisheknath2011, line 73: So when local storage is used there is no default type. Do you think we can keep the default type as `local` (at code level) that would simplify things?  As part of deployment default type can be overridden through config with hdfs or other types?

abhisheknath2011 (commented):
Thanks @HotSushi . Added some comments.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
```diff
@@ -0,0 +1,93 @@
...
+  @Autowired StorageType storageType;
+
+  @Autowired List<Storage> storagesBeans;
+
+  /**
+   * Validate the storage properties.
+   *
+   * <p>It validates storage properties as follows
+   *
+   * <p>1. default-type for storage is not set and storage types are not provided, ie. LocalStorage
```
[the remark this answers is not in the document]
HotSushi: Yes your interpretation is correct. Thanks for this simplification! I've adopted in the PR.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
```diff
@@ -0,0 +1,93 @@
...
+   *
+   * <p>2. default-type for storage is set and storage types are provided ,and it contains
+   * default-type **valid**
+   *
+   * <p>all other configurations are **invalid**
+   */
+  @PostConstruct
+  public void validateProperties() {
+    String clusterYamlError = "Cluster yaml is incorrectly configured: ";
+    Preconditions.checkArgument(
```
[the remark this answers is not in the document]
HotSushi: checkState would lead to `IllegalStateException`, I like bad cluster.yaml  giving `checkArgument`, would vote to keep the same.

Adopted if block with precondition for simplification.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
HotSushi: yes better! thank you.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorageClient.java
HotSushi: I looked into our docker setup and deployments, we do not use core-site, hdfs-site. Do you think we should still provision for it here?.. its possible through configurable properties.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
HotSushi: yes it will be populated from cluster.yaml.

> can we enforce..simplify the validation 

that was one path we considered.. but decided to go ahead with this approach because we want to keep table-service backward compatible for when cluster.yaml is not provided. This setup is used for  1)quick bootrun 2) tables-test-fixtures.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
HotSushi: the  clusteryaml -> storageProperties parser doesn't support keeping default-type the way ClusterProperties does. The if condition are basically:
1- clusteryaml configured at all
2- clusteryaml properly configured with default-type and types

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
```diff
@@ -0,0 +1,94 @@
...
+ * appropriate storage implementation based on the configuration.
+ */
+@Component
+public class StorageManager {
+
+  @Autowired StorageProperties storageProperties;
+
+  @Autowired StorageType storageType;
+
+  @Autowired List<Storage> storages;
```
jainlavina, line 27: What is this list. Given that this class has a single storageType and storageProperties field, it looks like it manages only one storage type, the one that is configured.

But this list indicated that it stores a list of storage types configured?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * Validate the storage properties.
+   *
+   * <p>It validates storage properties as follows
```
jainlavina, line 32: nit follows:

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
```diff
@@ -0,0 +1,94 @@
...
+   * <p>all other configurations are **invalid**
+   */
+  @PostConstruct
+  public void validateProperties() {
+    String clusterYamlError = "Cluster yaml is incorrectly configured: ";
+    if (!StringUtils.hasText(storageProperties.getDefaultType())) {
+      // default-type is not configured, types should be null or empty
+      Preconditions.checkArgument(
+          CollectionUtils.isEmpty(storageProperties.getTypes()),
+          clusterYamlError + "default-type is not set, storage types should not be provided");
```
jainlavina, line 49: Likely the user error here is that they forgot to configure default vs mistakenly provided the storage types. Do you agree?
In that case, should we change the error message to: "default-type must be set if storage types are configured"?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
```diff
@@ -0,0 +1,94 @@
...
+    } else {
+      // default-type is configured, types should contain the default-type
+      Preconditions.checkArgument(
+          !CollectionUtils.isEmpty(storageProperties.getTypes())
+              && storageProperties.getTypes().containsKey(storageProperties.getDefaultType()),
+          clusterYamlError + "storage types should contain the default-type");
```
jainlavina, line 55: Log the default-type here?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorageClient.java
```diff
@@ -0,0 +1,59 @@
...
+
+  @Autowired private StorageProperties storageProperties;
+
+  private static final StorageType.Type HDFS_TYPE = StorageType.HDFS;
+
+  /** Initialize the HdfsStorageClient when the bean is accessed for the first time. */
+  @PostConstruct
+  public synchronized void init() throws IOException {
+    log.info("Initializing storage client for type: " + HDFS_TYPE);
+    Preconditions.checkArgument(
```
jainlavina, line 34: Remove negation to make the code easier to read. Is this the same as:
Preconditions.checkArgument(getTypes() != null && !getTypes().isEmpty() && getTypes().containsKey())

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorageClient.java
```diff
@@ -0,0 +1,59 @@
...
+        !(storageProperties.getTypes() == null
+            || storageProperties.getTypes().isEmpty()
+            || !storageProperties.getTypes().containsKey(HDFS_TYPE.getValue())),
+        "Storage properties doesn't contain type: " + HDFS_TYPE.getValue());
+    StorageProperties.StorageTypeProperties hdfsStorageProperty =
```
jainlavina, line 39: nit: s/hdfsStorageProprty/hdfsStorageProperties

jainlavina (approved):
Left a few comments. But looks good otherwise.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorageClient.java
abhisheknath2011: Yes we have not used them through configs. Like you mentioned we can make them available through configurable properties.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
abhisheknath2011: Got it. Thanks for the clarification.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
HotSushi: Springboot configures it with all beans of type "Storage" and we iterate through it to return  the required type

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorageClient.java
HotSushi: fixed! my brain likes to think negation first 🤣

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
HotSushi: Good point! addressed

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorageClient.java
HotSushi: do you think we should still pursue this, or wait for the feature request?

abhisheknath2011: I think we should be good for now. Good to keep a note of this for future.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorageClient.java
```diff
@@ -0,0 +1,59 @@
...
+
+  @Autowired private StorageProperties storageProperties;
+
+  private static final StorageType.Type HDFS_TYPE = StorageType.HDFS;
+
+  /** Initialize the HdfsStorageClient when the bean is accessed for the first time. */
+  @PostConstruct
+  public synchronized void init() throws IOException {
+    log.info("Initializing storage client for type: " + HDFS_TYPE);
+    Preconditions.checkArgument(
```
abhisheknath2011, line 35: Should we move all the preconditions check to a separate method and invoke from here to keep the client initialization part clean. If precondition fails it should throw exception and should not proceed with initialization.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
```diff
@@ -0,0 +1,96 @@
...
+   *
+   * <p>2. If default-type is set, then the value of the default type must exist in the configured
+   * storage types. **valid**
+   *
+   * <p>all other configurations are **invalid**
+   */
+  @PostConstruct
+  public void validateProperties() {
+    String clusterYamlError = "Cluster yaml is incorrectly configured: ";
+    if (!StringUtils.hasText(storageProperties.getDefaultType())) {
```
maluchari, line 45: Nit: For readability can we do 
if ( condition) .... else
instead of if (!condition) .... else
In this method and also in https://github.com/linkedin/openhouse/pull/90/files#diff-3eab08eaf8a381dc4108e6db8d27ff2f34a126442a904758bb0f5ca9211c54bdR75

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorage.java
```diff
@@ -0,0 +1,39 @@
...
+import com.linkedin.openhouse.cluster.storage.StorageClient;
+import com.linkedin.openhouse.cluster.storage.StorageType;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Component;
+
+/**
+ * The HdfsStorage class is an implementation of the Storage interface for HDFS storage. It uses a
+ * HdfsStorageClient to interact with the HDFS file system. The HdfsStorageClient uses an Apache
+ * Hadoop FileSystem to interact with the HDFS file system.
```
maluchari, line 13: The HdfsStorageClient uses an Apache Hadoop FileSystem to interact with the HDFS file system.
Not following this comment.

HotSushi: got it! replaced the confusing string with class link.

abhisheknath2011 (approved):
LGTM. Thanks for addressing the comments/questions!

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/StorageManager.java
jainlavina: Then may be add a comment explaining that this is a list of all available storage types.

jainlavina (approved):

abhisheknath2011 (approved):

PR #100, opened 2024-05-09
Refactor: Remove @LegacyFileIO and Make DelegationRefreshToken use new cluster.yaml

HotSushi:
## Summary
Let's start refactoring the old code out 🙌 

There are two refactors:
1) Remove old [FileIO instance provisioned via FsStorageProvider](https://github.com/linkedin/openhouse/blob/bc5287cfa86cea52c2215fbf9259b11992f18a42/services/tables/src/main/java/com/linkedin/openhouse/tables/config/MainApplicationConfig.java#L54)
```
// old code
public class MainApplicationConfig
  @Bean
  public FileIO provideIcebergFileIO() {
         return new HadoopFileIO(fsStorageProvider.storageClient().getConf())
  }
```
2) `DelegationTokenRefresher` should use the new cluster.yaml
```
 // new cluster yaml
storages:
  default-type: "hdfs"
  types:
     hdfs:
        rootpath: "/tmp/unittest"
        endpoint: "hdfs://localhost:9000"
        parameters:
           token.refresh.enabled: "true"             <----------- THIS
           token.refresh.schedule.cron: "* * * * * ?"          <----------- THIS
```

There are no major changes to code otherwise.

## Changes
- [X] Refactoring
- [X] Tests

## Testing Done
✅ Added unit tests for token refresh that were missing before 🥳 
✅ mint build and existing tests pass
✅ manually tested token refresh code in docker setup
✅ manually tested fileIO code in docker setup


HotSushi:
cc: @ctrezzo

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsDelegationTokenRefresherConfig.java
```diff
@@ -0,0 +1,55 @@
...
+   * @return HdfsDelegationTokenRefresher
+   */
+  @Bean
+  public HdfsDelegationTokenRefresher getDelegationTokenRefresher() {
+    if (!hdfsStorage.isConfigured()) {
+      log.debug(
+          "Hdfs storage is not configured, ignoring HdfsDelegationTokenRefresher bean creation");
+      return null;
+    }
+    String refreshEnabled =
```
autumnust, line 44: Is there a need of this param if there's something as `hdfsStorage.isConfigured()` ?
I mean, not configured == not enabled. Is that sufficient?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
```diff
@@ -35,9 +36,9 @@ public class OpenHouseInternalCatalog extends BaseMetastoreCatalog {
 
   @Autowired HouseTableRepository houseTableRepository;
 
-  @Autowired
-  @Qualifier("LegacyFileIO")
-  FileIO fileIO;
+  @Autowired FileIOManager fileIOManager;
```
autumnust, line 39: [Comments For discussion, no AI needed]
When you do the delta prototype, is this whole catalog implementation re-done using Delta's counterpart, so that we won't need to instantiate this bean ? If that's the case, it is nice that we separate `FileIOManager` out of `StorageManager` !

autumnust (changes requested):
Is there coordination required when picking up this change internally, since you are making some config changes ?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsDelegationTokenRefresherConfig.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.scheduling.annotation.EnableScheduling;
+
+/**
+ * Configuration class to conditionally enable/disable creation of {@link
+ * HdfsDelegationTokenRefresher} bean and enable delegation token refresh scheduling.
+ */
+@Slf4j
+@Configuration
+@EnableScheduling
+public class HdfsDelegationTokenRefresherConfig {
```
abhisheknath2011, line 16: I believe this is the existing class and rename is done. With some refactoring, the functionality remains the same, right?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsDelegationTokenRefresherConfig.java
```diff
@@ -0,0 +1,55 @@
...
+   */
+  @Bean
+  public HdfsDelegationTokenRefresher getDelegationTokenRefresher() {
+    if (!hdfsStorage.isConfigured()) {
+      log.debug(
+          "Hdfs storage is not configured, ignoring HdfsDelegationTokenRefresher bean creation");
+      return null;
+    }
+    String refreshEnabled =
+        hdfsStorage.getProperties().getOrDefault(HDFS_TOKEN_REFRESH_ENABLED, "false");
```
abhisheknath2011, line 45: Should we enable delegation token refresh by default if HDFS storage is configured? Because token needs to be refreshed if HDFS is accessed.

abhisheknath2011: One issue I see with OSS codebase is token is not available by default. The HADOOP_TOKEN_FILE_LOCATION can be presented token is not in UGI. Off course this can be enabled using config. For li internal repo all of these are enabled by default using deployment.

abhisheknath2011: But again OSS perspective there could other option as well Kerberos etc, might not rely on UGI approach like we rely on UGI and hadoop token file internally at Li. I think should be okay to have it disabled by default.

ctrezzo: It seems like it would make sense to have token refresh disabled by default. Security is disabled by default in OSS Hadoop.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsDelegationTokenRefresherConfig.java
HotSushi: yes!

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsDelegationTokenRefresherConfig.java
HotSushi: yes it is sufficient for `not configured == not enabled` case, but we need a yaml flag to turn the feature on/off if hdfs is configured

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
HotSushi: ➕ that's my plan! FileIOManager should only be internal to iceberg layer and StorageManager is OH-level. I explored this idea in [design decision:Can we add a top level TableStorage abstraction for combining all FS operations?]( https://docs.google.com/document/d/1gSQhmn7SVsjVCipojSnlx2-6rRBq1-hK-KiFRuWTmXU/edit#heading=h.pueww4o8c12s). For delta we'd need to reimplement it appropriately.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsDelegationTokenRefresherConfig.java
```diff
@@ -0,0 +1,55 @@
...
+   * </pre>
+   *
+   * @return HdfsDelegationTokenRefresher
+   */
+  @Bean
+  public HdfsDelegationTokenRefresher getDelegationTokenRefresher() {
+    if (!hdfsStorage.isConfigured()) {
+      log.debug(
+          "Hdfs storage is not configured, ignoring HdfsDelegationTokenRefresher bean creation");
+      return null;
```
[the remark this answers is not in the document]
HotSushi: I gave that a shot during the early implementations for clusterspec and storageProperties. ConditionalOnBean/ConditionalOnExpression do not seem to be working nicely with storageProperties. Springboot would evaluate those conditions before the storageproperties were set. Returning null is a good alternative.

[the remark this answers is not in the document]
rohitkum2506: Te drawback of this is it requires a null check in code using `HdfsDelegationTokenRefresher` otherwise NPE is on cards.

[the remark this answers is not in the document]
abhisheknath2011: Earlier implementation was using @Conditonal annotation. hmm, so the refactoring change does not work well with conditional.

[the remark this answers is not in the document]
HotSushi: Not really, the behavior is exactly the same. If null is returned by this method,
- its not the case that bean *is available with null value*
- its the case that bean *is not available*

abhisheknath2011 (approved):
LGTM

rohitkum2506 (approved):

autumnust (approved):
thanks for answering the questions. lgtm

PR #109, opened 2024-05-27
Data layout optimization (strategy generation). Part 2: data source for statistics/query logs

teamurko:
## Summary

This is part 2 of a new feature: data layout optimization library, strategy generation.
Added data source interface/implementation. This PR builds on top of https://github.com/linkedin/openhouse/pull/108

The following 3 components will be added eventually:
1) DLO library that has primitives for generating data layout optimization strategies
2) App that generates strategies for all tables
3) Scheduling of the app

## Changes

- [ ] Client-facing API Changes
- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Performance Improvements
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

For all the boxes checked, please include additional details of the changes made in this pull request.  

## Testing Done

- [ ] Manually Tested on local docker setup. Please include commands ran, and their output.
- [x] Added new tests for the changes made.
- [ ] Updated existing tests to reflect the changes made.
- [ ] No tests added or updated. Please explain why. If unsure, please feel free to ask for help.
- [ ] Some other form of testing like staging or soak time in production. Please explain.

For all the boxes checked, include a detailed description of the testing done for the changes made in this pull request.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations
- [x] Large PR broken into smaller PRs, and PR plan linked in the description.

For all the boxes checked, include additional details of the changes made in this pull request.

PR #116, opened 2024-06-04
Data layout optimization (strategy generation). Part 3: compaction strategy generation with cost/gain scores

teamurko:
## Summary
This is part 3 of a new feature: data layout optimization library, strategy generation. This PR is co-authored with @anjagruenheid.

Added compaction strategy generation with rewrite cost as serial rewrite time and rewrite gain as time-saving from number of files reduced. This PR builds on top of https://github.com/linkedin/openhouse/pull/109

The following 3 components will be added eventually:
1) DLO library that has primitives for generating data layout optimization strategies
2) App that generates strategies for all tables
3) Scheduling of the app

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


sumedhsakdeo:
Review posted on https://github.com/teamurko/openhouse/pull/2

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+  private static final long FILE_BLOCK_MARGIN_BYTES = 10 * MB;
+  private static final long NUM_FILE_GROUPS_PER_COMMIT = 100;
+  private static final long MAX_NUM_COMMITS = 30;
+  private static final long MAX_BYTES_SIZE_RATIO = 10;
+  private static final long REWRITE_BYTES_PER_SECOND = 2 * MB;
+  private static final long EXECUTOR_MEMORY_MB = 2048;
+  private static final int MAX_CONCURRENT_FILE_GROUP_REWRITES = 50;
+  private static final int REWRITE_PARALLELISM = 900; // number of Spark tasks to run in parallel
+  private static final long TARGET_BYTES_SIZE = 2 * FILE_BLOCK_SIZE_BYTES - FILE_BLOCK_MARGIN_BYTES;
+  private static final double COMPUTE_STARTUP_COST_GB_HR = 0.5;
```
teamurko, line 32: Most of these constants will be configurable in the next PR when I add the wrapper app

libs/datalayout/src/test/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGeneratorTest.java
```diff
@@ -0,0 +1,62 @@
...
+      List<DataLayoutOptimizationStrategy> strategies = strategyGenerator.generate();
+      Assertions.assertEquals(1, strategies.size());
+      DataLayoutOptimizationStrategy strategy = strategies.get(0);
+      // few groups, expect 1 commit
+      Assertions.assertEquals(1, strategy.getConfig().getPartialProgressMaxCommits());
+      Assertions.assertTrue(strategy.getConfig().isPartialProgressEnabled());
+
+      Assertions.assertTrue(
+          strategy.getGain() >= 4,
+          "Gain for 6 files compaction in 2 partitions should be at least 4");
```
anjagruenheid, line 55: Shouldn't it be exactly 4 since 3+3 files (3 files each in 2 partitions) would be compacted to 1+1?

teamurko: It would be exactly 4 if we grouped files by partition when calculating cost/gain, it's not happening now due to complexity and low ROI, for production tables I believe the error wouldn't matter.

anjagruenheid: Okay but this is just an assertion for the gain check, right? So for the gain to be correctly calculated, it should be an equivalence check here?

anjagruenheid:
I thought a bit more about the gain calculation and we can generally calculate it as follows: For each query, the gain (in terms of files touched), is the new number of files retrieved for the query subtracted from the original number of files. So assume that your underlying data is unsorted and the data is completely evenly distributed. Say you have 10 files, each containing values in col1 (0-100) and col2 (0-10) and for every query, we'd have to touch all current files because none of the files have tight bounds. Now you can compare it to the gain for 2 different strategies:
1) Partitioned
For each query, the gain is the current number of files minus the files in each partition touched by the query. In summary, SUM_j #current files - (SUM_{P_i,j}(v_ij)) where j denotes all queries q_j and v_ij is either 0 if q_j does not touch partition P_i or the number of files if it touches the partition.
2) Sorted
For each query, the gain is the current number of files minus the subset of files that fall within the range of the query's values. I.e., SUM_j #current files - v_j where v_j denotes all files that are touched by query q_j when the underlying data is sorted.

To compute the gain, you'd need to a) create in-memory structures that represent the file layout and b) 'rerun' the queries of your workload against the structure, simulating how the new data layout would perform. For the former, you'll need to mimic how the data would get rearranged. So you'd read the data once and for example for partitioning, you maintain buckets for each partitioning key, counting their frequency and then splitting it according to how many records you'd expect per file. So if you have only 5 records/file at most and your partition has 6 records, then upper(6/5)=2 files.

Also note that if the workload does not (substantially) change, you don't have to rerun this calculation, so it's a one-off effort for data layout selection. In contrast, data maintenance cost/benefit calculations have to be executed frequently to maintain the chosen data layout.

libs/datalayout/src/test/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGeneratorTest.java
teamurko: Yes, we can make it exact and change the value after taking partitions into account later.

teamurko:
> I thought a bit more about the gain calculation and we can generally calculate it as follows: For each query, the gain (in terms of files touched), is the new number of files retrieved for the query subtracted from the original number of files. So assume that your underlying data is unsorted and the data is completely evenly distributed. Say you have 10 files, each containing values in col1 (0-100) and col2 (0-10) and for every query, we'd have to touch all current files because none of the files have tight bounds. Now you can compare it to the gain for 2 different strategies:
> 
> 1. Partitioned
>    For each query, the gain is the current number of files minus the files in each partition touched by the query. In summary, SUM_j #current files - (SUM_{P_i,j}(v_ij)) where j denotes all queries q_j and v_ij is either 0 if q_j does not touch partition P_i or the number of files if it touches the partition.
> 2. Sorted
>    For each query, the gain is the current number of files minus the subset of files that fall within the range of the query's values. I.e., SUM_j #current files - v_j where v_j denotes all files that are touched by query q_j when the underlying data is sorted.
> 
> To compute the gain, you'd need to a) create in-memory structures that represent the file layout and b) 'rerun' the queries of your workload against the structure, simulating how the new data layout would perform. For the former, you'll need to mimic how the data would get rearranged. So you'd read the data once and for example for partitioning, you maintain buckets for each partitioning key, counting their frequency and then splitting it according to how many records you'd expect per file. So if you have only 5 records/file at most and your partition has 6 records, then upper(6/5)=2 files.
> 
> Also note that if the workload does not (substantially) change, you don't have to rerun this calculation, so it's a one-off effort for data layout selection. In contrast, data maintenance cost/benefit calculations have to be executed frequently to maintain the chosen data layout.

Thank you, Anja, for brainstorming ideas. I have been looking at it a bit more. Let me share some findings and ideas and we'll see how we can consolidate it all together and make it work for both compaction and query optimization.

My current understanding about partitioning vs sorting is that these are orthogonal. A table can be partitioned or not, if it's sorted and partitioned, sorting within partitions happens independently in Iceberg. WLOG, we can assume non-partitioned table.

I'm thinking about 2 aspects of query optimization: 1) # files opened; 2) # records scanned.

(*) For a splittable file format like orc, parquet with standard compression, Iceberg creates file split tasks. If we have 1 file 1gb with configurable split size 200mb, there will be 5 splits, or 5 tasks. If we have 10 files 100mb each, there will be still 5 splits, each split referencing 2 files.  In the first case there will be 5 file opens, in second - 10. File split is a kind of abstraction from files.

If the file open cost is high and we have many small files, e.g. 1000 files of 1mb each, we'll pay (file open cost) * 1000 / (# executors), it's worse if there a skew such that most file opens happened in one task, this would delay query execution. Although sorting on its own could reduce # file opens, say if we have mostly non-intersecting small files, compaction is part of sorting rewrite strategy in Iceberg, so small files problem automatically goes away.

Compaction can't solve 2), generally all records in all file splits has to be scanned. Sorting works on top of bin packing strategy in Iceberg. For a large enough table, there are N file group each rewritten independently. Records within a group are sorted globally into K files. Without sorting all rows has to be scanned, with sorting only rows in intersecting row groups of intersecting file splits need to scanned. This makes it a bit tricky to calculate gain which is # rows skipped = # rows total - # rows scanned. Using # files skipped as a proxy might be not feasible due to (*), # rows skipped is not related to # files.

I guess, roughly compaction reduces # files (file open as the result), sorting reduces # row scans.

When we have # row skips part, I think we can just do current gain (storage part) + W * (# row skips), we can learn W from prod operation. Now I feel kind of OK with the current formula we have for pure compaction, it's extensible to sorting use case as mentioned above.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/PartitionStat.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/** Represents table partition. */
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class PartitionStat implements Serializable {
+  private List<String> values;
```
sumedhsakdeo, line 16:
```suggestion
    private static final long serialVersionUID = 1L;

    private List<String> values;
```

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/PartitionStat.java
```diff
@@ -0,0 +1,18 @@
...
+  private int fileCount;
```
sumedhsakdeo, line 17: Add a 
```
@Override
    public String toString() {
        return "PartitionStat{" +
                "values=" + values +
                ", fileCount=" + fileCount +
                '}';
    }
```

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/PartitionStat.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/** Represents table partition. */
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class PartitionStat implements Serializable {
```
sumedhsakdeo, line 15:
```suggestion
public class TablePartitionStat implements Serializable {
```

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/PartitionStat.java
sumedhsakdeo, line 16: what does values signify? Add comment and try different variable name.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/PartitionStat.java
```diff
@@ -0,0 +1,18 @@
+package com.linkedin.openhouse.datalayout.datasource;
+
+import java.io.Serializable;
+import java.util.List;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/** Represents table partition. */
```
sumedhsakdeo, line 10:
```suggestion
/** Represents table partition stats. */
```

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/PartitionStat.java
sumedhsakdeo: For compatibility during deserialization.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/FileStat.java
```diff
@@ -10,7 +11,7 @@
 @Builder
 @NoArgsConstructor
 @AllArgsConstructor
-public final class FileStat {
+public final class FileStat implements Serializable {
```
sumedhsakdeo, line 14:
```suggestion
public final class FileStat implements Serializable {
    private static final long serialVersionUID = 1L;
```

sumedhsakdeo, line 14: Meta question given you made it Serializable where are we persisting FileStat. Are we persisting it for all files in a table? If so that can be a long list.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/PartitionStat.java
sumedhsakdeo, line 15: Meta question given you made it Serializable where are we persisting TablePartitionStat. Are we persisting it for all files in a table? If so that can be a long list.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/TableFileStats.java
```diff
@@ -16,7 +16,7 @@ public class TableFileStats implements DataSource<FileStat> {
   @Override
   public Dataset<FileStat> get() {
     return spark
-        .sql(String.format("SELECT file_path, file_size_in_bytes FROM %s.files", tableName))
+        .sql(String.format("SELECT file_path, file_size_in_bytes FROM %s.data_files", tableName))
```
sumedhsakdeo, line 19: I would imagine this code layout to be either

```
libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/TableFileDataSourceImpl.java
libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/TablePartitionDataSourceImpl.java
libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/dao/FileStat.java
libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/dao/PartitionStats.java
```
given the last two are part of your "spec". If there are just in-memory DTOs then replace `dao` with `dto`

sumedhsakdeo (changes requested):

sumedhsakdeo (commented):
Added review for datasource module

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+  private DataLayoutOptimizationStrategy generateCompactionStrategy() {
+    List<Long> fileSizes =
+        tableFileStats
+            .get()
+            .map((MapFunction<FileStat, Long>) FileStat::getSize, Encoders.LONG())
+            .filter(
+                (FilterFunction<Long>)
+                    size ->
+                        size < TARGET_BYTES_SIZE * DataCompactionConfig.MIN_BYTE_SIZE_RATIO_DEFAULT)
+            .collectAsList();
```
sumedhsakdeo, line 58: We are getting all the fileSizes in-memory and then computing sum() and count(). Does it make sense for us to delegate the computation for sum of fileSizes that are candidate for compaction to spark?
Cache it after computing it once using spark to be used later.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+                / estimatedTasksPerFileGroupCount));
+
+    // don't split large files
+    configBuilder.maxByteSizeRatio(MAX_BYTES_SIZE_RATIO);
+
+    long filesReducedCount = estimateReducedFilesCount(fileSizes);
+    double computeGbHr = estimateComputeGbHr(fileSizes);
+    double filesReducedCountPerComputeGbHr = filesReducedCount / computeGbHr;
+    return DataLayoutOptimizationStrategy.builder()
+        .config(configBuilder.build())
```
sumedhsakdeo, line 94:
```suggestion
        .type(PARTITIONED_TABLE_FILE_RIGHT_SIZING_STRATEGY)
        .config(configBuilder.build())
```

Let's introduce a strategy enum for example. 
```
enum {
     PARTITIONED_TABLE_FILE_RIGHT_SIZING_STRATEGY,
     PARTITIONED_TABLE_SORTING_STRATEGY //  (add in future)
}
```

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+  @Override
+  public List<DataLayoutOptimizationStrategy> generate() {
+    // skip single partition and non-partitioned tables
+    if (tablePartitionStats.get().count() <= 1) {
+      return Collections.emptyList();
+    }
+    return Collections.singletonList(generateCompactionStrategy());
+  }
+
+  private DataLayoutOptimizationStrategy generateCompactionStrategy() {
```
sumedhsakdeo, line 49:
```suggestion
  /**
   * Method to compute the file right sizing strategy for a partitioned table. In this method we compute number of files that are sized lesser than a threshold, compute predicted cost and predicted gain and finally computing a score for the strategy.
   * predicted cost is computed as the GBhr spent in performing the compaction.
   * predicted gain is computed as the file counts reduced.
   * predicted score is computed as gain / cost, the number of files reduced per GB-hr of compute. Higher the score, better is the strategy.
   */
  private DataLayoutOptimizationStrategy computePartitionedTableFileRightSizingStrategy() {
```

@anjagruenheid this is what Stas decided for cost, gain and score for file rightsizing. We will have a to manually iterate on threshold for score for picking a table for compaction. Further we will compare the predicted cost, gain, and score with actual cost, gain, and score and check how close they are. Please let us know what you think.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+            MAX_CONCURRENT_FILE_GROUP_REWRITES,
+            (estimatedTasksPerFileGroupCount + REWRITE_PARALLELISM - 1)
+                / estimatedTasksPerFileGroupCount));
+
+    // don't split large files
+    configBuilder.maxByteSizeRatio(MAX_BYTES_SIZE_RATIO);
+
+    long filesReducedCount = estimateReducedFilesCount(fileSizes);
+    double computeGbHr = estimateComputeGbHr(fileSizes);
+    double filesReducedCountPerComputeGbHr = filesReducedCount / computeGbHr;
```
sumedhsakdeo, line 92: handle divide by zero error

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+    long candidateFilesBytes = fileSizes.stream().reduce(0L, Long::sum);
+    long compactedFilesCount = (candidateFilesBytes + TARGET_BYTES_SIZE - 1) / TARGET_BYTES_SIZE;
+    return Math.max(0, fileSizes.size() - compactedFilesCount);
+  }
+
+  private double estimateComputeGbHr(List<Long> fileSizes) {
+    double candidateFilesBytes = fileSizes.stream().reduce(0L, Long::sum);
+    double rewriteSeconds = candidateFilesBytes / REWRITE_BYTES_PER_SECOND;
+    double rewriteHours = rewriteSeconds / 3600.0;
+    return rewriteHours * EXECUTOR_MEMORY_MB / 1024 + COMPUTE_STARTUP_COST_GB_HR;
```
sumedhsakdeo, line 111: @teamurko `EXECUTOR_MEMORY_MB / 1024` not following this part of the computation.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+        .build();
+  }
+
+  private long estimateReducedFilesCount(List<Long> fileSizes) {
+    long candidateFilesBytes = fileSizes.stream().reduce(0L, Long::sum);
+    long compactedFilesCount = (candidateFilesBytes + TARGET_BYTES_SIZE - 1) / TARGET_BYTES_SIZE;
+    return Math.max(0, fileSizes.size() - compactedFilesCount);
+  }
+
+  private double estimateComputeGbHr(List<Long> fileSizes) {
```
sumedhsakdeo, line 107: @shardulm94 can you please take a look at the math in this function where we are trying to compute the estimated GBhr for the compaction job that we will run?

```
to_hour(totalBytesToWriteAsPartOfFileRightSizing / rewrite_throughput_in_bytes_per_second)
```

sumedhsakdeo: Also, please confirm `COMPUTE_STARTUP_COST_GB_HR` from your observations. p50 should be a good start.

sumedhsakdeo (changes requested):
Reviewed the data layout generator. Thanks @teamurko

teamurko:
> > > I thought a bit more about the gain calculation and we can generally calculate it as follows: For each query, the gain (in terms of files touched), is the new number of files retrieved for the query subtracted from the original number of files. So assume that your underlying data is unsorted and the data is completely evenly distributed. Say you have 10 files, each containing values in col1 (0-100) and col2 (0-10) and for every query, we'd have to touch all current files because none of the files have tight bounds. Now you can compare it to the gain for 2 different strategies:
> > > 
> > > 1. Partitioned
> > >    For each query, the gain is the current number of files minus the files in each partition touched by the query. In summary, SUM_j #current files - (SUM_{P_i,j}(v_ij)) where j denotes all queries q_j and v_ij is either 0 if q_j does not touch partition P_i or the number of files if it touches the partition.
> > > 2. Sorted
> > >    For each query, the gain is the current number of files minus the subset of files that fall within the range of the query's values. I.e., SUM_j #current files - v_j where v_j denotes all files that are touched by query q_j when the underlying data is sorted.
> > > 
> > > To compute the gain, you'd need to a) create in-memory structures that represent the file layout and b) 'rerun' the queries of your workload against the structure, simulating how the new data layout would perform. For the former, you'll need to mimic how the data would get rearranged. So you'd read the data once and for example for partitioning, you maintain buckets for each partitioning key, counting their frequency and then splitting it according to how many records you'd expect per file. So if you have only 5 records/file at most and your partition has 6 records, then upper(6/5)=2 files.
> > > Also note that if the workload does not (substantially) change, you don't have to rerun this calculation, so it's a one-off effort for data layout selection. In contrast, data maintenance cost/benefit calculations have to be executed frequently to maintain the chosen data layout.
> > 
> > 
> > Thank you, Anja, for brainstorming ideas. I have been looking at it a bit more. Let me share some findings and ideas and we'll see how we can consolidate it all together and make it work for both compaction and query optimization.
> > My current understanding about partitioning vs sorting is that these are orthogonal. A table can be partitioned or not, if it's sorted and partitioned, sorting within partitions happens independently in Iceberg. WLOG, we can assume non-partitioned table.
> > I'm thinking about 2 aspects of query optimization: 1) # files opened; 2) # records scanned.
> > (*) For a splittable file format like orc, parquet with standard compression, Iceberg creates file split tasks. If we have 1 file 1gb with configurable split size 200mb, there will be 5 splits, or 5 tasks. If we have 10 files 100mb each, there will be still 5 splits, each split referencing 2 files. In the first case there will be 5 file opens, in second - 10. File split is a kind of abstraction from files.
> > If the file open cost is high and we have many small files, e.g. 1000 files of 1mb each, we'll pay (file open cost) * 1000 / (# executors), it's worse if there a skew such that most file opens happened in one task, this would delay query execution. Although sorting on its own could reduce # file opens, say if we have mostly non-intersecting small files, compaction is part of sorting rewrite strategy in Iceberg, so small files problem automatically goes away.
> > Compaction can't solve 2), generally all records in all file splits has to be scanned. Sorting works on top of bin packing strategy in Iceberg. For a large enough table, there are N file group each rewritten independently. Records within a group are sorted globally into K files. Without sorting all rows has to be scanned, with sorting only rows in intersecting row groups of intersecting file splits need to scanned. This makes it a bit tricky to calculate gain which is # rows skipped = # rows total - # rows scanned. Using # files skipped as a proxy might be not feasible due to (*), # rows skipped is not related to # files.
> > I guess, roughly compaction reduces # files (file open as the result), sorting reduces # row scans.
> > When we have # row skips part, I think we can just do current gain (storage part) + W * (# row skips), we can learn W from prod operation. Now I feel kind of OK with the current formula we have for pure compaction, it's extensible to sorting use case as mentioned above.
> 
> Makes sense, I wasn't completely sure which statistics you have available but it should be okay to compute either the number of rows skipped or the number of files skipped, depending on what's easier to start with. Essentially, you want to use a formula that computes the per-query gain and sum that up across all available queries. The main idea is to 'simulate' the partitioning/sorting/combination thereof with an intermediate, in-memory structure which allows you to compute/mimic the rows or files skipped. And it's super helpful if you can extend the current gain calculation to encompass this idea, that would be ideal actually! What do you think, is this feasible for the optimal data layout selection part?

I'd suggest not supporting query optimization part for now. We should do it after current compaction model is tuned to work.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/FileStat.java
teamurko: Thanks, I'm going to remove all serializable since we don't use default java object serialization scheme, Spark uses it's own, for strategies we use json

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/PartitionStat.java
teamurko: `PartitionStat(values=[2024-01-01, 0], fileCount=1)` @Data annotation generates toString

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/PartitionStat.java
teamurko: Added comment

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
teamurko: I didn't want to optimize prematurely, 8 bytes per file * 10m = 80MiB. After giving it more thought and experiencing recent memory tuning, I think it's worth computing in Spark, it's fast.

sumedhsakdeo: Yes I also have recency bias when I made that comment. :)

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
sumedhsakdeo: IMO, compaction is a very broad term so I was suggesting something very specific. One could argue that sorting improves compressions and hence compacts data.

teamurko: Compaction is Iceberg terminology for this operation https://iceberg.apache.org/docs/latest/maintenance/#compact-data-files, it become more like compaction when we consider delete files. Right-sizing has resource allocation relation, although it's more specific.

sumedhsakdeo: Fair. Let's do compaction

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
teamurko: Added comment

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
teamurko: Added comment.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
teamurko: FIxed.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/TablePartitionStats.java
```diff
@@ -0,0 +1,55 @@
...
+   * Get partition statistics dataset for the table.
+   *
+   * @return Dataset of partition statistics sorted by partition values.
+   */
+  @Override
+  public Dataset<PartitionStat> get() {
+    StructType partitionSchema =
+        spark.sql(String.format("SELECT * FROM %s.partitions", tableName)).schema();
+    try {
+      partitionSchema.apply("partition");
```
sumedhsakdeo, line 29: Where is this partitionSchema used?

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/TablePartitionStats.java
```diff
@@ -0,0 +1,55 @@
...
+      return spark
+          .sql(
+              String.format(
+                  "SELECT partition, file_count FROM %s.partitions ORDER BY partition", tableName))
```
sumedhsakdeo, line 33:
```suggestion
                  "SELECT partition, file_count FROM %s.partitions", tableName))
```
Not sure ordering is needed. If yes, method should be getOrdered() or something.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
sumedhsakdeo: It is making it into a persistent layer of tblproperty. Changing is later makes it a bit difficult.

sumedhsakdeo (dismissed):
Looking forward to experimenting this and describing scores. One thing I would still suggest is because you have introduced `DataLayoutOptimizationStrategy` it might be ok add type field in it to identify the "kind of" optimization so that we don't need to rewrite it. However, I am not going to block the PR on it in lieu of making progress on experimentation.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
teamurko: Thanks, Sumedh. When strategy list is updated with overwrite, no need to make older version compatible, the future app that reads the state can be disabled during rollout. I'm not sure if it should be type attribute or smth else, let's keep it open for now.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/TablePartitionStats.java
teamurko: it will raise IllegalArgumentException if the column 'partition' doesn't exist, a way to work with unpartitioned tables, partition value will be 'null' in that case

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/TablePartitionStats.java
teamurko: Removed sorting for now, generator could do sorting later if needed.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/PartitionStat.java
teamurko: Removed serializable b/c custom serde is used

teamurko: This class is used in table context always

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/PartitionStat.java
teamurko: FIxed

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/FileStat.java
teamurko: Removed serializable

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutGenerator.java
teamurko: we can start with 0.5 and tune later

sumedhsakdeo (approved):

PR #119, opened 2024-06-06
Introduce `AllocateTableStorage` in Storage interface 

HotSushi:
## Summary
We introduce a new method in the storage interface as follows:
```
Storage {
  String allocateTableSpace(...)
}
```
which returns the tablelocation where table objects will be created. 

Implementations of this api should also execute one time operations such as creating buckets, assigning permissions etc. 

Please note the comment: 
`this method should avoid creating TableFormat specific directories (ex: /data, /metadata for Iceberg or _delta_log for DeltaLake)`,it describes what **shouldn't** be done as part of this implementation.

## Layering with internal implementations
Internal implementations such as `LiHdfsStorage` will now have flexibiliity to do one-time operations such as table provisioning.

## Changes
- [X] New Features
- [X] Tests

## Testing Done

- [X] Existing tests run ✅ 
- [X] Gradle build succeeds ✅ 
- [X] E2E local tests work as well ✅

### DIFF
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
@@ -53,17 +53,11 @@ public abstract class BaseStorage implements Storage {
    * @param tableId the table id of the table
    * @param tableUUID the UUID of the table
    * @param tableCreator the creator of the table
-   * @param skipProvisioning Set to true if heavy-lifting allocation work needs to be skipped and
-   *     only the table location needs to be returned
    * @return the table location where the table data should be stored
    */
   @Override
   public String allocateTableLocation(
-      String databaseId,
-      String tableId,
-      String tableUUID,
-      String tableCreator,
-      boolean skipProvisioning) {
+      String databaseId, String tableId, String tableUUID, String tableCreator) {
     Preconditions.checkArgument(databaseId != null, "Database ID cannot be null");
     Preconditions.checkArgument(tableId != null, "Table ID cannot be null");
     Preconditions.checkArgument(tableUUID != null, "Table UUID cannot be null");
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
@@ -45,7 +45,7 @@ public abstract class BaseStorage implements Storage {
   }
 
   /**
-   * Allocates Table Space for the storage.
+   * Allocates Table Storage and return location.
    *
    * <p>Default tableLocation looks like: {endpoint}/{rootPrefix}/{databaseId}/{tableId}-{tableUUID}
    *
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
@@ -49,10 +49,16 @@ public abstract class BaseStorage implements Storage {
    *
    * <p>Default tableLocation looks like: {endpoint}/{rootPrefix}/{databaseId}/{tableId}-{tableUUID}
    *
+   * @param databaseId the database id of the table
+   * @param tableId the table id of the table
+   * @param tableUUID the UUID of the table
+   * @param tableCreator the creator of the table
+   * @param skipProvisioning Set to true if heavy-lifting allocation work needs to be skipped and
+   *     only the table location needs to be returned
    * @return the table location where the table data should be stored
    */
   @Override
-  public String allocateTableSpace(
+  public String allocateTableLocation(
       String databaseId,
       String tableId,
       String tableUUID,
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
@@ -1,6 +1,8 @@
 package com.linkedin.openhouse.cluster.storage;
 
+import com.google.common.base.Preconditions;
 import com.linkedin.openhouse.cluster.storage.configs.StorageProperties;
+import java.net.URI;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
@@ -40,5 +42,38 @@ public abstract class BaseStorage implements Storage {
         .map(StorageProperties.StorageTypeProperties::getParameters)
         .map(HashMap::new)
         .orElseGet(HashMap::new);
+  }
+
+  /**
+   * Allocates Table Space for the storage.
+   *
+   * <p>Default tableLocation looks like: {endpoint}/{rootPrefix}/{databaseId}/{tableId}-{tableUUID}
+   *
+   * @return the table location where the table data should be stored
+   */
+  @Override
+  public String allocateTableSpace(
+      String databaseId,
+      String tableId,
+      String tableUUID,
+      String tableCreator,
+      boolean skipProvisioning) {
+    Preconditions.checkArgument(databaseId != null, "Database ID cannot be null");
+    Preconditions.checkArgument(tableId != null, "Table ID cannot be null");
+    Preconditions.checkArgument(tableUUID != null, "Table UUID cannot be null");
+    Preconditions.checkState(
+        storageProperties.getTypes().containsKey(getType().getValue()),
+        "Storage properties doesn't contain type: " + getType().getValue());
+    return URI.create(
+            getClient().getEndpoint()
+                + getClient().getRootPrefix()
+                + "/"
+                + databaseId
+                + "/"
+                + tableId
+                + "-"
+                + tableUUID)
+        .normalize()
+        .toString();
   }
 }
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
@@ -48,4 +48,32 @@ public interface Storage {
    * @return a client to interact with the storage
    */
   StorageClient<?> getClient();
+
+  /**
+   * Allocates Table Space for the storage.
+   *
+   * <p>Allocating involves creating directory structure/ creating bucket, and setting appropriate
+   * permissions etc. Please note that this method should avoid creating TableFormat specific
+   * directories (ex: /data, /metadata for Iceberg or _delta_log for DeltaLake). Such provisioning
+   * should be done by the TableFormat implementation.
+   *
+   * <p>After allocation is done, this method should return the table location where the table data
+   * should be stored. Example: /rootPrefix/databaseId/tableId-UUID for HDFS storage
+   * /tmp/databaseId/tableId-UUID for Local storage s3://bucket/databaseId/tableId-UUID for S3
+   * storage
+   *
+   * @param databaseId the database id of the table
+   * @param tableId the table id of the table
+   * @param tableUUID the UUID of the table
+   * @param tableCreator the creator of the table
+   * @param skipProvisioning Set to true if heavy-lifting allocation work needs to be skipped and
+   *     only the table location needs to be returned
+   * @return the table location after provisioning is done
+   */
+  String allocateTableSpace(
+      String databaseId,
+      String tableId,
+      String tableUUID,
+      String tableCreator,
+      boolean skipProvisioning);
 }
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
@@ -58,8 +58,8 @@ public interface Storage {
    * should be done by the TableFormat implementation.
    *
    * <p>After allocation is done, this method should return the table location where the table data
-   * should be stored. Example: /rootPrefix/databaseId/tableId-UUID for HDFS storage;
-   * /tmp/databaseId/tableId-UUID for Local storage; s3://bucket/databaseId/tableId-UUID for S3
+   * should be stored. Example: hdfs:///rootPrefix/databaseId/tableId-UUID for HDFS storage;
+   * file:/tmp/databaseId/tableId-UUID for Local storage; s3://bucket/databaseId/tableId-UUID for S3
    * storage
    *
    * @param databaseId the database id of the table
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
@@ -50,7 +50,7 @@ public interface Storage {
   StorageClient<?> getClient();
 
   /**
-   * Allocates Table Space for the storage.
+   * Allocates Table Storage and return location.
    *
    * <p>Allocating involves creating directory structure/ creating bucket, and setting appropriate
    * permissions etc. Please note that this method should avoid creating TableFormat specific
@@ -70,7 +70,7 @@ public interface Storage {
    *     only the table location needs to be returned
    * @return the table location after provisioning is done
    */
-  String allocateTableSpace(
+  String allocateTableLocation(
       String databaseId,
       String tableId,
       String tableUUID,
--- a/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
+++ b/cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
@@ -66,14 +66,8 @@ public interface Storage {
    * @param tableId the table id of the table
    * @param tableUUID the UUID of the table
    * @param tableCreator the creator of the table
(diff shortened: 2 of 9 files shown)


HotSushi:
cc: @sumedhsakdeo , @jainlavina , @ctrezzo

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
```diff
@@ -41,4 +43,37 @@ public Map<String, String> getProperties() {
...
+   *
+   * @return the table location where the table data should be stored
+   */
+  @Override
+  public String allocateTableSpace(
+      String databaseId,
+      String tableId,
+      String tableUUID,
+      String tableCreator,
+      boolean skipProvisioning) {
```
sumedhsakdeo, line 60: what is skipProvisioning? javadoc missing too for params

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/Storage.java
```diff
@@ -48,4 +48,32 @@ public interface Storage {
...
+   * should be stored. Example: /rootPrefix/databaseId/tableId-UUID for HDFS storage;
+   * /tmp/databaseId/tableId-UUID for Local storage; s3://bucket/databaseId/tableId-UUID for S3
+   * storage
+   *
+   * @param databaseId the database id of the table
+   * @param tableId the table id of the table
+   * @param tableUUID the UUID of the table
+   * @param tableCreator the creator of the table
+   * @param skipProvisioning Set to true if heavy-lifting allocation work needs to be skipped and
+   *     only the table location needs to be returned
```
sumedhsakdeo, line 70: Still not clear what this means.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
```diff
@@ -41,4 +43,37 @@ public Map<String, String> getProperties() {
...
+
+  /**
+   * Allocates Table Space for the storage.
+   *
+   * <p>Default tableLocation looks like: {endpoint}/{rootPrefix}/{databaseId}/{tableId}-{tableUUID}
+   *
+   * @return the table location where the table data should be stored
+   */
+  @Override
+  public String allocateTableSpace(
```
sumedhsakdeo, line 55:
```suggestion
  public String allocateTableStorage(
```

sumedhsakdeo (commented):
code looks good, had one question about _skipProvisioning_

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
abhisheknath2011, line 55: Can you update the java docs to reflect the params?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorage.java
```diff
@@ -36,4 +37,22 @@ public StorageType.Type getType() {
...
+   * @return the table location
+   */
+  @Override
+  public String allocateTableSpace(
+      String databaseId,
+      String tableId,
+      String tableUUID,
+      String tableCreator,
+      boolean skipProvisioning) {
+    return Paths.get(getClient().getRootPrefix(), databaseId, tableId + "-" + tableUUID).toString();
```
abhisheknath2011, line 56: The BasedStorage implementation did not use `java.nio.file.Paths`. So why it has to be done differently?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
HotSushi: skipProvisioning is a way to get TableLocation without the allocation logic, its code will look like

```
String allocateTableSpace() {
   if (!skipProvisioning) {
        allocateBucket()
        setPermission()
        createSubdirectories()
   }
   return tableLocation;
}
```

We can use `allocateTableSpace(skipprovisioning=true)` to quickly get the table location. The use case for this is CTAS, when we try to match tableLocation from the `putSnapshot` call without rerunning the logic.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
HotSushi: ack! avoided repeating documentation from the `Storage` API class.

HotSushi: Will add.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorage.java
HotSushi: Good observation, we do this to ensure backward-compatibility with current implementation. the difference with baseStorage class is the no-endpoint, basically baseStorage gives `hdfs://data/openhouse` whereas the override here gives `/data/openhouse`

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
```diff
@@ -41,4 +43,37 @@ public Map<String, String> getProperties() {
...
+      String tableCreator,
+      boolean skipProvisioning) {
+    Preconditions.checkArgument(databaseId != null, "Database ID cannot be null");
+    Preconditions.checkArgument(tableId != null, "Table ID cannot be null");
+    Preconditions.checkArgument(tableUUID != null, "Table UUID cannot be null");
+    Preconditions.checkState(
+        storageProperties.getTypes().containsKey(getType().getValue()),
+        "Storage properties doesn't contain type: " + getType().getValue());
+    return URI.create(
+            getClient().getEndpoint()
```
ctrezzo, line 74: How does URI.create or normalization handle situations where you have double slashes? For example, if the following endpoint and rootPrefix were provided:
Endpoint: hdfs://localhost:8080/
Rootprefix: /userFoo/

This would pass in the string: "hdfs://localhost:8080//userFoo//databaseId/tableId-tableUUID"

Are we relying on URI.create to do the validation of the user supplied endpoint and rootPrefix? If so, we might want to add some unit tests covering what we expect the output string to be.

In a perfect world, the following configurations would all produce the same location (i.e. the strings produced by this method are equal). I am not sure if these are all covered by normalization:

Config1:
Endpoint: hdfs://localhost:8080/
Rootprefix: /userFoo/

Config2:
Endpoint: hdfs://localhost:8080
Rootprefix: /userFoo

Config3:
Endpoint: hdfs://localhost:8080
Rootprefix: /userFoo/

Config4:
Endpoint: hdfs://localhost:8080/
Rootprefix: /userFoo

Also, is this considered an invalid configuration?
Config5:
Endpoint: hdfs://localhost:8080/
Rootprefix: userFoo

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorage.java
ctrezzo, line 56: This concatenation logic could also go into the TableLocation class. Why do we not need an endpoint for HDFS and local filesystem? Is this because we are relying on default filesystem in configuration? It seems like we want to be consistent here at the Storage interface level and state that the table location returned is either absolute or relative. I am guessing we probably want the former.

Also wondering if we would want to use URI.resolve instead of Paths class to keep consistency? Paths class assumes a default filesystem. I am not entirely sure of the complete behavior of URI.resolve with two relative URIs though.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/InternalRepositoryUtils.java
```diff
@@ -142,8 +143,9 @@ static TableDto convertToTableDto(
             .tableUUID(megaProps.get(getCanonicalFieldName("tableUUID")))
             .tableLocation(
                 URI.create(
```
ctrezzo, line 145: Should probably encapsulate this logic in the same place as the other concatenation logic (i.e. TableLocation class).

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
ctrezzo: +1 I like the idea of "location" as a term.

ctrezzo (commented):
Thank you for the work!

I left some comments mainly centered around how we encapsulate the table location manipulation logic.

I have scar tissue from the MapReduce code and converting between Paths, URIs, URLs and having that logic scattered everywhere, so sorry in advance if I am focused too much on this.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
```diff
@@ -41,4 +43,37 @@ public Map<String, String> getProperties() {
...
+      String tableUUID,
+      String tableCreator,
+      boolean skipProvisioning) {
+    Preconditions.checkArgument(databaseId != null, "Database ID cannot be null");
+    Preconditions.checkArgument(tableId != null, "Table ID cannot be null");
+    Preconditions.checkArgument(tableUUID != null, "Table UUID cannot be null");
+    Preconditions.checkState(
+        storageProperties.getTypes().containsKey(getType().getValue()),
+        "Storage properties doesn't contain type: " + getType().getValue());
+    return URI.create(
```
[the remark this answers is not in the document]
HotSushi: Thanks Chris! would like to understand this more, is this accurate impl of your comment?
```
TableLocation allocateStorage(...){
        return getTableLocation();
}
TableLocation getTableLocation(...)
```

----------------------
If so, there were couple of reasons I chose the current design.

- You are correct about potential of TableLocation being a common place for validation logic. But besides that, I'm unable to see the value it might bring in prod code, ie. There won't be any place where we do `tableLocation.getXYZ`. The end state is always going to be `tableLocation.toString` because of iceberg layer. If the missing piece is about validations , its up to each storage to validate if the URL is correct, which can be done in allocateStorageSpace itself.
- it's possible to split into two methods: `allocateStorage`, `getTableLocation`. The reason I chose a single method was to avoid the split and centralize the whole logic into a single method. This way we avoid `getTableLocation` as a utility method and the usage of `allocateStorage` is limited during tableCreation time (either CT or CTAS). Let me know if you feel strongly and we can discuss more.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
HotSushi: Nice test cases! let me add them. I think current implementation should yield same result for all of them. 

 Normalize removes double slashes besides the scheme. 
Config 5 seem valid and likely what we should use for S3. RootPrefix should just be bucketName.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorage.java
HotSushi: > Why do we not need an endpoint for HDFS and local filesystem? Is this because we are relying on default filesystem in configuration? It seems like we want to be consistent here at the Storage interface level and state that the table location returned is either absolute or relative. I am guessing we probably want the former.

We are doing this because of backward compatibliity reasons. Ideally we go to a state towards absolute but because this code gets used in li-openhouse we've to be careful. 
I plan to do as follows:
- openhouse: /db/tb, li-openhouse: <missing Li-HDFSStorage>
- openhouse: /db/tb, li-openhouse: /db/tb
- openhouse: hdfs://db/tb,  li-openhouse: db/tbl

>Also wondering if we would want to use URI.resolve instead of Paths class to keep consistency? Paths class assumes a default filesystem. I am not entirely sure of the complete behavior of URI.resolve with two relative URIs though.

Ack, let me give it a shot. I kept the implementation backward compatible. I do see a value in just aligning towards URI.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/InternalRepositoryUtils.java
HotSushi: Because this return value is coming from Iceberg, it's in string format. Even if we add a TableLocation class, we'll then need to define `TableLocation.parse(String)`. String seems simpler and easier to reason with **For now**. Can re-consider if we see tablelocation usage proliferate.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
sumedhsakdeo: So the name of the function is allocate - table - location and it has a flag that says skip provisioning in which case the function acts like a get - table - location? Something is off. Is there another way to support CTAS?

HotSushi: @ctrezzo  had a similar feedback of splitting the methods [into allocateTableStorage, getTableLocation](https://github.com/linkedin/openhouse/pull/119#discussion_r1630366741). I can adopt this feedback as its common.  My reasoning was:
> it's possible to split into two methods: allocateStorage, getTableLocation. The reason I chose a single method was to avoid the split and centralize the whole logic into a single method. This way we avoid getTableLocation as a utility method and the usage of allocateStorage is limited during tableCreation time (either CT or CTAS). Let me know if you feel strongly and we can discuss more.

Let me know what you think. I'm not sure if we can support CTAS in a different way in the short term.

sumedhsakdeo: So we want the createTableStartTransaction in CTAS to call this method with all the below steps.

- allocateBucket() 
- setPermission() 
- createSubdirectories()

We do not want the createTableCommitTransaction in CTAS to not call this method allocateTableSpace because it was already called as part of createTableStartTransaction.

Shouldn't the caller check, if already_staged = true which would be the case for createTableCommitTransaction do NOT call allocateTableSpace?

Why would we need getTableLocation as the additional method here?

HotSushi: Thanks for the comments Sumedh, its clear the parameter is adding confusion right now, we'll introduce it back when we implement CTAS. 

to answer your questions,

- createTableStartTransaction in CTAS will not call this method to allocateBucket() etc. instead it will call it toGetTablePath so that it can do comparison like `metadata_json_path.isPrefix(allocateTableStorage(skipAllocation=true))`

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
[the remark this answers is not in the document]
HotSushi: Addressed offline, main concern was about centralizing logic for concatenating strings and not let that logic proliferate. 
The centralize logic exists today and that's in "BaseStorage::AllocateStorage()". The only reason we override this in HdfsStorage, LocalStorage was because of https://github.com/linkedin/openhouse/issues/121

As for `TableLocation`  class as an abstraction, we decided to not proceed right now because String returned by allocateTableStorage is treated as immutable and would only be handed off to  iceberg as string.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
HotSushi: Added tests and comments like discussed. This one is not the underlying assumption is endpoint+rootprefix should be valid.

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorage.java
HotSushi: Discussed offline, summary: https://github.com/linkedin/openhouse/pull/119#discussion_r1631640467

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/InternalRepositoryUtils.java
HotSushi: Discussed offline, see summary: https://github.com/linkedin/openhouse/pull/119#discussion_r1631640467

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
sumedhsakdeo: Yes, let's discuss it when we reach use-case where we need to support stage creation of table directory at the start of the transaction. Otherwise lgtm, do wait for @ctrezzo to approve this as he has a lot of good comments.

ctrezzo (approved):
Looks good to me. Thank you for providing the extra clarifying comments around issue #121 . I think it makes sense that the logic will be contained in BaseStorage once this issue is fixed and the refactor around allocateTableLocation is complete.

sumedhsakdeo (approved):

Issue #121, opened 2024-06-07
[Feature] HdfsStorage and LocalStorage's TableLocation should be full URI with scheme


HotSushi:
### Willingness to contribute

{"label"=>"Yes. I can contribute a fix for this bug independently.", "value"=>"independent"}

### Feature Request Proposal

Currently HDFSStorage and LocalStorage set TableLocation without scheme, ie: `/data/openhouse/db/tbl-UUID` instead of full URI `hdfs:///data/openhouse/db/tbl-UUID`.

This was done to ensure backward compatibility, replication and migration.

Iceberg spec mandates a full-URI (containing scheme) as part of its metadata. source: https://iceberg.apache.org/spec/?ref=blog.min.io#manifests

As part of this bugFix, HDFSStorage and LocalStorage should start using fully-qualified URI.

### Motivation

> #### What is the use case for this feature?
Making this change will generalize tableLocation behavior for various Storages such as s3, hdfs, local etc.

### What component(s) does this feature affect?

- [X] `Table Service`: This is the RESTful catalog service that stores table metadata. `:services:tables`
- [ ] `Jobs Service`: This is the job orchestrator that submits data services for table maintenance. `:services:jobs`
- [ ] `Data Services`: This is the jobs that performs table maintenance. `apps:spark`
- [X] `Iceberg internal catalog`: This is the internal Iceberg catalog for OpenHouse Catalog Service. `:iceberg:openhouse`
- [ ] `Spark Client Integration`: This is the Apache Spark integration for OpenHouse catalog. `:integration:spark`
- [ ] `Documentation`: This is the documentation for OpenHouse. `docs`
- [ ] `Local Docker`: This is the local Docker environment for OpenHouse. `infra/recipes/docker-compose`
- [ ] `Other`: Please specify the component.

PR #134, opened 2024-07-11
Add schema to table location

kmcclenn:
## Summary

[Issue](https://github.com/linkedin/openhouse/issues/121) Changed table locations to include the scheme.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [x] Refactoring
- [ ] Documentation
- [x] Tests

Changed both the implementation of creating the table location (and dealing with its ramifications) and the tests that test the table location. Allows all table locations to include the schema (eg. `hdfs://tmp/...` instead of `tmp/...`). Furthermore, added this fix for the old `constructTablePath` method and the new `allocateTableStorage` method. Removed outdated overriding methods for Local and HDFS storage that don't include the scheme.

## Testing Done

Changed tests to reflect the table locations now containing the schema.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
@@ -235,7 +235,8 @@ public final class Operations implements AutoCloseable {
   }
 
   private Path getTrashPath(String path, String filePath, String trashDir) {
-    return new Path(filePath.replace(path, new Path(path, trashDir).toString()));
+    return new Path(
+        filePath.replace(new Path(path).toString(), new Path(path, trashDir).toString()));
   }
 
   /**
--- a/apps/spark/src/test/java/com/linkedin/openhouse/jobs/spark/OperationsTest.java
+++ b/apps/spark/src/test/java/com/linkedin/openhouse/jobs/spark/OperationsTest.java
@@ -9,6 +9,7 @@ import com.linkedin.openhouse.tables.client.model.Policies;
 import com.linkedin.openhouse.tables.client.model.Retention;
 import com.linkedin.openhouse.tablestest.OpenHouseSparkITest;
 import io.opentelemetry.api.metrics.Meter;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -186,15 +187,21 @@ public class OperationsTest extends OpenHouseSparkITest {
       DeleteOrphanFiles.Result result =
           ops.deleteOrphanFiles(table, TRASH_DIR, System.currentTimeMillis(), false);
       List<String> orphanFiles = Lists.newArrayList(result.orphanFileLocations().iterator());
-      log.info("Detected {} orphan files", orphanFiles.size());
-      for (String of : orphanFiles) {
+      List<String> orphanFilesWithoutMetadata =
+          orphanFiles.stream()
+              .filter(f -> !f.endsWith("metadata.json"))
+              .collect(Collectors.toList());
+      log.info("Detected {} orphan files (without metadata)", orphanFilesWithoutMetadata.size());
+      for (String of : orphanFilesWithoutMetadata) {
         log.info("File {}", of);
       }
       Assertions.assertTrue(
           fs.exists(new Path(table.location(), new Path(TRASH_DIR, testOrphanFileName))));
-      Assertions.assertEquals(1, orphanFiles.size());
+      Assertions.assertEquals(1, orphanFilesWithoutMetadata.size());
       Assertions.assertTrue(
-          orphanFiles.get(0).endsWith(table.location() + "/" + testOrphanFileName));
+          orphanFilesWithoutMetadata
+              .get(0)
+              .endsWith(Paths.get(table.location(), testOrphanFileName).toString()));
       Assertions.assertFalse(fs.exists(orphanFilePath));
     }
   }
@@ -226,8 +233,13 @@ public class OperationsTest extends OpenHouseSparkITest {
       DeleteOrphanFiles.Result result2 =
           ops.deleteOrphanFiles(table, TRASH_DIR, System.currentTimeMillis(), false);
       List<String> orphanFiles2 = Lists.newArrayList(result2.orphanFileLocations().iterator());
-      log.info("Detected {} orphan files", orphanFiles2.size());
-      Assertions.assertEquals(0, orphanFiles2.size());
+      List<String> orphanFilesWithoutMetadata =
+          orphanFiles2.stream()
+              .filter(f -> !f.endsWith("metadata.json"))
+              .collect(Collectors.toList());
+      log.info(
+          "Detected {} orphan files (excluding metadata files)", orphanFilesWithoutMetadata.size());
+      Assertions.assertEquals(0, orphanFilesWithoutMetadata.size());
       Assertions.assertTrue(fs.exists(trashFilePath));
     }
   }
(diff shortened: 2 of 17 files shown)


apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
```diff
@@ -235,7 +235,8 @@ public void runRetention(
   }
 
   private Path getTrashPath(String path, String filePath, String trashDir) {
-    return new Path(filePath.replace(path, new Path(path, trashDir).toString()));
+    return new Path(
+        filePath.replace(new Path(path).toString(), new Path(path, trashDir).toString()));
```
HotSushi, line 239: why did we need this change?

apps/spark/src/test/java/com/linkedin/openhouse/jobs/spark/OperationsTest.java
```diff
@@ -186,15 +187,21 @@ public void testOrphanFilesDeletionJavaAPI() throws Exception {
       DeleteOrphanFiles.Result result =
           ops.deleteOrphanFiles(table, TRASH_DIR, System.currentTimeMillis(), false);
       List<String> orphanFiles = Lists.newArrayList(result.orphanFileLocations().iterator());
-      log.info("Detected {} orphan files", orphanFiles.size());
-      for (String of : orphanFiles) {
+      List<String> orphanFilesWithoutMetadata =
+          orphanFiles.stream()
```
HotSushi, line 191: why metadata.json files filtered out?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
```diff
@@ -60,10 +60,6 @@ public String allocateTableLocation(
       String databaseId, String tableId, String tableUUID, String tableCreator) {
     Preconditions.checkArgument(databaseId != null, "Database ID cannot be null");
     Preconditions.checkArgument(tableId != null, "Table ID cannot be null");
-    Preconditions.checkArgument(tableUUID != null, "Table UUID cannot be null");
```
HotSushi, line 63: this change seem regressive, its valid to ensure tableUUID is not null/ storageproperties is correctly configured

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/HdfsStorage.java
```diff
@@ -37,19 +36,4 @@ public StorageType.Type getType() {
...
-   * Allocates Table Space for the HDFS storage.
-   *
-   * <p>tableLocation looks like: /{rootPrefix}/{databaseId}/{tableId}-{tableUUID} We strip the
-   * endpoint to ensure backward-compatibility. This override should be removed after resolving <a
-   * href="https://github.com/linkedin/openhouse/issues/121">
-   *
-   * @return the table location
-   */
-  @Override
-  public String allocateTableLocation(
```
HotSushi, line 51: we ought to be careful with this. The reason we didn't make this change earlier was to ensure backward compatibilty. I would hold off checking this in until we keep the behavior compatible in [li-openhouse](https://github.com/linkedin-multiproduct/li-openhouse/blob/master/services/tables-extensions/src/main/java/com/linkedin/openhouse/cluster/storage/hdfs/LiHdfsStorage.java).  Could you make a change there to not call `super.allocateTableStorage`?

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/local/LocalStorage.java
```diff
@@ -52,17 +51,4 @@ public StorageType.Type getType() {
...
-
-  /**
-   * Allocates Table Space for the Local Storage.
-   *
-   * <p>tableLocation looks like: /{rootPrefix}/{databaseId}/{tableId}-{tableUUID} We strip the
-   * endpoint to ensure backward-compatibility. This override should be removed after resolving <a
-   * href="https://github.com/linkedin/openhouse/issues/121">
-   */
-  @Override
-  public String allocateTableLocation(
```
HotSushi, line 64: we have to be careful with this change as well. tables-test-fixtures uses non-schemed paths. Please validate tables-test-fixtures continue to work after this change cc: @jiang95-dev

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/InternalRepositoryUtils.java
```diff
@@ -35,14 +34,19 @@ private InternalRepositoryUtils() {
...
-  public static java.nio.file.Path constructTablePath(
+  public static URI constructTablePath(
       StorageManager storageManager, String databaseID, String tableId, String tableUUID) {
     // TODO: Default storage is used here. Support for non-default storage type per table needs to
     // be added.
-    return Paths.get(
-        storageManager.getDefaultStorage().getClient().getRootPrefix(),
-        databaseID,
-        tableId + "-" + tableUUID);
+    String endpoint = storageManager.getDefaultStorage().getClient().getEndpoint();
```
HotSushi, line 41: There's no need to make this change yet. constructTablePath is only getting used in CTAS (and only for hdfs). non-hdfs storages need much more things to correctly support ctas: BDP-23723

services/tables/src/main/java/com/linkedin/openhouse/tables/utils/TableUUIDGenerator.java
```diff
@@ -144,8 +145,10 @@ private void validatePathOfProvidedRequest(
         extractFromTblPropsIfExists(databaseId + "." + tableId, tableProperties, TBL_RAW_KEY);
 
     java.nio.file.Path previousPath =
-        InternalRepositoryUtils.constructTablePath(
-            storageManager, dbIdFromProps, tblIdFromProps, tableUUIDProperty);
+        Paths.get(
```
HotSushi, line 148: lets avoid making change to this class. CTAS support is not present for non-hdfs storages, and this needs to be solved thoughtfully: BDP-23723

see `doesPathExist(..) UnsupportedOperartionException`

HotSushi (changes requested):
Thanks for the PR @kmcclenn I'm curious about the necessity of this change. There are some prereqs before we tackle https://github.com/linkedin/openhouse/issues/121 .

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
kmcclenn: With the schema, the string `path` was `file:///tmp/...`. However, we were trying to replace it with something like `file:/tmp/...`, because the `Path` type only allows for single forward slashes. So, I converted it to a `Path` to standardize the format to `file:/tmp/...` so they would match.

apps/spark/src/test/java/com/linkedin/openhouse/jobs/spark/OperationsTest.java
kmcclenn: Seems that the tests were meant for locations without schema: [See this slack message from Stas](https://linkedin-randd.slack.com/archives/C02Q94UU6TE/p1720562662594949?thread_ts=1720556484.877769&cid=C02Q94UU6TE)

cluster/storage/src/main/java/com/linkedin/openhouse/cluster/storage/BaseStorage.java
kmcclenn: Yes, I was a bit confused here. I removed it because it was failing [this](https://github.com/linkedin/openhouse/blob/5ba28ac94bfb064bb00314e75ba67655b431bed4/services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/RepositoryTest.java#L331) test because the `STAGED_TABLE_DTO` did not have a `UUID`; there is probably a better solution here but I wasn't sure what was best

kmcclenn:
@HotSushi Thanks for the comments. I originally started the PR because the ADLS storage was failing without schema in the table location. I added changes for `constructTablePath` and all of the related tests before I realized that this was changed in `main` by the `allocateTableLocation` function. So it may not be needed anymore, especially if I will be breaking the backwards compatibility with the PR. Do you think I should just close it and move forward with `allocateTableLocation` for ADLS?

HotSushi:
> @HotSushi Thanks for the comments. I originally started the PR because the ADLS storage was failing without schema in the table location. I added changes for `constructTablePath` and all of the related tests before I realized that this was changed in `main` by the `allocateTableLocation` function. So it may not be needed anymore, especially if I will be breaking the backwards compatibility with the PR. Do you think I should just close it and move forward with `allocateTableLocation` for ADLS?

yes `allocateTableLocation` is the way to go about it. We can pursue this PR later once backward compatibility is ensured.

jiang95-dev:
Thanks @kmcclenn. You can close this pr if `allocateTableLocation` solves the problem.
Thanks @HotSushi. I have some questions about the design and plan:
1. Are we going to remove `allocateTableLocation` once the backward compatibility is ensured? I feel it is still needed to parse the location whether or not it is with scheme?
2. What will be the state for the backward compatibility? Old tables storing non-schemed tableLocation, and new tables storing schemed tableLocation, and OH server can read/write both correctly?

PR #146, opened 2024-07-30
New metric dimension "client_name" propagated if passed by API clients via "X-Client-Name" header

cbb330:
## Summary
problem: we cannot view metrics such as http_request_seconds_count per each type of client (e.g. trino) because there is no dimension for it

Solution: http client header set by API client which tells us which client is requesting the API. we use this header to tag the metrics and then can query the metrics via /metrics endpoint.

the http header is settable via:
- passing key in property map to OpenHouseCatalog
- directly to ApiClient init with ApiClient.addDefaultHeader

see proof: 
image

In addition, I also exposed our favicon.ico because it was spamming our logs/metrics with NoHandlerFoundException because the favicon was missing and also required authorization. 
see the error that was removed:
```bash
openhouse-tables-1 | {"startTimestamp":"1722276154615","endTimestamp":"1722276154617","sessionId":null,"clusterName":"LocalFSCluster","user":"undefined","serviceName":"UNRECOGNIZED","method":"GET","uri":"/favicon.ico","requestPayload":null,"statusCode":400,"responseErrorMessage":"The combination of the method [GET] and Path [uri=/favicon.ico] cannot be resolved by server. Please check and making sure you are using the right version of API.","stacktrace":"org.springframework.web.servlet.NoHandlerFoundException: No handler found for GET /favicon.ico\n\tat org.springframework.web.servlet.DispatcherServlet.noHandlerFound(DispatcherServlet.java:1287)\n\tat org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1050)\n\tat org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:965)\n\tat org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1006)\n\tat
...
...
```

pull request in 2-3 lines.

## Changes

- [x] Internal API Changes
- [x] Bug Fixes
- [x] New Features
- [ ] Code Style
- [x] Refactoring
- [ ] Documentation
- [x] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
```diff
@@ -39,7 +41,9 @@ public abstract class WebClientFactory {
   // Set the default HttpConnectionStrategy as POOLED connection
   private HttpConnectionStrategy strategy = HttpConnectionStrategy.POOLED;
 
-  private String sessionId = null;
+  @Setter private String sessionId = null;
```
cbb330, line 44: this had a custom setter before, using lombok instead now to automate.

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
```diff
@@ -24,6 +24,8 @@ public abstract class WebClientFactory {
   private static final String HTTP = "http";
   private static final String HTTPS = "https";
   private static final String SESSION_ID = "session-id";
+
+  private static final String HTTPHEADER_CLIENT_NAME = "X-Client-Name";
```
cbb330, line 28: unfortunately can't use constant from services.common.Constants, so defining this string key twice in this PR.

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
```diff
@@ -61,13 +65,17 @@ protected WebClient createWebClient(
       @NonNull String baseUrl, String token, String truststoreLocation)
       throws MalformedURLException, SSLException {
     String transportProtocol = getTransportProtocol(baseUrl);
+    HttpClient httpClient = null;
     if (HTTPS.equals(transportProtocol)) {
-      return createSecureWebClient(baseUrl, truststoreLocation);
+      httpClient = createSecureHttpClient(truststoreLocation);
```
cbb330, line 70: small refactor to deduplicate very similar methods

instead of building WebClient, just get HttpClient and pass to shared method for getting webClient

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
```diff
@@ -154,14 +154,16 @@ private WebClient createSecureWebClient(String baseUrl, String truststoreLocatio
...
     }
 
-    ClientHttpConnector connector = new ReactorClientHttpConnector(client);
+    return client;
+  }
+
+  private WebClient.Builder getWebClientBuilder(String baseUrl) {
     WebClient.Builder webClientBuilder = createWebClientBuilder();
     setSessionIdInWebClientHeader(webClientBuilder);
+    setClientNameInWebClientHeader(webClientBuilder);
```
cbb330, line 163: this is new: will take clientName method and pass to defaultHeader of WebClient.

WebClient will automagically pass defaultHeaders to all http requests  (or so I read on documentation, unfortunately i could not find a way to test this well)

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
```diff
@@ -182,16 +184,6 @@ public void setStrategy(HttpConnectionStrategy httpConnectionStrategy) {
       strategy = httpConnectionStrategy;
     }
   }
-
-  /**
-   * Set sessionId
-   *
-   * @param sessionId
-   */
-  public void setSessionId(String sessionId) {
```
cbb330, line 191: setter removed and replaced with lombok @Setter annotation

client/secureclient/src/test/java/com/linkedin/openhouse/client/ssl/TablesApiClientFactoryTest.java
```diff
@@ -8,13 +8,34 @@
...
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mockito;
 
 public class TablesApiClientFactoryTest {
 
+  static File tmpCert;
+
+  static TablesApiClientFactory tablesApiClientFactorySpy;
+
+  @BeforeAll
```
cbb330, line 25: taking these methods from the single test below and bringing them up to shared methods because now two unittests use these mocks/spys.

client/secureclient/src/test/java/com/linkedin/openhouse/client/ssl/TablesApiClientFactoryTest.java
```diff
@@ -63,20 +82,27 @@ public void testWebClientForHttpUrl() throws Exception {
...
         .createSslContext(truststoreLocationCapture.capture());
     assertNotNull(apiClient);
     assertEquals(tmpCert.getAbsolutePath(), truststoreLocationCapture.getValue());
   }
+
+  @Test
+  public void testSetClientNameCalled() throws Exception {
+    ArgumentCaptor<String> clientNameCapture = ArgumentCaptor.forClass(String.class);
+
+    tablesApiClientFactorySpy.setClientName("trino");
```
cbb330, line 101: small test to verify that when setClientName is called, apiClient is created, that the default headers of webclient (the core dependency of of apiclient) are set

infra/recipes/docker-compose/common/debug-profile/housetables.yml
```diff
@@ -3,4 +3,4 @@ services:
     ports:
       - 6005:6005
     environment:
-      - JAVA_TOOL_OPTIONS=-agentlib:jdwp=transport=dt_socket,server=y,address=6005,suspend=n
\ No newline at end of file
+      - JAVA_TOOL_OPTIONS=-agentlib:jdwp=transport=dt_socket,server=y,address=*:6005,suspend=n
```
cbb330, line 6: these are required to enable debugging on my machine, i'm not sure why no one else needs them, but i'm going ahead and adding them for everyone's assistance.

integrations/java/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseCatalog.java
```diff
@@ -106,13 +108,15 @@ public void initialize(String name, Map<String, String> properties) {
     String truststore = properties.getOrDefault(TRUST_STORE, "");
     String token = properties.getOrDefault(AUTH_TOKEN, null);
     String httpConnectionStrategy = properties.getOrDefault(HTTP_CONNECTION_STRATEGY, null);
+    String clientName = properties.getOrDefault(CLIENT_NAME, null);
```
cbb330, line 111: if properties map includes CLIENT_NAME, e.g. here: https://github.com/linkedin-multiproduct/trino-plugins/blob/a48ca1ac698426ce216eb93a03ee1f79bc7a945a/trino-metacat/src/main/java/com/linkedin/trino/metacat/OpenHouseCatalogClient.java#L220

then we will pass them to the tableApiClient

services/common/src/main/java/com/linkedin/openhouse/common/Constants.java
```diff
@@ -0,0 +1,16 @@
+package com.linkedin.openhouse.common;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class Constants {
```
cbb330, line 6: shared constants for use in logic and tests

services/tables/src/main/java/com/linkedin/openhouse/tables/config/MainApplicationConfig.java
```diff
@@ -74,6 +81,32 @@ public UserTableApi provideApiInstance() {
     return new UserTableApi(apiClient);
   }
 
+  @Bean
+  public WebMvcTagsContributor clientIdTagContributor() {
```
cbb330, line 85: this is the core piece of this PR: takes header from client request and sets it into the metrics registry tags for use in the handler interceptor for metric emission

services/tables/src/main/java/com/linkedin/openhouse/tables/config/TablesMvcConfigurer.java
```diff
@@ -42,7 +42,8 @@ public void addInterceptors(InterceptorRegistry registry) {
           registry
               .addInterceptor(interceptor)
               .addPathPatterns("/**")
-              .excludePathPatterns("/actuator/**", "/**/api-docs/**", "/**/swagger-ui/**");
+              .excludePathPatterns(
+                  "/actuator/**", "/**/api-docs/**", "/**/swagger-ui/**", "/favicon.ico", "/error");
```
cbb330, line 46: the favicon asset has been added, and thus needs to be provided without auth to enable loading via swagger or (weirdly) /actuator/prometheus also requests the favicon.

cbb330: /error is also added in case any missing route is hit 

/error will be presented to client

but the auth handler interceptor will prevent it from being shown to client as 404 and convert it into a 401 which is confusing. so fixing that here

services/tables/src/main/resources/application.properties
```diff
@@ -1,5 +1,6 @@
 spring.mvc.throw-exception-if-no-handler-found=true
-spring.web.resources.add-mappings=false
+spring.mvc.static-path-pattern=favicon.ico
```
cbb330, line 2: this property allows the favicon to be presented to browsers and thus fixes issue with metrics/logs spamming errors when trying to load the favicon via swagger or /actuator/prometheus

context for the old property is:

> We have seen a few instances where the malformed/wrong URL causes 404 error code and makes users believe they are actually requesting some resource that doesn't exist any more. However this is obviously different from true positive case where request resource doesn't exist. This problem is more severe with SQL API on top of REST API, as SQL API will simply translate 404 to "Table Not found" whereas it should be "URL malformed / Bad Request"
This PR addresses that by customizing handling of "NoHandlerFoundException" which by default is treated as 404 by springboot framework.

which this method _should_ solve both of above issues.

services/tables/src/main/java/com/linkedin/openhouse/tables/config/MainApplicationConfig.java
autumnust: actually, we may want to have a primary bean within LI and there we puts constraints to only allow spark and trino

autumnust (changes requested):
mainly questions regarding some changes that don't seem to be needed, as well as where to put the client-name constraints.

services/common/src/main/java/com/linkedin/openhouse/common/Constants.java
```diff
@@ -0,0 +1,16 @@
...
+  public static final String HTTPHEADER_CLIENT_NAME = "X-Client-Name";
+
+  public static final String CLIENT_NAME_DEFAULT_VALUE = "";
```
autumnust, line 9: shall we have more information default value

services/tables/src/main/java/com/linkedin/openhouse/tables/config/MainApplicationConfig.java
```diff
@@ -74,6 +81,32 @@ public UserTableApi provideApiInstance() {
...
+    return new WebMvcTagsContributor() {
+      @Override
+      public Iterable<Tag> getTags(
+          HttpServletRequest request,
+          HttpServletResponse response,
+          Object handler,
+          Throwable exception) {
+        String clientName = request.getHeader(HTTPHEADER_CLIENT_NAME);
```
autumnust, line 93: do we need to deal with null value here?

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
```diff
@@ -205,6 +197,18 @@ private void setSessionIdInWebClientHeader(WebClient.Builder webClientBuilder) {
...
 
+  /**
+   * Set clientName in the header of webClient. If a clientName is not provided, do nothing.
+   *
+   * @param webClientBuilder
+   */
+  private void setClientNameInWebClientHeader(WebClient.Builder webClientBuilder) {
+    if (clientName != null) {
+      webClientBuilder.defaultHeaders(h -> h.add(HTTPHEADER_CLIENT_NAME, clientName));
+    }
```
autumnust, line 208: I think this is the right choice, to leave this client_name_header as null when clientName is not available from client-side, since we will live with a mix of client that doesn't have / have this field for a while ? 
But accordingly have we fully tested server behavior for this ?

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
abhisheknath2011, line 28:
```suggestion
  private static final String HTTP_HEADER_CLIENT_NAME = "X-Client-Name";
```

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
```diff
@@ -154,14 +154,16 @@ private WebClient createSecureWebClient(String baseUrl, String truststoreLocatio
...
+    return client;
+  }
+
+  private WebClient.Builder getWebClientBuilder(String baseUrl) {
     WebClient.Builder webClientBuilder = createWebClientBuilder();
     setSessionIdInWebClientHeader(webClientBuilder);
+    setClientNameInWebClientHeader(webClientBuilder);
     return webClientBuilder
-        .baseUrl(baseUrl)
-        .clientConnector(connector)
```
abhisheknath2011, line 162: I think we can set the connector as well here and the builder initialization will be completed in one place together and we can directly return webClient.

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
```diff
@@ -61,13 +65,17 @@ protected WebClient createWebClient(
...
     if (HTTPS.equals(transportProtocol)) {
-      return createSecureWebClient(baseUrl, truststoreLocation);
+      httpClient = createSecureHttpClient(truststoreLocation);
     } else if (HTTP.equals(transportProtocol)) {
-      return createWebClient(baseUrl);
+      httpClient = createHttpClient(baseUrl);
     } else {
       throw new RuntimeException("The transport protocol must be https/http");
     }
+    return getWebClientBuilder(baseUrl)
```
abhisheknath2011, line 76:
```suggestion
    return buildWebClient(baseUrl)
```
We can move the connector part to buildWebClient itself.

services/tables/src/main/java/com/linkedin/openhouse/tables/config/MainApplicationConfig.java
cbb330: why primary? if i understand correctly, if beans collide then its a a startup exception. isn't that desirable? otherwise user may have a bean that doesn't get registered and have to debug. 

>actually, we may want to have a primary bean within LI and there we puts constraints to only allow spark and trino

ok good idea. i'm thinking to add the 
`public static final List<String> ALLOWED_CLIENT_NAME_VALUES = Arrays.asList("trino", "spark");` field as LI bean, and default to an empty array if not provided. since all OS users may benefit from this header. what do you think?

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
```diff
@@ -154,14 +154,16 @@ private WebClient createSecureWebClient(String baseUrl, String truststoreLocatio
...
+  private WebClient.Builder getWebClientBuilder(String baseUrl) {
     WebClient.Builder webClientBuilder = createWebClientBuilder();
     setSessionIdInWebClientHeader(webClientBuilder);
+    setClientNameInWebClientHeader(webClientBuilder);
     return webClientBuilder
-        .baseUrl(baseUrl)
-        .clientConnector(connector)
         .codecs(codecs -> codecs.defaultCodecs().maxInMemorySize(IN_MEMORY_BUFFER_SIZE))
-        .build();
+        .baseUrl(baseUrl);
```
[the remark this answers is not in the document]
cbb330: no the order doesn't matter, i double checked. all of the methods here return 'Builder' type and inside the DefaultWebClientBuilder implementation that we are using, they all return a "this" value. none of the methods used here edit state -- just setting private attributes.

integrations/java/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseCatalog.java
```diff
@@ -106,13 +108,15 @@ public void initialize(String name, Map<String, String> properties) {
...
     try {
       TablesApiClientFactory tablesApiClientFactory = TablesApiClientFactory.getInstance();
       tablesApiClientFactory.setStrategy(HttpConnectionStrategy.fromString(httpConnectionStrategy));
+      tablesApiClientFactory.setClientName(clientName);
+      this.apiClient = tablesApiClientFactory.createApiClient(uri, token, truststore);
```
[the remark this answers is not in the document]
cbb330: this ordering here was a mistake. fixed.

but also note that i'm changing

`TablesApiClientFactory.getInstance().createApiClient(uri, token, truststore);`

to

`tablesApiClientFactory.createApiClient(uri, token, truststore);`

because the previous code must never have worked -- setSession and setStrategy were made on an instance and then the instance was never used, a new instance was created. I checked ligghtly whether anyopne was using these configs and it appears no.

services/common/src/main/java/com/linkedin/openhouse/common/Constants.java
cbb330: the main user for this metric is prometheus formatted metric scrapers.

these scrapers would want to query with filters by either "trino" (for example) or not use the filter at all, or use aggregates like distinct or group by. in this last case (distinct or group by) so **yes, it would be more informative to have a string value of "unspecified" vs ""** for use cases like displaying "unspecified" in a linechart.

didn't see a clear best practice or required spec online, but did find this advice from a knowledgeable user of prometheus:
> Prometheus treats empty string as label not being present at all, but it's not recommended to have inconsistent sets of labels on your metrics, as it makes them harder to aggregate over.  A concrete value like "-" is better.

https://groups.google.com/g/prometheus-users/c/18UG8HFiFLg?pli=1

cbb330: fixed to "unspecified"

services/tables/src/main/java/com/linkedin/openhouse/tables/config/MainApplicationConfig.java
cbb330: https://docs.oracle.com/cd/E17802_01/products/products/servlet/2.5/docs/servlet-2_5-mr2/javax/servlet/http/HttpServletRequest.html

documents that getHeader will return null if not found. as well, there isn't an exception in the interface to be thrown if the value isn't found. 

then the value is used in "List<String>.contains()" later which will attempt to find the null value  in the list which won't exist. so this code as is should handle null. 

this case is under test too, `testMetricsWithClientNameHeaderNotPresentDefaultValue`

client/secureclient/src/test/java/com/linkedin/openhouse/client/ssl/TablesApiClientFactoryTest.java
```diff
@@ -8,13 +8,34 @@
...
+    SslContext mockSslContext = Mockito.mock(SslContext.class);
+    tmpCert = File.createTempFile("tmpcacert", ".crt");
+    tablesApiClientFactorySpy = Mockito.spy(TablesApiClientFactory.getInstance());
+    Mockito.doReturn(mockSslContext).when(tablesApiClientFactorySpy).createSslContext(anyString());
+  }
+
+  @AfterAll
+  public static void tearDown() {
+    tmpCert.deleteOnExit();
+    ;
```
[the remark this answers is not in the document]
cbb330: fixed (needs to be added as hook in next line after file creation rather than in tearDown)

services/common/src/main/java/com/linkedin/openhouse/common/Constants.java
cbb330: good point, i moved the class to only be used in the tables MvcConfig context, but couldn't make fields package private due to unittests in a diff package using these fields

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
cbb330: fixed

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
cbb330: fair point, fixed this

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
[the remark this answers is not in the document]
cbb330: i ended up changing this code to be more similar to main, removing the change for the order.

services/tables/src/main/java/com/linkedin/openhouse/tables/config/MainApplicationConfig.java
autumnust: that sounds fine also, to have a separate bean.

autumnust (changes requested):
other than the changes request for moving spark/trino-constraints into internal app, lgtm

cbb330:
@autumnust pushed one more commit to address moving trino/spark options into configuration. 

but I chose to push them into cluster properties yml because these are just string values and the workflow for adding a new accepted client header would just require yaml update. 

let me know what you think and can approve PR

autumnust (approved):
LGTM!

PR #173, opened 2024-08-23
DLO: file entropy scoring.

anjagruenheid:
## Summary

pull request in 2-3 lines.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


anjagruenheid:
FYI, it wouldn't let me create a branch without spotless which triggered a whole bunch of code changes. The relevant changes for this PR are in the datalayout package. Please advise how I can fix the spotless issue.

teamurko:
> FYI, it wouldn't let me create a branch without spotless which triggered a whole bunch of code changes. The relevant changes for this PR are in the datalayout package. Please advise how I can fix the spotless issue.

one simple way is
```
git reset f10c2663b43a61fc0dd26be87cddf9db1bd7f0fc^ -- `git show f10c2663b43a61fc0dd26be87cddf9db1bd7f0fc --name-only | egrep -v libs/datalayout | grep java`
```
then commit staged, checkout unstaged files

we need to setup hooks in IDE, or we can also terminal during commit/push

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/FileEntropyDataLayoutStrategyGenerator.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.spark.api.java.function.MapFunction;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Encoders;
+
+/**
+ * Data layout optimization strategies generator for OpenHouse. Generates a list of strategies with
+ * score.
+ */
+@Builder
+public class FileEntropyDataLayoutStrategyGenerator implements DataLayoutStrategyGenerator {
```
teamurko, line 22: I also thought about approach to add entropy attribute to https://github.com/linkedin/openhouse/blob/8619c9f97ab2d89013e5ba4c36d214257940c56a/libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/strategy/DataLayoutStrategy.java#L13 and just calculate it in the existing generator, it's easier downstream to enable it and analyze, especially if we don't go with it as a separate strategy to consider but just need to collect data.

anjagruenheid: It's easier downstream, I agree, but if you do this with every alternative approach, you'll end up with a lot of confusion, right? I was thinking about implementing this as a separate strategy, because logically, it is separate from the existing strategy. @sumedhsakdeo, what do you think?

anjagruenheid:
> > FYI, it wouldn't let me create a branch without spotless which triggered a whole bunch of code changes. The relevant changes for this PR are in the datalayout package. Please advise how I can fix the spotless issue.
> 
> one simple way to mitigate
> 
> ```
> git reset f10c2663b43a61fc0dd26be87cddf9db1bd7f0fc^ -- `git show f10c2663b43a61fc0dd26be87cddf9db1bd7f0fc --name-only | egrep -v libs/datalayout | grep java`
> ```
> 
> then commit staged, checkout unstaged files
> 
> we need to setup hooks in IDE, or we can also use terminal during commit/push

Even if I do this + a local commit for a file path only, it won't let me do it. Should I just do a separate PR with all the spotless fixes to help others, too?

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/FileEntropyDataLayoutStrategyGenerator.java
sumedhsakdeo: Good perspectives on both ends.

It might be useful to standardize code on a framework like OODA (observe -> orient -> decide -> act, with feedback loop builtin), or the one that was proposed by GSL.

- observe - orient / trait-generation (calculates estimated file-count reduction (gain), compute usage (cost), entropy (mse), etc.)
- decide / candidate selection (score calculations based on mse or simple gain/cost formula or using multi-objective optimization, followed by ranking of scores and picking of top-K)
- act / scheduling (perform the work, and compare the estimated cost, gain and entropy with actual cost, gain and entropy)

Today, we have combined trait-generation (cost, gain and simple scoring) under in one method https://github.com/linkedin/openhouse/blob/8619c9f97ab2d89013e5ba4c36d214257940c56a/libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutStrategyGenerator.java#L58

To be truly extensible we might want to refactor this to fit in a framework like discussed above. Class hierarchy organized accordingly. But we are not there yet. Right now we need to iterate fast here and record observations and compare options. Hence, I am fine with adding entropy calculation to the existing function as a trait to make quicker progress and making computation (in a single job) and analysis simple, with the expectation that we might have to undergo some refactoring in future.

Ultimately, I propose we need to compute strategies along two dimensions:
_Scope_: Data being considered for compaction
_Algorithm_: Used for scoring and top-K

| Strategy  | Simple | Entropy | MultiObjectiveOptimization |
| ------------- | ------------- | ------------- | ------------- |
|  FullTable |  x | x | x |
|  TablePartitions |  x | x | x |
|  TableSnapshots |  x | x | x |

* Scope*
_FullTable_ means we are looking at data for the entire table. This is what is currently implemented and being analyzed.
_TablePartitions_ means we are looking at incremental data for the last N partitions written into from the last M snapshots.
_TableSnapshots_ means we are looking at incremental data for the last M snapshots.

* Algorithm *
_Simple_ is score computed as file count reduced per GBhr
_TablePartitions_ is score computed as mean squared error 
_MultiObjectiveOptimization_ is score computed by standardizing multiple objectives like file count reduction, compute cost, files scanned during querying, etc.

* x *
top-K selected for compaction.

Hope that helps.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/FileEntropyDataLayoutStrategyGenerator.java
```diff
@@ -0,0 +1,67 @@
...
+    // Normalize.
+    mse = mse / fileSizes.count();
+
+    return DataLayoutStrategy.builder()
+        .config(
+            DataCompactionConfigUtil.createDefaultCompactionConfig(
+                TARGET_BYTES_SIZE,
+                tablePartitionStats.get().count(),
+                TableFileStatsUtil.getAggregatedFileStats(tableFileStats, TARGET_BYTES_SIZE)._1))
+        .score(mse)
```
sumedhsakdeo, line 64: How is entroy used in the multi objective optimization where we need to rank based on compute cost objective and then pick top-K?
How do we compute cost and gain for the entropy strategy?

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/FileEntropyDataLayoutStrategyGenerator.java
jcamachor: @sumedhsakdeo , I like the framework and believe it aligns well with what we proposed from the GSL side. In fact, I think merging the two approaches, including the feedback loop, makes sense; I'll follow up on this offline.

Regarding implementation. While I agree we should move fast, I suggest that we shape the code around these ideas as early as possible, since we are just getting started, rather than intertwining all these concepts. Otherwise, while we will move faster now, it will become more difficult to incorporate new ideas and contributions as we move forward. </2cents>

sumedhsakdeo: I asked @teamurko to draft a UML for the class hierarchy for the framework, something which is extensible. However, I am also curious on the entropy readings, so ok to wrap up that part which @anjagruenheid has started in the current code and not take up newer extension until we have refactored it.

anjagruenheid: I actually think it's not going to be too different from what you have right now, if I understand @sumedhsakdeo correctly. As components, you'd have: 
1) A score calculator (trait generation) for each candidate, i.e., given a set of table stats and/or partition stats calculates a score per input set. That roughly translates to the current *DataStrategy classes.
2) You'd have a decision component (abstract class), essentially, this should be a component that takes as input the scores for each candidate set and then ranks them in some manner. A specific implementation of that is the top-k decision maker.
3) You'd have the scheduler which would trigger compaction for those tables selected in step 2, either all at once in some other manner.

You'd wrap this to accommodate the scope (FullTable / TablePartitions / TableSnapshots). The components themselves are imo independent of the scope, i.e., their functionality would not change. For example, for entropy, you hand a list of file sizes to the score calculator. If the list is for a full table/partition/snapshot is irrelevant to the score calculation.
Essentially, an outer loop would generate the appropriate candidates within the scope, then trigger the first component for each candidate, followed by components two and three. Scope, scoring/decision algorithm, and scheduler type could be set in a configuration, thus, it's easily extendible and mix-and-match depending on what you want to run/test.

I'm adjusting the code to add entropy as an attribute for now, will ping here once it's done.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/FileEntropyDataLayoutStrategyGenerator.java
anjagruenheid: Entropy is a version of gain, i.e., the higher the entropy, the higher the potential gain (a higher entropy means more deviation from the target file size). If you're calculating the score as gain/cost, you could also alternatively use entropy/cost for the top-k calculation and a higher score means that the table has higher priority for compaction according to its entropy.

sumedhsakdeo: Sounds good so I would sort by entropy for ranking, and use the sum(compute cost) < compute budget for top-k selection. That works!

Impact in terms of file count reduction can be continued to be measured as sum(gains).

anjagruenheid:
> > > FYI, it wouldn't let me create a branch without spotless which triggered a whole bunch of code changes. The relevant changes for this PR are in the datalayout package. Please advise how I can fix the spotless issue.
> > 
> > 
> > one simple way to mitigate
> > ```
> > git reset f10c2663b43a61fc0dd26be87cddf9db1bd7f0fc^ -- `git show f10c2663b43a61fc0dd26be87cddf9db1bd7f0fc --name-only | egrep -v libs/datalayout | grep java`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > then commit staged, checkout unstaged files
> > we need to setup hooks in IDE, or we can also use terminal during commit/push
> 
> Even if I do this + a local commit for a file path only, it won't let me do it. Should I just do a separate PR with all the spotless fixes to help others, too?

Fixed! :)

teamurko (approved):

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutStrategyGenerator.java
```diff
@@ -125,4 +130,14 @@ private double estimateComputeGbHr(long rewriteBytes) {
...
   }
+
+  private double computeEntropy(Dataset<Long> fileSizes) {
+    double mse = 0.0;
+    Iterator<Long> itr = fileSizes.toLocalIterator();
+    while (itr.hasNext()) {
+      mse += Math.pow(TARGET_BYTES_SIZE - itr.next(), 2);
+    }
+    // Normalize.
+    return mse / fileSizes.count();
```
sumedhsakdeo, line 141: Divide by zero handling?

sumedhsakdeo (changes requested):

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutStrategyGenerator.java
```diff
@@ -125,4 +130,20 @@ private double estimateComputeGbHr(long rewriteBytes) {
...
+
+  /** Computes the file entropy as the difference of a set of file's target and actual file size. */
+  private double computeEntropy(Dataset<Long> fileSizes) {
+    // If no files available, MSE is 0.
+    if (fileSizes.count() == 0) {
+      return 0;
+    }
+    // Compute the mean-squared error.
+    double mse = 0.0;
+    Iterator<Long> itr = fileSizes.toLocalIterator();
```
sumedhsakdeo, line 142: Is this more efficient to compute in a spark query. Especially when number of files in table grow large to 10s of millions 
Thoughts? CC: @teamurko

anjagruenheid: Should be faster for tables with a large number of files since it would parallelize, but we'd have to pass the target file size into the query in addition to the tableFileName potentially as another function in TableFileStats? Currently the target file size is a property of the strategy and not defined in the Spark app.

teamurko: I'd bet that data_files metadata table will be partitioned, and local iterator will keep memory usage low. In terms of serial calculation, yeah, it could be a little slow since there is fetching of partitions happening kind of one by one, not because the code itself that sums up stuff.

We could rewrite in spark, spark should be able to serialize the target size constant, right? But I'm OK with the current implementation as well.

sumedhsakdeo (approved):

teamurko (approved):

PR #174, opened 2024-08-26
Feature Toggle Part III: Using autogen HTS client for feature-toggle

autumnust:
## Summary

This PR completes `BaseTableToggleStatusesRepository` by injecting `ToggleStatusApi` so that it will connect to HTS to check the status of a table's feature, defined by `featureId`, `tableId` and `databaseId`. 

Meanwhile, this PR also introduces the first use case for feature-toggle: Disallow table creation when toggled off.  See the change in `creationEligibilityCheck`. This change should have small to little impact towards mysql-storage layer for the feature toggle, as there are only <6K table-creation request (from which this code-path will be triggered) in last 7d. Plus, this is a read-only operation for mysql. 

pull request in 2-3 lines.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

Adding unit tests cases for the new code added. 

Also, tested the changes in the local environment, in the following sequence: 
```
INSERT IGNORE INTO table_toggle_rule (feature, database_pattern, table_pattern, id, creation_time_ms) \
 VALUES ('stop_create', 'demodb', 'demotable', DEFAULT, DEFAULT);

 mysql> select * from table_toggle_rule;
+-------------+------------------+---------------+----+------------------+----------------------------+
| feature     | database_pattern | table_pattern | id | creation_time_ms | ETL_TS                     |
+-------------+------------------+---------------+----+------------------+----------------------------+
| demo        | demodb           | demotable     |  1 |             NULL | 2024-08-27 23:42:39.599344 |
| stop_create | demodb           | demotable     |  2 |             NULL | 2024-08-27 23:49:33.036207 |
+-------------+------------------+---------------+----+------------------+----------------------------+
```

First step is ensuring the new feature-rule is inserted. 

Then, creating the tables: 
```
curl "${curlArgs[@]}" -XPOST http://localhost:8000/v1/databases/demodb/tables/ \
--data-raw '{
  "tableId": "demotable",
  "databaseId": "demodb",
  "baseTableVersion": "INITIAL_VERSION",
  "clusterId": "LocalS3Cluster",
  "schema": "{\"type\": \"struct\", \"fields\": [{\"id\": 1,\"required\": true,\"name\": \"id\",\"type\": \"string\"},{\"id\": 2,\"required\": true,\"name\": \"name\",\"type\": \"string\"},{\"id\": 3,\"required\": true,\"name\": \"ts\",\"type\": \"timestamp\"}]}",
  "timePartitioning": {
    "columnName": "ts",
    "granularity": "HOUR"
  },
  "tableProperties": {
    "key": "value"
  }
}'
```

The command above get the following responses: 
```
{"status":"FORBIDDEN","error":"Forbidden","message":"Feature [stop_create] has been toggled off for the [demodb.demotable]","stacktrace":null,"cause":"Not Available"}%
```

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


autumnust:
The next-step is to allow feature-toggle to be specified using wildcard, to fully satisfy the requirement from stoping table-creation briefly if necessary.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
```diff
@@ -57,6 +63,18 @@ protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
         new MetricsReporter(this.meterRegistry, METRICS_PREFIX, Lists.newArrayList()));
   }
 
+  /** Overwritten for annotation purpose. */
+  @Override
+  @IcebergFeatureGate(value = STOP_TABLE_CREATION_TOGGLE)
```
autumnust, line 68: @cbb330  @HotSushi  this is the code to enable the feature toggle

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/toggle/annotation/FeatureGateAspect.java
```diff
@@ -0,0 +1,30 @@
...
+@Aspect
+@Component
+@Slf4j
+public class FeatureGateAspect {
+
+  @Autowired private TableFeatureToggle tableFeatureToggle;
+
+  @Before("@annotation(featureGate)")
+  public void checkIcebergFeatureFlag(JoinPoint joinPoint, IcebergFeatureGate featureGate) {
+    TableIdentifier tableIdentifier = (TableIdentifier) joinPoint.getArgs()[0];
```
HotSushi, line 22: is there an assumption that first parameter will always be tableidentifier? if so can we add it to the documentation?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
```diff
@@ -57,6 +63,18 @@ protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
         new MetricsReporter(this.meterRegistry, METRICS_PREFIX, Lists.newArrayList()));
   }
 
+  /** Overwritten for annotation purpose. */
+  @Override
```
HotSushi, line 67: we override the same method on li-openhouse side, do you anticipate the code working ok there?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
```diff
@@ -15,6 +15,8 @@ public final class CatalogConstants {
   /** Used to uniquely identify an update towards a table from user side. */
   public static final String COMMIT_KEY = "commitKey";
 
+  static final String STOP_TABLE_CREATION_TOGGLE = "stop_create";
```
HotSushi, line 18: can we make all toggle keys to follow same pattern name? maybe FEATURE_TOGGLE_STOP_CREATE, FEATURE_TOGGLE_S3_STORAGE etc

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/toggle/annotation/IcebergFeatureGate.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/** Used to gate the feature of the corresponding method where this annotation is applied. */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface IcebergFeatureGate {
```
HotSushi, line 11: Why `Iceberg`?

services/tables/src/main/java/com/linkedin/openhouse/tables/config/MainApplicationConfig.java
```diff
@@ -52,14 +53,23 @@ public class MainApplicationConfig extends BaseApplicationConfig {
 
   @Autowired StorageManager storageManager;
 
+  @Bean
+  public UserTableApi provideApiInstance() {
+    return new UserTableApi(getHtsConfiguredApiClient());
+  }
+
+  @Bean
+  public ToggleStatusApi provideToggleApiInstance() {
```
HotSushi, line 62: should we introduce a new configuration file specific to feature toggle?

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -84,6 +85,8 @@ public class OpenHouseInternalRepositoryImpl implements OpenHouseInternalReposit
 
   @Autowired PreservedKeyChecker preservedKeyChecker;
 
+  @Autowired TableFeatureToggle tableFeatureToggle;
```
HotSushi, line 88: why needed here?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
```diff
@@ -57,6 +63,18 @@ protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
...
+  @IcebergFeatureGate(value = STOP_TABLE_CREATION_TOGGLE)
+  public Table createTable(
+      TableIdentifier identifier,
+      Schema schema,
+      PartitionSpec spec,
+      String location,
+      Map<String, String> properties) {
+    return super.createTable(identifier, schema, spec, location, properties);
```
HotSushi, line 75: I don't see an easy way to use the annotation feature on Storage Selectors. We have Selector interface as follows:

````
Class StorageSelector {
    Storage selectStorage(db, table) {
            isFeatureEnabled = ...      // planning to call FeatureToggle
            if (isFeatureEnabled) {
                    return Storage;
            }
           return storageManager.defaultstorage()
    }
}
````

We can always use `@ToggleStatusesRepository` but not the annotation easily. Let me know if there's some way

File: iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java [hunk 9]
```diff
@@ -9,12 +10,17 @@
 import com.linkedin.openhouse.internal.catalog.model.HouseTablePrimaryKey;
 import com.linkedin.openhouse.internal.catalog.repository.HouseTableRepository;
 import com.linkedin.openhouse.internal.catalog.repository.exception.HouseTableRepositoryException;
+import com.linkedin.openhouse.internal.catalog.toggle.annotation.IcebergFeatureGate;
```
HotSushi, line 13: We should move the classes out of `iceberg` module, they are generic enough to apply to any format. Also, storage classes wouldn't be able to use it easily if they are part of internalcatalog module.

HotSushi (changes requested):
few comments. Biggest feedback is about moving the FeatureToggle classes out of `internalcatalog` into `table-service`, it is not specific to iceberg and should be generic enough to apply to different formats.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/toggle/annotation/FeatureGateAspect.java
```diff
@@ -0,0 +1,30 @@
...
+
+    if (tableFeatureToggle.isFeatureActivated(
+        tableIdentifier.namespace().toString(), tableIdentifier.name(), featureGate.value())) {
+      throw new ResourceGatedByToggledOnFeatureException(
```
cbb330, line 26: this behavior makes sense for STOP_TABLE feature

but what about ENABLE_SPACE_CREATE (as a hypothetical)?

I think bare minimum we should be able to (without needing to catch an exception / add boilerplate):
- pass through feature and fallback to default behavior ( e.g. skip a void function (but i don't feel comfortable requiring side-affect functions)? or return false from a boolean function?)
- refuse a request (like is done here, throw exception)

and then a stretch / aspiration / something to design for the future would be things like:
- Return a Default Value
- Execute an alternative function
- Retry the feature
- Emit additional metrics or logs or events

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/toggle/ResourceGatedByToggledOnFeatureException.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * The exception thrown when the request feature of a resource is toggled on thus being gated access
+ * or modification.
+ */
+public class ResourceGatedByToggledOnFeatureException extends RuntimeException {
+  public ResourceGatedByToggledOnFeatureException(
+      String featureId, String databaseId, String tableId, Throwable throwable) {
+    this(
+        String.format(
+            "Feature [%s] has been toggled on for the [%s.%s]", featureId, databaseId, tableId),
```
cbb330, line 12: It would be very helpful (and with many more features and dimensions to match with, necessary) to also return and log what conditions led to the feature being toggled or not toggled

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
cbb330, line 18: a future nice to have would be features that are type safe, perhaps all of the configuration for such features is stored in java in a schema that is then deployed to the database like in an ORM.

Then the column value of the feature can be changed at runtime.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
autumnust: hmm thanks for catching that, I didn't realize this is being overwritten. I will need to test with that then.

autumnust:
```
    public Table createTable(TableIdentifier identifier, Schema schema, PartitionSpec spec, String location, Map<String, String> properties) {
        log.info("Creating table {} at {}", identifier.toString(), location);
        properties.put(TableProperties.WRITE_METADATA_LOCATION, String.format("%s/%s", location, METADATA_SUB_FOLDER));
        properties.put(TableProperties.WRITE_DATA_LOCATION, String.format("%s/%s", location, DATA_SUB_FOLDER));
        // TODO: Safe to remove TableProperties.WRITE_FOLDER_STORAGE_LOCATION from properties when clients
        //  have upgraded.
        properties.put(TableProperties.WRITE_FOLDER_STORAGE_LOCATION,
            String.format("%s/%s", location, DATA_SUB_FOLDER));

        return super.createTable(identifier, schema, spec, location, properties);
    }
```

What I anticipated is the `super.createTable` call will trigger this annotation aspect.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
autumnust: addressed it.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
autumnust: no need here. addressed.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/toggle/annotation/IcebergFeatureGate.java
autumnust: Because this annotation will stay within this module and becomes an internal-catalog specific one. 
Any better suggestion?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
autumnust: got it, addresssed.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/toggle/ResourceGatedByToggledOnFeatureException.java
autumnust: This is the user-side message look like ```
{"status":"INTERNAL_SERVER_ERROR","error":"Internal Server Error","message":"com.linkedin.openhouse.tables.toggle.ResourceGatedByToggledOnFeatureException: Feature [stop_create] has been toggled on for the [demodb.demotable]","stacktrace":null,"cause":"Not Available"}%``` 

Does that look good or you are referring to something different ?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/toggle/annotation/FeatureGateAspect.java
autumnust: I did some generalization of this method, like not limiting `TableIdentifier` to be the first parameter. 
But the constraints of method being annotated having to have a `TableIdentifier` object still exists. Thus, I am alway naming the annotation with `Iceberg` in it which should answer the question here: https://github.com/linkedin/openhouse/pull/174#discussion_r1744618336

services/tables/src/main/java/com/linkedin/openhouse/tables/config/MainApplicationConfig.java
autumnust: This is the only new addition for the feature toggle, don't think we need to consider a separate config. 
And, this is a HTS related config so makes sense to be part of mainAppConfig?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/toggle/annotation/FeatureGateAspect.java
autumnust: I think for enabling-style feature like `ENABLE_SPACE_CREATE`: 

- We would first refactor the code (if necessary) so that there will be a bean-interface preferably returning boolean value and manage the control flow of whether enabling space-creation or not. 
- We will develop an annotation and the corresponding aspect-method to influence the result of that boolean method using feature-activation result + logical operator

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
autumnust: See this: https://github.com/linkedin/openhouse/pull/174#discussion_r1747542004 
But yes annotation is just a syntactic sugar.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/toggle/ResourceGatedByToggledOnFeatureException.java
cbb330: whats missing in that message is the full trace of why the feature got toggled on.

for example,

`Feature [stop_create] has been toggled on for the [demodb.demotable] due to predicate [tableId] operand matches [*.*]`

but a short term could be that the operator views the source definition of [stop_create], however with many more features and operand types for the predicate (not limited to tableId, perhaps headless account or client platform) this might become challenging.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
```diff
@@ -1,5 +1,6 @@
 package com.linkedin.openhouse.internal.catalog;
 
+import static com.linkedin.openhouse.internal.catalog.CatalogConstants.*;
```
HotSushi, line 3: lets remove * and declare explicit constants?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
HotSushi: Ack! let's ensure we test this out particularly.

services/tables/src/main/java/com/linkedin/openhouse/tables/config/MainApplicationConfig.java
HotSushi: Can we add comments around how these beans are used. 
nit: The reason for asking a separate file was to make it easier for us to lift and shift this FeatureGate to other services if need be (jobs-service etc). ok to skip if unnecessary efforts.

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/ResourceGatedByToggledOnFeatureException.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  public ResourceGatedByToggledOnFeatureException(
+      String featureId, String databaseId, String tableId) {
+    this(
+        String.format(
+            "Feature [%s] has been toggled on for the [%s.%s]", featureId, databaseId, tableId));
+  }
+
+  public ResourceGatedByToggledOnFeatureException(String message, Throwable throwable) {
```
HotSushi, line 23: replace this with lombok's @StandardException?

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/FeatureGateAspect.java
```diff
@@ -0,0 +1,51 @@
...
+@Aspect
+@Component
+@Slf4j
+public class FeatureGateAspect {
+
+  @Autowired private TableFeatureToggle tableFeatureToggle;
+
+  @Before("@annotation(featureGate)")
+  public void checkIcebergFeatureFlag(JoinPoint joinPoint, IcebergFeatureGate featureGate) {
+    TableIdentifier tableIdentifier = obtainIdFromAnnotationArgs(joinPoint);
```
HotSushi, line 26: `TableIdentifier` makes me think its better to leave the class in internalcatalog, but rest of the classes look good in tables

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/repository/ToggleStatusesRepositoryImpl.java
```diff
@@ -1,29 +1,48 @@
...
 
   @Override
-  public Optional<TableToggleStatus> findById(ToggleStatusKey toggleStatusKey) {
-    return Optional.empty();
+  public <S extends TableToggleStatus> Iterable<S> saveAll(Iterable<S> entities) {
+    throw new UnsupportedOperationException(
+        "Write Operation into Toggle status API is not supported");
   }
 
   @Override
```
HotSushi, line 48: can we throw unsupportedException for other methods as well like existsByID, findall etc?

HotSushi (changes requested):
few more comments, will look good after that

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/toggle/IcebergFeatureGate.java
```diff
@@ -0,0 +1,16 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Used to gate the feature of the corresponding method where this annotation is applied, on the
+ * granularity of Iceberg table identified by {@link org.apache.iceberg.catalog.TableIdentifier}
+ */
+@Retention(RetentionPolicy.RUNTIME)
```
HotSushi, line 12: Please add more documentation around the expectations (first parameter being tableidentifier), behaviors (throw exceptions), and an example. Coder typically wouldn't hunt for the Aspect file instead rely on annotation to provide all the information.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
autumnust: addressed.

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/FeatureGateAspect.java
autumnust: This feature aspect is not going to only server internal catalog though. 
My intention for it is a central annotation behavior's registration hub. For now, we have an annotation in catalog but we will soon add a new one which expects `TableDto` as part of signature. 

The perfect solution will be to have a cluster:annotation module for such central annotation registration.

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/ResourceGatedByToggledOnFeatureException.java
autumnust: Our Lombok version is 1.18.20; This feature is added in 1.18.22. 
OK to pursue that but since Lombok generated code get into runtime, that will involver bigger testing efforts. Shall we do this separately?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/toggle/IcebergFeatureGate.java
autumnust: addressed.

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/repository/ToggleStatusesRepositoryImpl.java
autumnust: addressed.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/toggle/ResourceGatedByToggledOnFeatureException.java
autumnust: I see, so the predicate part. 
I think this is enough amount of info to give to end users. Note that this is not dev facing error message. 
But I do plan to make some improvements for exception handling for `ResourceGatedByToggledOnFeatureException` using exception-handler, in a separate PR.

autumnust:
A few next-step lined up from reviewing: 
- Upgrade Lombok to 1.8.22+ to enable `@StandardException`, and test. 
- Exception handling for `ResourceGatedByToggledOnFeatureException`
- wildcard support for feature-toggle, so that for the use case of stop-creation, we can use `*,*,stop-create` 
- refactor property-check bean to ensure it has `tableDTO` included in the signature. 
- wiring preserved-key check with feature-toggle, allow MoR props to be enabled for some DB/tables.

HotSushi (approved):

services/tables/src/main/java/com/linkedin/openhouse/tables/config/MainApplicationConfig.java
autumnust: let's revisit that when feature toggle is needed to be in other services, because that might be a bigger change beyond spring boot-app config object.

cbb330 (approved):

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
autumnust: I tried adding a feature registry in the new PR: , See `cluster/configs/src/main/java/com/linkedin/openhouse/cluster/configs/TblPropsToggleRegistry.java`

autumnust: not exactly the thing that can be strongly typed in database schema. wanted to think about how to create constraints in mysql so that we could avoid leaving these feature Id being free-form string as much as possible.

PR #177, opened 2024-08-27
Handle null case when table metadata is empty. Also increase timeout to 180 seconds

maluchari:
## Summary

After retrieving list of tables in databases, it is possible for getTable call to error due to any intermittent issues or due to 404 if table was just deleted. But in this case, the responseBody is null and accessing tableCreator leads to NPE and thus failing to collect any table from the particular database. This PR returns null for creator in these cases and hence the client can move on to collect further tables in that database. Also 60 seconds seem less for a getTable call as it involves underlying storage, hts backend and also authorization component that is used. Hence increased to 180 seconds.

## Changes

- [ ] Internal API Changes
- [X] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

- [X] No tests added or updated. Please explain why. If unsure, please feel free to ask for help.

Existing tests passed. 

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/client/TablesClient.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/client/TablesClient.java
@@ -202,8 +202,12 @@ public class TablesClient {
                       return Optional.ofNullable(response.getResults())
                           .map(Collection::stream)
                           .orElseGet(Stream::empty)
-                          .map(this::mapTableResponseToTableMetadata)
-                          .filter(databaseFilter::apply)
+                          .flatMap(
+                              result ->
+                                  Optional.ofNullable(mapTableResponseToTableMetadata(result))
+                                      .filter(databaseFilter::apply)
+                                      .map(Stream::of)
+                                      .orElseGet(Stream::empty))
                           .collect(Collectors.toList());
                     },
                 Collections.emptyList()));
@@ -307,11 +311,7 @@ public class TablesClient {
 
     if (tableResponseBody == null) {
       log.error("Error while fetching metadata for table: {}", metadata);
-      return TableMetadata.builder()
-          .creator(null)
-          .dbName(responseBody.getDatabaseId())
-          .tableName(responseBody.getTableId())
-          .build();
+      return null;
     }
 
     String creator = tableResponseBody.getTableCreator();
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/client/TablesClient.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/client/TablesClient.java
@@ -38,7 +38,7 @@ import org.springframework.retry.support.RetryTemplate;
 @Slf4j
 @AllArgsConstructor
 public class TablesClient {
-  private static final int REQUEST_TIMEOUT_SECONDS = 60;
+  private static final int REQUEST_TIMEOUT_SECONDS = 180;
   private final RetryTemplate retryTemplate;
   private final TableApi tableApi;
   private final DatabaseApi databaseApi;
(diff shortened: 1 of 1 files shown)


apps/spark/src/main/java/com/linkedin/openhouse/jobs/client/TablesClient.java
```diff
@@ -291,7 +291,18 @@ protected TableMetadata mapTableResponseToTableMetadata(GetTableResponseBody res
             .dbName(responseBody.getDatabaseId())
             .tableName(responseBody.getTableId())
             .build();
-    String creator = getTable(metadata).getTableCreator();
+    GetTableResponseBody tableResponseBody = getTable(metadata);
+
+    if (tableResponseBody == null) {
+      log.error("Error while fetching metadata for table: {}", metadata);
+      return TableMetadata.builder()
+          .creator(null)
```
teamurko, line 299: good catch! we can skip those tables since we can't set proxy user

maluchari: I have updated to return null and skip including that is the results of getTables. PTAL.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/client/TablesClient.java
```diff
@@ -291,7 +295,14 @@ protected TableMetadata mapTableResponseToTableMetadata(GetTableResponseBody res
             .dbName(responseBody.getDatabaseId())
             .tableName(responseBody.getTableId())
             .build();
-    String creator = getTable(metadata).getTableCreator();
+    GetTableResponseBody tableResponseBody = getTable(metadata);
+
+    if (tableResponseBody == null) {
+      log.error("Error while fetching metadata for table: {}", metadata);
+      return null;
```
abhisheknath2011, line 302:
```suggestion
      return TableMetadata;
```

apps/spark/src/main/java/com/linkedin/openhouse/jobs/client/TablesClient.java
```diff
@@ -291,7 +295,14 @@ protected TableMetadata mapTableResponseToTableMetadata(GetTableResponseBody res
             .dbName(responseBody.getDatabaseId())
             .tableName(responseBody.getTableId())
             .build();
-    String creator = getTable(metadata).getTableCreator();
+    GetTableResponseBody tableResponseBody = getTable(metadata);
+
+    if (tableResponseBody == null) {
+      log.error("Error while fetching metadata for table: {}", metadata);
```
abhisheknath2011, line 301: Should this be just db.table instead of whole metadata?

maluchari: Metadata is only db and table here

apps/spark/src/main/java/com/linkedin/openhouse/jobs/client/TablesClient.java
```diff
@@ -38,7 +38,7 @@
 @Slf4j
 @AllArgsConstructor
 public class TablesClient {
-  private static final int REQUEST_TIMEOUT_SECONDS = 60;
+  private static final int REQUEST_TIMEOUT_SECONDS = 180;
```
abhisheknath2011, line 41: Should we make slightly less than gateway timeout which is also 180?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/client/TablesClient.java
maluchari: Not following this.The change is to return null and skip that as part of results aggregation so prefer to return null here

abhisheknath2011: `tableResponseBody` is already null here.

maluchari: Makes sense now. I think suggestion was to return tableMetadata before edit. Prefer to return explicitly  as null.

teamurko (approved):

apps/spark/src/main/java/com/linkedin/openhouse/jobs/client/TablesClient.java
maluchari: Will start with 180. We can tune it back if needed.

PR #186, opened 2024-09-06
Add an option to enable/disable storage stats collection as part of TableStatsCollector

maluchari:
## Summary

Storage stats computation can be expensive as this could involve recursive directory walk in the file system.  Hence making the storage stats computation optional so that this can be invoked as needed. 

## Changes

- [X] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/TableStatsCollectionSparkApp.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/TableStatsCollectionSparkApp.java
@@ -13,20 +13,24 @@ import org.apache.commons.cli.Option;
  * Class with main entry point to collect Iceberg stats for a table.
  *
  * <p>Example of invocation: com.linkedin.openhouse.jobs.spark.TableStatsCollectionSparkApp
- * --tableName db.testTable
+ * --tableName db.testTable --skipStorageStats true
  */
 @Slf4j
 public class TableStatsCollectionSparkApp extends BaseTableSparkApp {
 
-  public TableStatsCollectionSparkApp(String jobId, StateManager stateManager, String fqtn) {
+  private final Boolean skipStorageStatsCollection;
+
+  public TableStatsCollectionSparkApp(
+      String jobId, StateManager stateManager, String fqtn, Boolean skipStorageStatsCollection) {
     super(jobId, stateManager, fqtn);
+    this.skipStorageStatsCollection = skipStorageStatsCollection;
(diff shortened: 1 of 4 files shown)


apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/TableStatsCollectionSparkApp.java
```diff
@@ -43,10 +47,16 @@ protected void publishStats(IcebergTableStats icebergTableStats) {
...
+        new Option("s", "skipStorageStats", false, "Whether to skip storage stats collection"));
+
     CommandLine cmdLine = createCommandLine(args, extraOptions);
     TableStatsCollectionSparkApp app =
         new TableStatsCollectionSparkApp(
-            getJobId(cmdLine), createStateManager(cmdLine), cmdLine.getOptionValue("tableName"));
+            getJobId(cmdLine),
+            createStateManager(cmdLine),
+            cmdLine.getOptionValue("tableName"),
+            cmdLine.hasOption("skipStorageStats"));
```
teamurko, line 59: How does the caller (scheduler?) determine when to skip? If it's ad-hoc, then caller knows and might as well not run at all, if the caller is scheduler, then we have filter assigned to the task where the table can be skipped, is this for another use case?

maluchari: I want to keep it disabled for all tables now and re-enable it when we have a better way to handle the storage part. Also apart from scheduler, when we run adhoc jobs, it might be better to have this enabled to collect storage stats as required.

teamurko: would this work
`kubectl patch cronjobs <job-name> -p '{"spec" : {"suspend" : true }}'`

maluchari: IIUC, This implies suspending the whole job. But this PR targets to disable only storage stats computation. Other table stats , like metadata and iceberg stats and current snapshot metrics will still be computed.

teamurko (approved):

PR #189, opened 2024-09-10
Feature-Toggle Use case: Allow preserved tblprops to be modified on Table basis

autumnust:
## Summary

This PR demonstrates how annotation-based feature-toggle can help with allowing certain "preserved table properties"  to be modifiable by privileged users of certain tables. Common use cases for this is when there are high-impact features that team is not feeling confident to rollout to all tables, we can use feature-toggle to allow some tables to be able to use this first. 

Here for demonstration purpose, we show how `openhouse.tableType` can become writable if the feature `enable_tabletype` is toggled on for certain tables, in a e2e unit test setup. 

This PR also contains some refactoring required to make this work. Specifically, for the interface used to manage preserved keys, we need to introduce table-awareness of it so that it can make decision based on the table identifier. There are also a few test cases adjustment (e.g. moved from e2e-type into mock as I see fit). 

Every time when developers need to add new table properties into this toggle-managed process, one needs to do the following: 

- Use `TblPropsToggleRegistry` to register feature and its corresponding table properties. For merge-on-read feature for example, one will need to add the 3 keys (`write.delete.mode`, `write.update.mode`, `write.merge.mode` ) into the map and with the value as pre-defined featureId, say `enable_mor`.  This is intentionally designed to be a code change.
- Deploy the code change above.
- Update feature toggle's backend to allow some table with specific naming patterns to have `enable_mor` activated

## Changes

- [ ] Internal API Changes
- [] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


cluster/configs/src/main/java/com/linkedin/openhouse/cluster/configs/TblPropsToggleRegistry.java
```diff
@@ -0,0 +1,31 @@
+package com.linkedin.openhouse.cluster.configs;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import javax.annotation.PostConstruct;
+import org.springframework.stereotype.Component;
+
+/** A central registry for all table properties that are managed through Feature Toggle. */
+@Component
```
cbb330, line 10: why must this class be spring managed? 

as it is defined here, the registry can be reduced to a constant map. is there a future vision to extend the registry?

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/InternalRepositoryUtils.java
cbb330, line 1: this utils file is static and requires injecting PreservedKeyChecker bean (which is already injected by springboot).

requiring the bean to be 1) injected in OpenHouseInternalRepositoryImpl and then additionally injected as a parameter into Utils method call. seems unwieldly for testing. 

do you think the Utils methods which rely on non-static parameters (such as the checker) should be moved to OpenHouseInternalRepositoryImpl? or maybe these utils should be moved to their own bean?

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/FeatureToggleAspect.java
```diff
@@ -17,10 +24,41 @@
...
+  public boolean checkTblPropEnabled(
+      ProceedingJoinPoint proceedingJoinPoint, TblPropsEnabler tblPropsEnabler) throws Throwable {
+    if (((MethodSignature) proceedingJoinPoint.getSignature()).getReturnType() == boolean.class
+        && proceedingJoinPoint.getArgs()[1] instanceof TableDto
+        && proceedingJoinPoint.getArgs()[0] instanceof String) {
+      TableDto tableDto = (TableDto) proceedingJoinPoint.getArgs()[1];
+      String key = (String) proceedingJoinPoint.getArgs()[0];
+
+      Optional<String> feature = tblPropsToggleRegistry.obtainFeatureByKey(key);
+      if (!feature.isPresent()) {
```
cbb330, line 43: nit: can we use map / orElseGet for the optional feature so that the reader knows that the code below this condition is only executed once the feature is known to be present

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BasePreservedKeyChecker.java
```diff
@@ -16,6 +21,16 @@ public boolean isKeyPreserved(String key) {
     return IS_OH_PREFIXED.test(key);
   }
 
+  /**
+   * Details of determining if a key is preserved for the given {@link TableDto} is hidden in the
+   * implementation of aspect that associates with the annotation {@link TblPropsEnabler}
+   */
+  @Override
+  @TblPropsEnabler
```
cbb330, line 29: could this annotation be more specific to better indicate that is may override the return value to true if table prop is enabled by feature?

e.g.

@OverrideIfTablePropsEnabled

cbb330:
overall, I like how this is designed, the plug into existing code is lightweight and easy to understand without boilerplate. thanks for implementing this as a good example for other feature toggles.

cluster/configs/src/main/java/com/linkedin/openhouse/cluster/configs/TblPropsToggleRegistry.java
```diff
@@ -0,0 +1,31 @@
...
+
+  public static final String ENABLE_TBLTYPE = "enable_tabletype";
+
+  // TODO: Using these vocabularies as MySQL validation
+  private final Map<String, String> featureKeys = new HashMap<>();
+
+  @PostConstruct
+  public void initializeKeys() {
+    // placeholders: demo purpose
+    featureKeys.put("openhouse.tableType", ENABLE_TBLTYPE);
```
HotSushi, line 21: this is test code sneaking into prod? is there a better way to do this? maybe make the featurekeys protected visiblefortesting?

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BasePreservedKeyChecker.java
```diff
@@ -16,6 +21,16 @@ public boolean isKeyPreserved(String key) {
     return IS_OH_PREFIXED.test(key);
   }
 
+  /**
+   * Details of determining if a key is preserved for the given {@link TableDto} is hidden in the
+   * implementation of aspect that associates with the annotation {@link TblPropsEnabler}
```
HotSushi, line 26: nit: `.. implementation of aspect {@link FeatureToggleAspect}`

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BasePreservedKeyChecker.java
```diff
@@ -16,6 +21,16 @@ public boolean isKeyPreserved(String key) {
     return IS_OH_PREFIXED.test(key);
   }
 
+  /**
+   * Details of determining if a key is preserved for the given {@link TableDto} is hidden in the
```
HotSushi, line 25: can you also explain whats the behavior for `@TblPropsEnabler` if its disabled? will it throw exception/return true etc in this doc?

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/FeatureToggleAspect.java
```diff
@@ -17,10 +24,41 @@
...
+        return (boolean) proceedingJoinPoint.proceed();
+      }
+
+      boolean tableFeatureEnabled =
+          tableFeatureToggle.isFeatureActivated(
+              tableDto.getDatabaseId(), tableDto.getTableId(), feature.get());
+
+      // feature activation overwriting the decision of annotated method
+      return ((boolean) proceedingJoinPoint.proceed()) || tableFeatureEnabled;
```
HotSushi, line 52: the annotated method is keypreserved. if feature  is `enable_mor`,  isKeypreserved should return false for `write.delete.mode` right? so that user can run alter table SQLs?  but here it will return `true` because `tableFeatureEnabled` is true. 
I may not be understanding this correctly

HotSushi (changes requested):

cluster/configs/src/main/java/com/linkedin/openhouse/cluster/configs/TblPropsToggleRegistry.java
autumnust: That's basically all benefits that dependency injection can bring: ease of testing, no constructor change etc. 
And yes We might want to extend this behavior in the internal app.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/InternalRepositoryUtils.java
autumnust: not sure if I follow ... this util class doesn't requires injection of `PreservedKeyChecker` bean, its member method has parameter like `com.linkedin.openhouse.tables.repository.impl.InternalRepositoryUtils#getUserTblProps`

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/FeatureToggleAspect.java
autumnust: addressed.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BasePreservedKeyChecker.java
autumnust: couldn't really think of a better name ... I will avoid "Override" since there's `@Override` annotation there.

cluster/configs/src/main/java/com/linkedin/openhouse/cluster/configs/TblPropsToggleRegistry.java
```diff
@@ -0,0 +1,31 @@
+package com.linkedin.openhouse.cluster.configs;
```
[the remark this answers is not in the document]
autumnust: addressed.

cluster/configs/src/main/java/com/linkedin/openhouse/cluster/configs/TblPropsToggleRegistry.java
autumnust: I can just remove this code.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/PreservedKeyChecker.java
```diff
@@ -1,12 +1,24 @@
...
+
 /**
  * Defining keys that are preserved for OpenHouse to use. Those keys, if present in table
  * properties, are read-only.
  */
 public interface PreservedKeyChecker {
 
+  /** @deprecated Please consider using {@link #isKeyPreservedForTable(String, TableDto) instead} */
+  @Deprecated
   boolean isKeyPreserved(String key);
```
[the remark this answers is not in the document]
autumnust: there are usage in li-openhouse.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BasePreservedKeyChecker.java
autumnust: addressed.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BasePreservedKeyChecker.java
autumnust: Adding this annotation will just have an interceptor examine the result and examine the result together whether there's a feature being active. Here the feature is allowing some keys to be modified. Not sure if that answers the question.

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/FeatureToggleAspect.java
autumnust: Yeah I think you caught it right. I confused myself for the semantic of `preservedKey` here. 

- `isKeypreserved` by default return true for `write.delete.mode` 
- with feature of "enable_mor" active for a table `t`, `isKeypreserved` return false for `write.delete.mode`  
- so the condition should be ` ((boolean) proceedingJoinPoint.proceed()) && !tableFeatureEnabled;` for `enable_preservedKey` feature

autumnust: unfortunately all my unit tests got flipped 👎  great catch.

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/FeatureToggleAspect.java
```diff
@@ -0,0 +1,95 @@
...
+        && proceedingJoinPoint.getArgs()[0] instanceof String) {
+      TableDto tableDto = (TableDto) proceedingJoinPoint.getArgs()[1];
+      String key = (String) proceedingJoinPoint.getArgs()[0];
+
+      return tblPropsToggleRegistry
+          .obtainFeatureByKey(key)
+          .map(
+              feature ->
+                  evalProceeding(proceedingJoinPoint)
+                      && !tableFeatureToggle.isFeatureActivated(
```
HotSushi, line 47: this condition will only work if we have the feature as `allowMOR` or `allowSomething`. It wouldn't be able to achieve `disallowSomething` for example say `disallowPoliciesForTable`. I understand that its not in the scope, but can we add an **important note in TblsProsToggleRegistry documentation saying only allowlisting is supported**?

services/tables/src/main/java/com/linkedin/openhouse/tables/config/TblPropsToggleRegistry.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.stereotype.Component;
+
+/** A central registry for all table properties that are managed through Feature Toggle. */
+@Component
+public class TblPropsToggleRegistry {
+
+  public static final String ENABLE_TBLTYPE = "enable_tabletype";
+
+  // TODO: Using these vocabularies as MySQL validation
+  private final Map<String, String> featureKeys = new HashMap<>();
```
HotSushi, line 16: add a note in documentation that ** only allowlisting is supported **

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/ToggleH2StatusesRepository.java
```diff
@@ -11,4 +16,29 @@
...
+public interface ToggleH2StatusesRepository extends ToggleStatusesRepository {
+  String FEATURE_ACTIVATION_TABLE_NAME = "acv_tbl";
+  TblPropsToggleRegistry registry = new TblPropsToggleRegistry();
+
+  /**
+   * A trick to activate a specific feature for a specific table-id. This is only for testing
+   * purpose.
+   */
+  @Override
+  default Optional<TableToggleStatus> findById(ToggleStatusKey toggleStatusKey) {
```
HotSushi, line 28: why this trick? couldn't we have autowired the repository and insertd the toggleKey before running the unit test. This trick will affect other test cases too?

HotSushi (changes requested):

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/FeatureToggleAspect.java
autumnust: yes the semantics of this annotation is VERY strict. 
At this point I think annotation is really just a syntactical sugar to avoid contaminating the main body of service implementation, not a general enough approach to extend to other semantics.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/ToggleH2StatusesRepository.java
autumnust: we don't support any write method for toggle endpoint. 

This is the same thing we do for mock test like below: 
```
public class MockTablesApiHandler implements TablesApiHandler {
  @Override
  public ApiResponse<GetTableResponseBody> getTable(
      String databaseId, String tableId, String actingPrincipal) {
    switch (databaseId) {
      case "d200":
        return ApiResponse.<GetTableResponseBody>builder()
            .httpStatus(HttpStatus.OK)
            .responseBody(RequestConstants.TEST_GET_TABLE_RESPONSE_BODY)
            .build();
      case "d404":
        throw new NoSuchUserTableException(databaseId, tableId);
      case "dnullpointer":
        throw new NullPointerException(); // test for exception handler
      default:
        return null;
    }
  }
```

services/tables/src/main/java/com/linkedin/openhouse/tables/config/TblPropsToggleRegistry.java
autumnust: I think the documentation should be added in either the annotation or `featureToggleAspect` right? 
This registry can be used for other things, like "disallow" style feature

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/FeatureToggleAspect.java
HotSushi: > At this point I think annotation is really just a syntactical sugar to avoid contaminating the main body 

thats alright! but lets use documentation as a key to convey strictness

HotSushi: can we add such documentation to `TblsProsToggleRegistry`?

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/ToggleH2StatusesRepository.java
HotSushi: I see, the test you share is not e2e right? its `/mock`? should we try injecting a custom TestToggleH2StatusesRepository that lets us insert mock entries instead of hardcoding?

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/PreservedPropsToggleEnabler.java
```diff
@@ -0,0 +1,19 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for the method with specific signature (see {@link
+ * com.linkedin.openhouse.tables.toggle.FeatureToggleAspect} which controls if a table property is
+ * considered preserved (read-only for users).
+ *
+ * <p>Usage of this annotation should be very specific to allow default-preserved keys to be
+ * non-preserved. (This annotation is used as syntactical sugar to avoid contaminating main body of
+ * service implementation)
```
HotSushi, line 15: nit: Do not use this annotation on a denyList for example disablePoliciesForTable

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/FeatureToggleAspect.java
autumnust: sure, addressed altogether with the comment here. https://github.com/linkedin/openhouse/pull/189#discussion_r1772175096

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/PreservedPropsToggleEnabler.java
autumnust: addressed.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/ToggleH2StatusesRepository.java
autumnust: yes this is from mock module. 
I just realize that the in-memory version of the repo can support `save` method for feature activation. Just use that to replace "trick" that was added in the earlier version.

HotSushi (approved):
LGTM!

PR #196, opened 2024-09-17
Handle tblProperties for replication flow

rohitkum2506:
## Summary

Table commits have strict checks for reserved table properties to avoid modifying them. This leads to issues for use cases  where cross cluster iceberg commits for tables with same tableId should be compatible which allows features like table backup. 
This PR introduces conditions when such check can be relaxed. Specifically for cross cluster table replication.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
@@ -131,12 +131,7 @@ public class OpenHouseInternalRepositoryImpl implements OpenHouseInternalReposit
     } else {
       table = catalog.loadTable(tableIdentifier);
       Transaction transaction = table.newTransaction();
-      if (!skipEligibilityCheck(table.properties(), tableDto.getTableProperties())) {
-        // eligibility check is relaxed for request from replication flow since preserved properties
-        // & tableType
-        // will differ in tableDto and existing table.
-        updateEligibilityCheck(table, tableDto);
-      }
+      updateEligibilityCheck(table, tableDto);
 
       boolean schemaUpdated =
           doUpdateSchemaIfNeeded(transaction, writeSchema, table.schema(), tableDto);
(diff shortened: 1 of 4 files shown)


services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -160,6 +166,19 @@ public TableDto save(TableDto tableDto) {
...
+      Map<String, String> existingTableProps, Map<String, String> newTableprops) {
+    TableType existingTableType =
+        TableType.valueOf(existingTableProps.get(getCanonicalFieldName(TABLE_TYPE_KEY)));
+    TableType newTableType =
+        TableType.valueOf(newTableprops.get(getCanonicalFieldName(TABLE_TYPE_KEY)));
+    return existingTableType == TableType.REPLICA_TABLE
+        && newTableType == TableType.PRIMARY_TABLE
+        && !existingTableProps
+            .get(getCanonicalFieldName(CLUSTER_ID))
+            .equals(newTableprops.get(getCanonicalFieldName(CLUSTER_ID)));
```
autumnust, line 179: can we just have this skip-logic as part of eligibility check? something like when the base table is replica, just skip the eligibility check?

autumnust: and, why is "cluster id" check required?

autumnust (changes requested):
I was initially thinking of bigger refactoring where we should have a tableType related object across the whole service cluster and make decision on if certain things apply for certain table type. 
Just writing this down for discussion. For the scope of this change, we can simply by just skipping eligibility check when dealing with replica. Do have some questions regarding the condition. 

Also, can we engage second reviewers for this change? @maluchari can u also take a Quick Look here

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
rohitkum2506: cluster_id and primary_table type check is to confirm the request is coming from other cluster and from a Primary Table.
Writes are blocked for replica_table at Auth layer, so for OSS this check would ensure proper exceptions are surfaced for requests coming from same cluster with discrepancy in tableType property.

maluchari: I think it is okay to include clusterID in this check. One nit: Can we just move this check to updateEligbilitycheck method itself instead of having a new one?

rohitkum2506: Would prefer to keep the conditions in a separate method considering more cases can come up in future. 
But yeah, can move the `if` clause inside the `updateEligbilitycheck`.

autumnust (approved):

rohitkum2506:
> I was initially thinking of bigger refactoring where we should have a tableType related object across the whole service cluster and make decision on if certain things apply for certain table type.

Layering for tableType handling was not evident and clean solution IMO. TableType property is part of tableDTO and requestObject so didn't add a new property for it.

PR #247, opened 2024-11-01
Skip field-id re-assignment during table creation

vikrambohra:
## Summary
Doc: https://docs.google.com/document/d/152W319aqZMWpxnAbEoO6L86vM8jObC5vkZrRNLqMvHs/edit?tab=t.0#heading=h.tpufqkd57g2w

During table creation, iceberg reassigns the field-ids of client supplied schema and recreates table metadata. This can cause schema field-id mismatch during replica table creation if source table schema has evolved. This change ensures that replica table schema will always have same field-ids as the source table schema during replica table creation.

Changes
1. Client adds table level property "client.table.schema" with serialized iceberg schema 
2. During table creation, new code creates a new table metadata object using the schema provided in step 1

## Changes

- [x] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
@@ -17,6 +17,8 @@ public final class CatalogConstants {
 
   static final String FEATURE_TOGGLE_STOP_CREATE = "stop_create";
 
+  public static final String CLIENT_TABLE_SCHEMA = "client.table.schema";
+
   private CatalogConstants() {
     // Noop
   }
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
@@ -17,7 +17,7 @@ public final class CatalogConstants {
 
   static final String FEATURE_TOGGLE_STOP_CREATE = "stop_create";
 
-  public static final String CLIENT_TABLE_SCHEMA = "client.table.schema";
+  static final String CLIENT_TABLE_SCHEMA = "client.table.schema";
 
   private CatalogConstants() {
     // Noop
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
@@ -186,8 +186,8 @@ public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperatio
           TableMetadata.buildFromEmpty()
               .setLocation(metadata.location())
               .addSchema(sourceSchema, metadata.lastColumnId())
-              .addPartitionSpec(clonePartitionSpec(metadata.spec(), sourceSchema))
-              .addSortOrder(cloneSortOrder(metadata.sortOrder(), sourceSchema))
+              .addPartitionSpec(rebuildPartitionSpec(metadata.spec(), sourceSchema))
+              .addSortOrder(rebuildSortOrder(metadata.sortOrder(), sourceSchema))
               .setProperties(metadata.properties())
               .build();
       metadata = newTableMetadata;
@@ -307,7 +307,7 @@ public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperatio
     }
   }
 
-  public static PartitionSpec clonePartitionSpec(PartitionSpec original, Schema schema) {
+  public static PartitionSpec rebuildPartitionSpec(PartitionSpec original, Schema schema) {
     // Create a builder with the new schema
     PartitionSpec.Builder builder = PartitionSpec.builderFor(schema);
 
@@ -342,7 +342,7 @@ public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperatio
     return builder.build();
   }
 
-  public static SortOrder cloneSortOrder(SortOrder original, Schema newSchema) {
+  public static SortOrder rebuildSortOrder(SortOrder original, Schema newSchema) {
     SortOrder.Builder builder = SortOrder.builderFor(newSchema).withOrderId(original.orderId());
 
     for (SortField field : original.fields()) {
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
@@ -178,19 +178,23 @@ public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperatio
   @Override
   protected void doCommit(TableMetadata base, TableMetadata metadata) {
 
-    if (base == null && metadata.properties().get("source.table.schema") != null) {
-      String schema = metadata.properties().get("source.table.schema");
-      Schema sourceSchema = SchemaParser.fromJson(schema);
-
-      TableMetadata newTableMetadata =
+    /**
+     * During table creation, the table metadata object that arrives here has the field-ids
+     * reassigned from the client supplied schema.This code block creates a new table metadata
+     * object using the client supplied schema by preserving its field-ids.
+     */
+    if (base == null && metadata.properties().get(CatalogConstants.CLIENT_TABLE_SCHEMA) != null) {
+      Schema clientSchema =
+          SchemaParser.fromJson(metadata.properties().get(CatalogConstants.CLIENT_TABLE_SCHEMA));
+      metadata =
           TableMetadata.buildFromEmpty()
               .setLocation(metadata.location())
-              .addSchema(sourceSchema, metadata.lastColumnId())
-              .addPartitionSpec(rebuildPartitionSpec(metadata.spec(), sourceSchema))
-              .addSortOrder(rebuildSortOrder(metadata.sortOrder(), sourceSchema))
+              .setCurrentSchema(clientSchema, metadata.lastColumnId())
+              .addPartitionSpec(
+                  rebuildPartitionSpec(metadata.spec(), metadata.schema(), clientSchema))
+              .addSortOrder(rebuildSortOrder(metadata.sortOrder(), clientSchema))
               .setProperties(metadata.properties())
               .build();
-      metadata = newTableMetadata;
     }
 
     int version = currentVersion() + 1;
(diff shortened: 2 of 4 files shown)


iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
```diff
@@ -17,6 +17,8 @@ public final class CatalogConstants {
 
   static final String FEATURE_TOGGLE_STOP_CREATE = "stop_create";
 
+  public static final String CLIENT_TABLE_SCHEMA = "client.table.schema";
```
autumnust, line 20: it doesn't have to be public since it only gets referenced in the same package.
Even if you want to reuse it in client side, it's probably not a good idea to have a server-side constant used as a dependency on client side.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -168,6 +177,26 @@ public void commit(TableMetadata base, TableMetadata metadata) {
   @SuppressWarnings("checkstyle:MissingSwitchDefault")
   @Override
   protected void doCommit(TableMetadata base, TableMetadata metadata) {
+
+    /**
+     * During table creation, the table metadata object that arrives here has the field-ids
+     * reassigned from the client supplied schema.This code block creates a new table metadata
+     * object using the client supplied schema by preserving its field-ids.
+     */
+    if (base == null && metadata.properties().get(CatalogConstants.CLIENT_TABLE_SCHEMA) != null) {
```
autumnust, line 186: why is this not implemented in https://github.com/linkedin/openhouse/blob/b754e5e9dbafb1e83061efa2bc5357658b3ef505/services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java#L92 ? 

`doCommit` is a too low level implementation to distinguish between creation vs. update

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -282,6 +311,98 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+  /**
+   * Build a new partition spec with new schema from original pspec. The new pspec has the same
+   * partition fields as the original pspec with source ids from the new schema
+   *
+   * @param originalPspec
+   * @param originalSchema
+   * @param newSchema
+   * @return new partition spec
+   */
+  public static PartitionSpec rebuildPartitionSpec(
```
autumnust, line 323: these methods shouldn't be public static. 
If this is purely for testing, mocking is a better way.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -282,6 +311,98 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+  public static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema) {
+    SortOrder.Builder builder = SortOrder.builderFor(newSchema);
+
+    for (SortField field : originalSortOrder.fields()) {
+      // Find the field name in the original schema based on the sourceId
+      String fieldName = originalSortOrder.schema().findField(field.sourceId()).name();
+      // Check if the sortorder field is present in new schema
+      if (newSchema.findField(fieldName) == null) {
+        throw new IllegalArgumentException(
+            "Field " + fieldName + " does not exist in the new schema");
```
autumnust, line 390: what does this exception become when it reaches client ?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -282,6 +311,98 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+      }
+      // Create a new SortField with the updated sourceId and original direction and null order
+      Term term = Expressions.ref(fieldName);
```
autumnust, line 393: For my understanding: `Term` object is the one where the field-id is carried ?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -282,6 +311,98 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+      } else if ("month".equalsIgnoreCase(transformString)) {
+        builder.month(fieldName);
+      } else if ("day".equalsIgnoreCase(transformString)) {
+        builder.day(fieldName);
+      } else if ("hour".equalsIgnoreCase(transformString)) {
+        builder.hour(fieldName);
+      } else {
+        throw new UnsupportedOperationException("Unsupported transform: " + transformString);
+      }
+    }
```
autumnust, line 368: this should be placed in a separate method where the transform is reconstructed.

integrations/spark/spark-3.1/openhouse-spark-itest/src/test/java/com/linkedin/openhouse/spark/catalogtest/CatalogOperationTest.java
```diff
@@ -68,6 +70,106 @@ public void testCatalogWriteAPI() throws Exception {
...
+
+      // Field ids reassigned (Status quo)
+      TableIdentifier tableIdentifier = TableIdentifier.of("replication_test", "t1");
+      Map<String, String> props = new HashMap<>();
+      Table table = icebergCatalog.createTable(tableIdentifier, schema, null, props);
+      Schema schemaAfterCreation = table.schema();
+      Assertions.assertFalse(schemaAfterCreation.sameSchema(schema));
+      Assertions.assertEquals(1, schemaAfterCreation.findField("a").fieldId());
+      Assertions.assertEquals(3, schemaAfterCreation.findField("a.b").fieldId());
+      Assertions.assertEquals(2, schemaAfterCreation.findField("c").fieldId());
```
autumnust, line 93: shall we just make it simpler here, by asserting it as false since we expect `a.b` to be 2 and `c` to be 3. 
and the line 101 to 115 can be simplify omitted. basically we will create a case where it failed before the changes but succeeded after the change.

autumnust (changes requested):
```Client adds table level property "client.table.schema" with serialized iceberg schema```
This is not yet implemented right ?

File: iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java (hunk 2 above)
Comment (vikrambohra) [reply iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java:186]: If I make the change in the high level code it will get overridden in this bock and committed right after https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/BaseMetastoreCatalog.java#L197-L204

The only place where this can be safely put without getting overridden is the commit() or doCommit() (which is called internally by commit). decided to do it in doCommit() as there are additional checks in super.commit() before doCommit() is called.

File: iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java (hunk 6 above)
Comment (vikrambohra) [reply iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java:368]: Separate method is not such a good idea here because we will be mutating the pspec builder in there and then build the pspec outside of it.

File: iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java (hunk 4 above)
Comment (vikrambohra) [reply iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java:390]: IIUC, it get translated to a 400 error HttpStatus.BAD_RREQUEST

https://github.com/linkedin/openhouse/blob/main/services/common/src/main/java/com/linkedin/openhouse/common/exception/handler/OpenHouseExceptionHandler.java#L365

File: iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java (hunk 3 above)
Comment (vikrambohra) [reply iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java]: done

File: iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java (hunk 2 above)
Comment (autumnust) [reply iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java:186]: To make sure the schema object is honored from that layer, I think what needs to happen is a customized `BaseMetastoreCatalogTableBuilder` extension so that when https://github.com/linkedin/openhouse/blob/b754e5e9dbafb1e83061efa2bc5357658b3ef505/services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java#L113 is called, the customized tableBuilder's method will be invoked and pass the schema object all the way to doCommit method.  

cc @HotSushi  since similar thing is done at the client side to provide a customized `OpenHouseTableBuilder`

File: iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java (hunk 6 above)
Comment (autumnust) [reply iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java:368]: not following. what's the problem of creating a method only having side effects

File: iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java (hunk 4 above)
Comment (autumnust) [reply iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java:390]: sg.

File: iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java (hunk 2 above)
Comment (vikrambohra) [reply iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java:186]: That would have been good but that approach is not ideal.

The [problematic code](https://github.com/apache/iceberg/blob/fe2f593cd025223e4ab5ab41a296fb106ce3b1cf/core/src/main/java/org/apache/iceberg/BaseMetastoreCatalog.java#L197) is in the middle of the create() method of the `BaseMetastoreCatalogTableBuilder ` 

https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/BaseMetastoreCatalog.java#L189-L207
```
public Table create() {
      TableOperations ops = newTableOps(identifier);
      if (ops.current() != null) {
        throw new AlreadyExistsException("Table already exists: %s", identifier);
      }

      String baseLocation = location != null ? location : defaultWarehouseLocation(identifier);
      tableProperties.putAll(tableOverrideProperties());
      **TableMetadata metadata =
          TableMetadata.newTableMetadata(schema, spec, sortOrder, baseLocation, tableProperties);**

      try {
        ops.commit(null, metadata);
      } catch (CommitFailedException ignored) {
        throw new AlreadyExistsException("Table was created concurrently: %s", identifier);
      }

      return new BaseTable(ops, fullTableName(name(), identifier), metricsReporter());
    }```

If we extend the BaseMetastoreCatalogTableBuilder, then we need to override the create() method entirely and replace just the problematic code while copy pasting the rest and skip calling the create() in parent class. Code maintainability becomes hard here to keep up with OSS changes. also there are private methods (tableOverrideProperties()) being called that need to be duplicated

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
vikrambohra: done

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
autumnust: I see, seems the issue of putting the fix in the tableBuilder layer is this method: 
(https://github.com/apache/iceberg/blob/5e139f48505b9b622e2cf389eae2bb74d6420510/core/src/main/java/org/apache/iceberg/TableMetadata.java#L80)

autumnust (approved):
thanks for addressing the comments, LGTM.

autumnust:
I hope the sort spec and partition-transform keywords are not going to change in the future, otherwise the whole reconstruct logic becomes fragile.

vikrambohra:
> I hope the sort spec and partition-transform keywords are not going to change in the future, otherwise the whole reconstruct logic becomes fragile.

+1 It's not ideal but iceberg has made some classes and methods package-private to make the reconstruction hard. For example. These [2 methods](https://github.com/apache/iceberg/blob/fe2f593cd025223e4ab5ab41a296fb106ce3b1cf/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L604C1-L613C6) would have made our life easier had they been public.

```
    Builder add(int sourceId, String name, Transform<?, ?> transform) {
      return add(sourceId, nextFieldId(), name, transform);
    }

    Builder add(int sourceId, int fieldId, String name, Transform<?, ?> transform) {
      checkAndAddPartitionName(name, sourceId);
      fields.add(new PartitionField(sourceId, fieldId, name, transform));
      lastAssignedFieldId.getAndAccumulate(fieldId, Math::max);
      return this;
    }
```

vikrambohra:
@autumnust One minor test case update. Need another ship it!

autumnust (approved):

PR #221, opened 2024-10-03
Add integration for openhouse-java-iceberg-1.5-runtime and openhouse-spark-3.5-runtime

jiang95-dev:
## Summary

Adding 3 integration artifacts: **openhouse-java-iceberg-1.5-runtime**, **openhouse-spark-3.5-runtime**, and **tables-test-fixtures-iceberg-1.5**. They are leveraging the **iceberg-core:1.5.2** and **iceberg-spark-runtime-3.5_2.12**, and can only be compiled under Java 9+. This client side upgrade is the step after the server side upgrade, and apps upgrade will be the next step.

## Changes

- [ ] Client-facing API Changes
- [ ] Internal API Changes
- [x] Bug Fixes
- [ ] New Features
- [ ] Performance Improvements
- [ ] Code Style
- [x] Refactoring
- [ ] Documentation
- [ ] Tests

There are no duplicated files after the clean up. Files changed are:
- build.gradle: Spark 3.5 needs jackson library to be at least 2.15.
- java-runtime/build.gradle
- java-itest/build.gradle
- spark-runtime/build.gradle
- spark-runtime: Changed the interface and implemented new functions for logical plans and execution plans to pass the build.
- spark-itest/build.gradle: Added jvmArgs for Java 17.
- spark-itest: Half of the tests need to be refactored since Spark 3.5 brought in some changes.
- tables-test-fixtures: Client can depend on this module after spark 3.5 is launched.
- spark-3.5-base-hadoop3.2.dockerfile: Build on spark 3.5 and the new spark runtime module.

## Testing Done

- [ ] Manually Tested on local docker setup. Please include commands ran, and their output.
- [x] Added new tests for the changes made.
- [x] Updated existing tests to reflect the changes made.
- [ ] No tests added or updated. Please explain why. If unsure, please feel free to ask for help.
- [ ] Some other form of testing like staging or soak time in production. Please explain.

All unit tests and integration tests passed.

Local docker tests:
1. Added `spark-3.5-base-hadoop3.2.dockerfile` and made local change of `spark-service.yml` for spark tests.
2. Launched spark shell with the command:
```
bin/spark-shell --packages org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:1.5.2  \
  --jars openhouse-spark-runtime_2.12-*-all.jar  \
[...]

PR #255, opened 2024-11-16
[PR1] refactoring :apps build and adding spark3.5/iceberg1.5 artifact

cbb330:
## Summary
This is Part 1 of 2 for enabling compaction based MoR.

1. bring spark3.5 for creating delete files in testing
2. enable delete file compaction in spark3.1/spark3.5 and add unittests for such in spark3.5

problem: enabling Merge-on-Read for customer tables requires the ability to compact delete files. Compacting delete files is supported in spark3.1/iceberg1.2 but requires us to test it. We cannot test it, because creating delete files is not supported in spark3.1 (only compacting existing ones).

solution: we add spark3.5/iceberg1.5 compaction application, which extends the existing compaction test suite. in a follow up PR, we will add tests which create delete files and compacts the delete files under the latest versions. 

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [X] New Features
- [ ] Code Style
- [X] Refactoring
- [ ] Documentation
- [X] Tests

## Testing Done

- [X] Updated existing tests to reflect the changes made.

all tests running on spark3.1 and spark3.5:
image

correct spark version resolutions for test runtime:
image

correct spark version resolutions for runtime:
image

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


autumnust:
Is this ready for review?

cbb330:
> Is this ready for review?

@autumnust waiting on https://github.com/linkedin/openhouse/pull/221 to merge so i can rebase. this PR is forked from that.

settings.gradle
```diff
@@ -25,6 +25,7 @@ pluginManagement {
 rootProject.name = 'openhouse'
 
 include ':apps:spark'
+include ':apps:spark-3.5'
```
jiang95-dev, line 28: Can we also rename this module?

libs/datalayout/build.gradle
```diff
@@ -1,7 +1,6 @@
 plugins {
     id 'openhouse.java-conventions'
     id 'openhouse.hadoop-conventions'
-    id 'openhouse.iceberg-conventions-1.2'
```
jiang95-dev, line 4: Why removing this?

buildSrc/src/main/groovy/openhouse.apps-spark-conventions.gradle
```diff
@@ -0,0 +1,123 @@
+plugins {
```
jiang95-dev, line 1: Nit. The name `apps-spark-conventions` doesn't sound right. Should `apps-spark-commons` be better?

apps/spark/build.gradle
```diff
@@ -1,38 +1,49 @@
 plugins {
-    id 'openhouse.java-conventions'
-    id 'openhouse.hadoop-conventions'
-    id 'openhouse.iceberg-conventions-1.2'
```
jiang95-dev, line 4: This line should not be removed.

apps/spark-3.5/build.gradle
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    implementation "org.apache.logging.log4j:log4j-slf4j-impl:${log4jVersion}"
+    implementation(project(':libs:datalayout')) {
+        exclude group: 'org.apache.iceberg', module: 'iceberg-spark-runtime-3.1_2.12'
+    }
+    implementation("org.apache.iceberg:iceberg-bundled-guava:${icebergVersion}")
+    implementation("org.apache.iceberg:iceberg-data:${icebergVersion}")
+    implementation("org.apache.iceberg:iceberg-core:${icebergVersion}")
+    implementation("org.apache.iceberg:iceberg-common:${icebergVersion}")
```
jiang95-dev, line 32: Add `iceberg-conventions-1.5.2.gradle` to remove them.

jiang95-dev (changes requested):
Thanks for adding the new module! One major request is to use `iceberg-convention-1.2.gradle `and `iceberg-convention-1.5.2.gradle` to avoid duplicate code.

libs/datalayout/build.gradle
```diff
@@ -10,10 +9,18 @@ ext {
     sparkVersion = '3.1.1'
     springVersion = '2.7.8'
     hadoopVersion = '2.10.0'
+    sparkVersionSuffix = "3.1"
+    openhouseSparkRuntimeModule = ":integrations:spark:spark-${sparkVersionSuffix}:openhouse-spark-runtime_2.12"
+    tablesTestFixturesModule = ":tables-test-fixtures:tables-test-fixtures_2.12"
```
autumnust, line 14: are changes in this file relevant to this PR?

build.gradle
```diff
@@ -60,8 +60,6 @@ allprojects {
   if (it.path != ':integrations:spark:spark-3.5:openhouse-spark-3.5-itest') {
     configurations.all {
       resolutionStrategy {
-        force 'com.fasterxml.jackson:jackson-bom:2.13.4'
-        force 'com.fasterxml.jackson.core:jackson-databind:2.13.4'
```
autumnust, line 64: are these changes relevant ?

apps/spark/build.gradle
```diff
@@ -1,38 +1,49 @@
...
         exclude group: 'org.apache.hadoop', module: 'hadoop-common'
         exclude group: 'org.apache.hadoop', module: 'hadoop-client'
     }
+
+    implementation "org.apache.logging.log4j:log4j-slf4j-impl:${log4jVersion}"
     implementation project(':libs:datalayout')
+    implementation("org.apache.iceberg:iceberg-bundled-guava") {
+        version {
+            strictly("${icebergVersion}")
+        }
```
autumnust, line 30: are these really helpful ?  In practice I never find `force` or `strictly` covering everything that I want, since there could always be fatJar in the dependency Tree and these keywords don't work for them.

integrations/spark/spark-3.5/openhouse-spark-runtime/build.gradle
```diff
@@ -55,7 +55,9 @@ dependencies {
   fatJarPackagedDependencies(project(path: ':integrations:spark:spark-3.1:openhouse-spark-runtime_2.12', configuration: 'shadow')) {
     transitive = false
   }
-  implementation(project(path: ':integrations:spark:spark-3.1:openhouse-spark-runtime_2.12', configuration: 'shadow'))
+  implementation(project(path: ':integrations:spark:spark-3.1:openhouse-spark-runtime_2.12', configuration: 'shadow')) {
+    exclude group: "org.apache.iceberg", module: "iceberg-spark-runtime-3.1_2.12"
```
autumnust, line 59: I thought this one is shaded already?

autumnust (changes requested):

libs/datalayout/build.gradle
cbb330: this convention defines strictly 1.2. that strictly causes issue when used as a dependency in :apps:spark-3.5 which needs iceberg 1.5.

we can also do:
1. replace this line with 1.5 conventions, since this can be overriden by :apps:spark-3.1 since it isn't strictly. this is the same approach you used with services modules to 1.5 and overriding with strictly 1.2
2. exclude the transitive dependency from :apps:spark-3.5, but I would prefer not to do that and use the version resolver instead

libs/datalayout/build.gradle
cbb330: yes, in this PR i must unpin 1.2 from the convention plugin removed above. i do this by removing convention plugin, and defining explicitly here and without any strictly version constraints. 

this is so that consumers of this library (:apps) can override with their required version such as iceberg 1.5.2

build.gradle
cbb330: this may actually be dangerous to remove, and I will fix it, since idk what is currently relying on this.

but this strict version constraint breaks :apps:spark-3.5 because that also needs jackson 2.15

apps/spark/build.gradle
cbb330: strictly here overrides any dependencies of compaction that are defining a higher version of 1.5.2, such as 

services
cluster:storage
internalcatalog
htscatalog

an alternative to forcing the version to be LOWER than what is defined in transitive dependencies, is excluding all iceberg stuff from all of the dependencies above. but i think this is less readable and from what I understand, utilizing the resolution rules is preferred to exclusions

integrations/spark/spark-3.5/openhouse-spark-runtime/build.gradle
cbb330: it was supposed to be but i think it was missed. both 3.1 and 3.5 were available on classpath before this exclusion.

settings.gradle
cbb330: when i attempt to rename `:apps:spark-3.5`, to `:apps:openhouse-spark-apps-1.5_2.12`

```bash
A problem occurred evaluating project ':apps:openhouse-spark-apps-1.5_2.12'.
> Could not get unknown property 'sourceSets' for project ':apps:openhouse-spark-apps_2.12' of type org.gradle.api.Project.
```

which is... unexpected. i will leave the module the same name then because I attempted a few different alternatives without success

libs/datalayout/build.gradle
cbb330: compiling this module with spark 3.5/iceberg1.5 breaks :apps:spark3.1 

but compiling this module with spark3.1/iceberg1.2 works in :apps:spark3.5 (although unsafe, but this is ok since it won't be productionized until next year)

so leaving this :libs:datalayout as a nonstrict 1.2 is preferred , which requires removing iceberg-conventions-1.2

settings.gradle
cbb330: found the issue... gradle was unable to automatically make the two projects depend on each other. fixed

apps/spark/build.gradle
cbb330: changed to 2 from list instead

settings.gradle
```diff
@@ -62,10 +63,11 @@ include ':scripts:java:tools:dummytokens'
 
 project(':integrations:spark:spark-3.1:openhouse-spark-runtime').name = 'openhouse-spark-runtime_2.12'
 project(':apps:spark').name = 'openhouse-spark-apps_2.12'
+project(':apps:spark-3.5').name = 'openhouse-spark-apps-1.5_2.12'
```
teamurko, line 66:
```suggestion
project(':apps:spark-3.5').name = 'openhouse-apps_spark-3.5_iceberg-1.5_2.12'
```

libs/datalayout/build.gradle
teamurko: ok with this approach, i'm curious what it would take to build 2 artifacts from this module

jiang95-dev: As an alternative, we can keep 2 versions of this module. But since it is only used internally, we can leverage the auto version upgrade of gradle and exclude the iceberg-spark-runtime in `:apps:spark3.5`. I'm fine with this approach.

jiang95-dev (approved):
Thanks for addressing the comments!

libs/datalayout/build.gradle
cbb330: @teamurko ^ helps to answer your q. we would have two gradle build scripts. one common source code dir (or two code dirs if the source code must diverge significantly between spark3.1/spark3.5 compiled versions)

settings.gradle
cbb330: @teamurko this is inconsistent with the existing spark3.1 naming convention on line 65. do you think we should still do it?

teamurko: i see that, i guess we'll have to change both artifacts, we can do that in a separate PR if needed

teamurko (approved):

autumnust (approved):

PR #261, opened 2024-12-04
Refactor HTS listing api

jiang95-dev:
## Summary
Refactor HTS listing api to reduce bandwidth though mysql pushdown.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [x] Refactoring
- [ ] Documentation
- [ ] Tests

The service layer provide ability for general search, but in the validation layer we expose the minimal capability to table services to accommodate current use cases. We can relax it whenever there is a new use case.

Added new `HouseTable` mapping methods to only map `databaseId` to avoid NPE. Changed `OpenHouseInternalCatalog` to use a fake `tableId` to avoid NPE.

## Testing Done

Docker tests:
```
scala> spark.sql("create table d1.t1 (col string)")
res4: org.apache.spark.sql.DataFrame = []

scala> spark.sql("create table d1.t2 (col string)")
res1: org.apache.spark.sql.DataFrame = []

scala> spark.sql("create table d2.t1 (col string)")
res4: org.apache.spark.sql.DataFrame = []

scala> spark.sql("show databases").show
+---------+
|namespace|
+---------+
|       d1|
|       d2|
+---------+

scala> spark.sql("show tables in d1").show
+---------+---------+
|namespace|tableName|
+---------+---------+
|       d1|       t1|
|       d1|       t2|
+---------+---------+
```

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
@@ -84,8 +84,7 @@ public class OpenHouseInternalCatalog extends BaseMetastoreCatalog {
     }
     if (namespace.isEmpty()) {
       return StreamSupport.stream(houseTableRepository.findAll().spliterator(), false)
-          .map(
-              houseTable -> TableIdentifier.of(houseTable.getDatabaseId(), houseTable.getTableId()))
+          .map(houseTable -> TableIdentifier.of(houseTable.getDatabaseId(), "Unused"))
           .collect(Collectors.toList());
     }
     return houseTableRepository.findAllByDatabaseId(namespace.toString()).stream()
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
@@ -82,6 +82,8 @@ public class OpenHouseInternalCatalog extends BaseMetastoreCatalog {
       throw new ValidationException(
           "Input namespace has more than one levels " + String.join(".", namespace.levels()));
     }
+    // TODO: Implement SupportsNamespace interface and listNamespaces() method to remove this
+    //  branch. This is anti-pattern and only a temporary solution.
     if (namespace.isEmpty()) {
       return StreamSupport.stream(houseTableRepository.findAll().spliterator(), false)
           .map(houseTable -> TableIdentifier.of(houseTable.getDatabaseId(), "Unused"))
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/validator/impl/OpenHouseUserTableHtsApiValidator.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/validator/impl/OpenHouseUserTableHtsApiValidator.java
@@ -67,16 +67,17 @@ public class OpenHouseUserTableHtsApiValidator
               userTable.getDatabaseId(), ALPHA_NUM_UNDERSCORE_ERROR_MSG));
     }
 
-    if (userTable.getTableId() != null
-        && !userTable.getTableId().matches(ALPHA_NUM_UNDERSCORE_PATTERN_SEARCH_REGEX)) {
-      validationFailures.add(
-          String.format(
-              "tableId provided: %s, %s",
-              userTable.getTableId(), ALPHA_NUM_UNDERSCORE_PATTERN_SEARCH_ERROR_MSG));
-    }
+    if (userTable.getTableId() != null) {
+      if (userTable.getDatabaseId() == null) {
+        validationFailures.add("tableId cannot be provided without databaseId");
+      }
 
-    if (userTable.getDatabaseId() == null && userTable.getTableId() != null) {
-      validationFailures.add("tableId cannot be provided without databaseId");
+      if (!userTable.getTableId().matches(ALPHA_NUM_UNDERSCORE_PATTERN_SEARCH_REGEX)) {
+        validationFailures.add(
+            String.format(
+                "tableId provided: %s, %s",
+                userTable.getTableId(), ALPHA_NUM_UNDERSCORE_PATTERN_SEARCH_ERROR_MSG));
+      }
     }
 
     if (!validationFailures.isEmpty()) {
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/validator/impl/OpenHouseUserTableHtsApiValidator.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/validator/impl/OpenHouseUserTableHtsApiValidator.java
@@ -51,6 +51,14 @@ public class OpenHouseUserTableHtsApiValidator
   public void validateGetEntities(UserTable userTable) {
     List<String> validationFailures = new ArrayList<>();
 
+    // This will be removed when we start to support general filters
+    if (!(userTable.getTableVersion() == null
+        && userTable.getMetadataLocation() == null
+        && userTable.getStorageType() == null
+        && userTable.getCreationTime() == null)) {
+      validationFailures.add("Only databaseId and tableId are supported for the query");
+    }
+
     if (userTable.getDatabaseId() != null
         && !userTable.getDatabaseId().matches(ALPHA_NUM_UNDERSCORE_REGEX)) {
       validationFailures.add(
@@ -60,10 +68,15 @@ public class OpenHouseUserTableHtsApiValidator
     }
 
     if (userTable.getTableId() != null
-        && !userTable.getTableId().matches(ALPHA_NUM_UNDERSCORE_REGEX)) {
+        && !userTable.getTableId().matches(ALPHA_NUM_UNDERSCORE_PATTERN_SEARCH_REGEX)) {
       validationFailures.add(
           String.format(
-              "tableId provided: %s, %s", userTable.getTableId(), ALPHA_NUM_UNDERSCORE_ERROR_MSG));
+              "tableId provided: %s, %s",
+              userTable.getTableId(), ALPHA_NUM_UNDERSCORE_PATTERN_SEARCH_ERROR_MSG));
+    }
+
+    if (userTable.getDatabaseId() == null && userTable.getTableId() != null) {
+      validationFailures.add("tableId cannot be provided without databaseId");
     }
 
     if (!validationFailures.isEmpty()) {
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/config/MainApplicationConfig.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/config/MainApplicationConfig.java
@@ -2,6 +2,7 @@ package com.linkedin.openhouse.housetables.config;
 
 import com.linkedin.openhouse.cluster.metrics.TagUtils;
 import com.linkedin.openhouse.common.config.BaseApplicationConfig;
+import com.linkedin.openhouse.common.metrics.MetricsConstant;
 import io.micrometer.core.instrument.MeterRegistry;
 import org.springframework.boot.actuate.autoconfigure.metrics.MeterRegistryCustomizer;
 import org.springframework.context.annotation.Bean;
@@ -10,11 +11,12 @@ import org.springframework.context.annotation.Configuration;
 /** Main application configuration to load cluster properties and define required beans */
 @Configuration
 public class MainApplicationConfig extends BaseApplicationConfig {
-  public static final String APP_NAME = "housetables";
-
   @Bean
   MeterRegistryCustomizer<MeterRegistry> provideMeterRegistry() {
     return registry ->
-        registry.config().commonTags(TagUtils.buildCommonTag(clusterProperties, APP_NAME));
+        registry
+            .config()
+            .commonTags(
+                TagUtils.buildCommonTag(clusterProperties, MetricsConstant.HOUSETABLES_SERVICE));
   }
 }
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/config/MainApplicationConfig.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/config/MainApplicationConfig.java
@@ -11,6 +11,8 @@ import org.springframework.context.annotation.Configuration;
 /** Main application configuration to load cluster properties and define required beans */
 @Configuration
 public class MainApplicationConfig extends BaseApplicationConfig {
+  public static final String APP_NAME = "housetables";
+
   @Bean
   MeterRegistryCustomizer<MeterRegistry> provideMeterRegistry() {
     return registry ->
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
@@ -6,6 +6,7 @@ import com.linkedin.openhouse.housetables.model.UserTableRowPrimaryKey;
 import com.linkedin.openhouse.housetables.repository.HtsRepository;
 import java.util.Optional;
 import org.jetbrains.annotations.NotNull;
+import org.springframework.data.jpa.repository.Query;
 
 /**
  * JDBC-backed {@link HtsRepository} for CRUDing {@link UserTableRow}
@@ -32,6 +33,14 @@ public interface UserTableHtsJdbcRepository
   boolean existsByDatabaseIdIgnoreCaseAndTableIdIgnoreCase(String databaseId, String tableId);
 
   void deleteByDatabaseIdIgnoreCaseAndTableIdIgnoreCase(String databaseId, String tableId);
+
+  @Query("SELECT DISTINCT databaseId FROM UserTableRow")
+  Iterable<String> findAllDistinctDatabaseIds();
+
+  Iterable<UserTableRow> findAllByDatabaseIdIgnoreCase(String databaseId);
+
+  Iterable<UserTableRow> findAllByDatabaseIdAndTableIdLikeAllIgnoreCase(
+      String databaseId, String tableIdPattern);
 
   /*
    * The following methods are required to maintain the generality of the interface {@link com.linkedin.openhouse.housetables.repository.HtsRepository}
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
@@ -1,7 +1,9 @@
 package com.linkedin.openhouse.housetables.services;
 
+import com.linkedin.openhouse.cluster.metrics.micrometer.MetricsReporter;
 import com.linkedin.openhouse.common.exception.EntityConcurrentModificationException;
 import com.linkedin.openhouse.common.exception.NoSuchUserTableException;
+import com.linkedin.openhouse.common.metrics.MetricsConstant;
 import com.linkedin.openhouse.housetables.api.spec.model.UserTable;
 import com.linkedin.openhouse.housetables.dto.mapper.UserTablesMapper;
 import com.linkedin.openhouse.housetables.dto.model.UserTableDto;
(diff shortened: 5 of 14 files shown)


sumedhsakdeo:
@jiang95-dev please update the description with details of testing done section in local docker.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
```diff
@@ -33,6 +34,14 @@ Optional<UserTableRow> findByDatabaseIdIgnoreCaseAndTableIdIgnoreCase(
 
   void deleteByDatabaseIdIgnoreCaseAndTableIdIgnoreCase(String databaseId, String tableId);
 
+  @Query("SELECT DISTINCT databaseId FROM UserTableRow")
```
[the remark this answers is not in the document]
abhisheknath2011: Yes +1 adding index on mysql table.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/validator/impl/OpenHouseUserTableHtsApiValidator.java
```diff
@@ -51,6 +51,14 @@ public void validateDeleteEntity(UserTableKey userTableKey) {
   public void validateGetEntities(UserTable userTable) {
     List<String> validationFailures = new ArrayList<>();
 
+    // This will be removed when we start to support general filters
+    if (!(userTable.getTableVersion() == null
```
abhisheknath2011, line 55: What this filter condition is doing?

maluchari: +1. The condition does not seem intuitive

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/validator/impl/OpenHouseUserTableHtsApiValidator.java
```diff
@@ -60,10 +68,15 @@ public void validateGetEntities(UserTable userTable) {
     }
 
     if (userTable.getTableId() != null
```
maluchari, line 70: Can 70 to 79 be combined?

if (userTable.getTableId() != null) {
check databaseId not null;
check for pattern;
}

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/validator/impl/OpenHouseUserTableHtsApiValidator.java
jiang95-dev: It means all other fields are null. We don't have any use case to search by other fields so I want to restrict the capability of this API for now.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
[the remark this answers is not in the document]
jiang95-dev: Index is good. But it should be done in another PR.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/validator/impl/OpenHouseUserTableHtsApiValidator.java
jiang95-dev: Sure.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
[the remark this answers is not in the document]
cbb330: indexes are a resource stored in mysql. 

because of this requires something like a schema evolution, but AFAIK we dont execute those on the database.

I actually think the index MUST be in place before this PR, because while we may be saving in transport bytes to/from mysql, the CPU utilization will increase on mysql when mapping/reducing the scan of database_ids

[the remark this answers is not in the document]
cbb330: steps do this are here:

EXPLAIN SELECT DISTINCT databaseId FROM UserTableRow;

CREATE INDEX idx_databaseId ON UserTableRow(databaseId);

EXPLAIN SELECT DISTINCT databaseId FROM UserTableRow;


then post the results in this PR

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -46,12 +48,41 @@ public UserTableDto getUserTable(String databaseId, String tableId) {
 
   @Override
   public List<UserTableDto> getAllUserTables(UserTable userTable) {
```
cbb330, line 50: we are overloading this function with parameter/return objects that have non-required fields, parsing the fields and branching based on them. its a bit of code smell because we're not using function signatures/types and thus have to do more testing and logic.

I'm also concerned that we are adding features to a single API which end clients will build into their pipelines, and make it harder for us to go move away from this pattern in the future, e.g. requiring from api version migration

help me answer, should we refactor this? if not, should we only implement one new feature on this endpoint and implement the rest later on new endpoints? (distinct tables)

e.g.  break this monolith function into individual functions (and update the upstream routes accordingly)
`getDatabases()`
`getUserTables(Database database)`
`getUserTables(TablePattern tablePattern)`
`getUserTables(Database database, TablePattern tablePattern)`
`getUserTables()`

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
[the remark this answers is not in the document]
jiang95-dev: I tested on staging, and looks like it is already searching in the primary key space. Even after I added the index, it is still using the primary key. So we don't need to add the index which can add overhead to the writes.

Before:
```
mysql> EXPLAIN SELECT DISTINCT database_id FROM user_table_row;
+----+-------------+----------------+------------+-------+---------------+---------+---------+------+------+----------+--------------------------+
| id | select_type | table          | partitions | type  | possible_keys | key     | key_len | ref  | rows | filtered | Extra                    |
+----+-------------+----------------+------------+-------+---------------+---------+---------+------+------+----------+--------------------------+
|  1 | SIMPLE      | user_table_row | NULL       | range | PRIMARY       | PRIMARY | 514     | NULL |   25 |   100.00 | Using index for group-by |
+----+-------------+----------------+------------+-------+---------------+---------+---------+------+------+----------+--------------------------+
```
After
```
mysql> EXPLAIN SELECT DISTINCT database_id FROM user_table_row;
+----+-------------+----------------+------------+-------+-------------------------+---------+---------+------+------+----------+--------------------------+
| id | select_type | table          | partitions | type  | possible_keys           | key     | key_len | ref  | rows | filtered | Extra                    |
+----+-------------+----------------+------------+-------+-------------------------+---------+---------+------+------+----------+--------------------------+
|  1 | SIMPLE      | user_table_row | NULL       | range | PRIMARY,idx_database_id | PRIMARY | 514     | NULL |   24 |   100.00 | Using index for group-by |
+----+-------------+----------------+------------+-------+-------------------------+---------+---------+------+------+----------+--------------------------+
```

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
jiang95-dev: First, we want to avoid adding new endpoints or modifying the spec as much as possible. Keeping the api unchanged can keep clients from code changes.

Second, HTS is only exposed to tables and jobs service, so no need to worry about external clients or pipelines.

Third, branching will happen at some layer if not in the service layer. I can break it down to individual functions if that is better for readability, but I think the routing can be done here and no need for upstream changes.

autumnust: IIUC what you're proposing is to do the routing at the upper layer, like handler layer ? The ultimate problem is we are using `userTable` as the bucket to store all the filtering logic, so somewhere the routing logic needs to exist right ?  I +1 to break this code into separate method, but not exactly sure if we should move the routing to a different layer though. 

Following is from cursor: 


Comparison:
1. Single Responsibility Principle
Handler Routing: Handler has to know about routing rules, which is more of a business logic concern
Service Routing: ✅ Service handles all business logic, handler just transforms requests/responses

2. API Contract Clarity
Handler Routing: ✅ Makes it very clear at the API level what different combinations are supported
Service Routing: Implementation details are hidden, which could make API behavior less obvious

3. Testing
Handler Routing: Need to test routing logic at both handler and service levels
Service Routing: ✅ Routing logic tested in one place, handler tests are simpler

4. Maintainability
Handler Routing: Changes to routing logic might require updates in multiple places
Service Routing: ✅ Changes to routing logic contained within service layer

5. Reusability
Handler Routing: Other services would need to reimplement routing logic
Service Routing: ✅ Routing logic can be reused by other services/handlers

Recommendation:
I would actually lean towards Approach 2 (Service Routing) because:
It better adheres to separation of concerns
The service layer is the right place for business logic
It's more maintainable and reusable
The handler stays focused on its primary responsibility of API contract handling
The main tradeoff is that the API behavior becomes less obvious at the handler level, but this can be addressed with good documentation and clear method names in the service interface.

abhisheknath2011: I am align with keeping the routing logic at the service layer itself considering the current implementation. Would recommend applying some filter logic (Preferably separate search filter) to segregate different search use cases. We could come up with an Enum for different search use cases and use the corresponding enum based on search filter. Based on that we should be able to extract different type of search into separate methods. That should increase the readability and simplify the search conditions.

cbb330: i favor different features per handler in the API thats presented to end users. 

reason being that this follows the REST and single action / single resource paradigm that springboot builds around. for example, if one API handled route does 5 different services (conditional logic may execute one of 5 different algorithms), we won't find as much value from our springboot QPS metric on this route because we don't know which Query is hitting which condition in the service. certain queries to the same endpoint will have different behavior.

i recognize that implementing a new API per each type of feature is more refactoring / coding, so i'm suggesting that we only implement one feature on this API and delegate the rest to net new APIs and integrate them upstream / expose them.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -46,12 +48,41 @@ public UserTableDto getUserTable(String databaseId, String tableId) {
...
+              htsJdbcRepository
+                  .findAllByDatabaseIdAndTableIdLikeAllIgnoreCase(
+                      userTable.getDatabaseId(), userTable.getTableId())
+                  .spliterator(),
+              false)
+          .map(userTableRow -> userTablesMapper.toUserTableDto(userTableRow))
+          .collect(Collectors.toList());
+    } else {
+      // general search
+      log.warn(
```
maluchari, line 78: This should never happen since validation layer kind of covers this? also if these calls indeed happen due to a miss how are we planning to track them if we actually have handling? checking warn messages might not scale?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
```diff
@@ -33,6 +34,14 @@ Optional<UserTableRow> findByDatabaseIdIgnoreCaseAndTableIdIgnoreCase(
...
+  Iterable<String> findAllDistinctDatabaseIds();
+
+  Iterable<UserTableRow> findAllByDatabaseIdIgnoreCase(String databaseId);
+
+  Iterable<UserTableRow> findAllByDatabaseIdAndTableIdLikeAllIgnoreCase(
```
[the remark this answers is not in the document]
jiang95-dev: List tables with pattern in a database.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
autumnust: it boils down to what exactly is the API contract for `/hts/tables/query` -- is this a standalone contract that takes responsibility of the full spectrum from table-scan to db-level search, or it should be separated into multiple contracts / apis. If the latter, then implementing in handler-routing makes sense. 

We currently have two API contracts: 
- point query with both element of pk provided, 
- range query with elements of pk partially / fuzzily provided. 

I didn't see the need to change this contracts thus still favoring the way it is now. but +1 to break down the method into smaller ones in the service still.

sumedhsakdeo: >> REST and single action / single resource paradigm that springboot builds around

You are right that this is a general convention @cbb330 and we follow this in our table-service very strictly. For HTS API is modelled as a generic k-v store API, and hence the way it is. I agree I don't think we need to change the routing layer here.

cbb330: Ah I understand now the existing API contract. i'm aligned

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
jiang95-dev: Added metrics so we can set up alerts later.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -96,4 +106,67 @@ public void deleteUserTable(String databaseId, String tableId) {
...
+                .findAllByDatabaseIdAndTableIdLikeAllIgnoreCase(
+                    userTable.getDatabaseId(), userTable.getTableId())
+                .spliterator(),
+            false)
+        .map(userTableRow -> userTablesMapper.toUserTableDto(userTableRow))
+        .collect(Collectors.toList());
+  }
+
+  private List<UserTableDto> searchTables(UserTable userTable) {
+    METRICS_REPORTER.count(MetricsConstant.HTS_GENERAL_SEARCH_REQUEST);
```
autumnust, line 138: nice - let's remember to create dashboard for this after deployment

services/housetables/src/main/java/com/linkedin/openhouse/housetables/config/MainApplicationConfig.java
```diff
@@ -10,11 +11,12 @@
 /** Main application configuration to load cluster properties and define required beans */
 @Configuration
 public class MainApplicationConfig extends BaseApplicationConfig {
-  public static final String APP_NAME = "housetables";
```
autumnust, line 13: sure about no li-openhouse usage for this public static? This change doesn't seem to be very necessary regardless

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -96,4 +106,67 @@ public void deleteUserTable(String databaseId, String tableId) {
     htsJdbcRepository.deleteById(
         UserTableRowPrimaryKey.builder().databaseId(databaseId).tableId(tableId).build());
   }
+
+  private List<UserTableDto> listDatabases() {
+    return StreamSupport.stream(htsJdbcRepository.findAllDistinctDatabaseIds().spliterator(), false)
```
autumnust, line 111: given we already have metric object installed, shall we also time this method ? 
Not blocking.

autumnust (changes requested):

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
```diff
@@ -84,8 +84,7 @@ public List<TableIdentifier> listTables(Namespace namespace) {
     }
     if (namespace.isEmpty()) {
       return StreamSupport.stream(houseTableRepository.findAll().spliterator(), false)
-          .map(
-              houseTable -> TableIdentifier.of(houseTable.getDatabaseId(), houseTable.getTableId()))
+          .map(houseTable -> TableIdentifier.of(houseTable.getDatabaseId(), "Unused"))
```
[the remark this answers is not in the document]
jiang95-dev: It is not unintentional. `houseTable.getTableId()` will be null and `TableIdentifier.of` does not accept a null value.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/config/MainApplicationConfig.java
jiang95-dev: Sure will add it back.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
jiang95-dev: Sure.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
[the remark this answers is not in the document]
autumnust: what happen if some table is called `unused` ?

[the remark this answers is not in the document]
autumnust: this seems a bit weird though -- if the intentional is to use `HouseTable` but only the databaseId, should this be implemented within `listNamespaces` instead?

autumnust (changes requested):
last around comment: 
Having HouseTable object only with databaseId populated and use in listTables's implementation seems a bit weird to me.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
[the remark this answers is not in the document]
jiang95-dev: Ideally yes. But this class didn't implement SupportsNamespace. Do we want to add the interface?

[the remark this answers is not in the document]
jiang95-dev: > what happen if some table is called `unused` ?

"Unused" means the table name won't be used in the upstream.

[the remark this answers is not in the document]
autumnust: > Ideally yes. But this class didn't implement SupportsNamespace. Do we want to add the interface?

can we leave a TODO for implementing this when we approach db-as-primary-resources ? cc @maluchari  @Will-Lo

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
```diff
@@ -84,8 +84,7 @@ public List<TableIdentifier> listTables(Namespace namespace) {
     }
     if (namespace.isEmpty()) {
```
autumnust, line 85: I actually wonder if this syntax is necessary at all. what we were doing here is when db is not given just list everything ... but isn't that really an anti-pattern? cc @HotSushi

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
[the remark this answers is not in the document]
jiang95-dev: Added TODO.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
jiang95-dev: Added TODO.

autumnust (approved):
discussed offline -- will revisit the listNamespace API when we have db as primary resources. For the time being, will stay with the "unused" table as the hack since this is an internal API call

PR #266, opened 2024-12-04
Consider WebClientRequestException as a CSUE

HotSushi:
## Summary
Network cutoff between Client and Server surface as WebClientRequestException.

Following situation can arise:
- Client makes a docommit() call that succeeds
- Server responds success
- Network cuts off
- Client gets WebClientRequestException
- Client deletes snapshots and added data

The state of the table can end up corrupted.

### When is WebClientRequestException Thrown
WebClientRequestException is thrown only **if something goes wrong before response (even if failure) is completely received** see the [code block here](https://github.com/spring-projects/spring-framework/blob/5e939ccb4ba26243e765dd37f37eda90f90292d8/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/ExchangeFunctions.java#L106): 
```
return this.connector
		.connect(httpMethod, url, httpRequest -> clientRequest.writeTo(httpRequest, this.strategies))
		.doOnRequest(n -> logRequest(clientRequest))
		.doOnCancel(() -> logger.debug(clientRequest.logPrefix() + "Cancel signal (to close connection)"))
		.onErrorResume(WebClientUtils.WRAP_EXCEPTION_PREDICATE, t -> wrapException(t, clientRequest))     ====> this code throws WebClientRequestException.
		.map(httpResponse -> {
			String logPrefix = getLogPrefix(clientRequest, httpResponse);
			logResponse(httpResponse, logPrefix);
			return new DefaultClientResponse(
					httpResponse, this.strategies, logPrefix, httpMethod.name() + " " + url,
					() -> createRequest(clientRequest));
		});
```
^ This means the request could've been processed.

## Changes

- [ ] Internal API Changes
- [x] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

- [X] Updated existing tests to reflect the changes made.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
@@ -108,8 +108,7 @@ public class OpenHouseTableOperationsTest {
     when(mockTableApi.updateTableV1(anyString(), anyString(), any()))
         .thenReturn(Mono.error(mock(WebClientRequestException.class)));
     Assertions.assertThrows(
-        WebClientWithMessageException.class,
-        () -> openHouseTableOperations.doCommit(base, metadata));
+        CommitStateUnknownException.class, () -> openHouseTableOperations.doCommit(base, metadata));
     when(mockTableApi.updateTableV1(anyString(), anyString(), any()))
         .thenReturn(Mono.error(mock(WebClientResponseException.class)));
     Assertions.assertThrows(
--- a/integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
@@ -9,6 +9,7 @@ import com.linkedin.openhouse.gen.tables.client.model.Policies;
 import com.linkedin.openhouse.gen.tables.client.model.PolicyTag;
 import com.linkedin.openhouse.gen.tables.client.model.Retention;
 import com.linkedin.openhouse.javaclient.exception.WebClientWithMessageException;
+import com.linkedin.openhouse.relocated.org.springframework.http.HttpStatus;
 import com.linkedin.openhouse.relocated.org.springframework.web.reactive.function.client.WebClientRequestException;
 import com.linkedin.openhouse.relocated.org.springframework.web.reactive.function.client.WebClientResponseException;
 import com.linkedin.openhouse.relocated.reactor.core.publisher.Mono;
@@ -109,11 +110,20 @@ public class OpenHouseTableOperationsTest {
         .thenReturn(Mono.error(mock(WebClientRequestException.class)));
     Assertions.assertThrows(
         CommitStateUnknownException.class, () -> openHouseTableOperations.doCommit(base, metadata));
+    WebClientResponseException exception40x =
+        mock(WebClientResponseException.MethodNotAllowed.class);
+    when(exception40x.getStatusCode()).thenReturn(HttpStatus.METHOD_NOT_ALLOWED);
     when(mockTableApi.updateTableV1(anyString(), anyString(), any()))
-        .thenReturn(Mono.error(mock(WebClientResponseException.class)));
+        .thenReturn(Mono.error(exception40x));
     Assertions.assertThrows(
         WebClientWithMessageException.class,
         () -> openHouseTableOperations.doCommit(base, metadata));
+    WebClientResponseException exception50x = mock(WebClientResponseException.BadGateway.class);
+    when(exception50x.getStatusCode()).thenReturn(HttpStatus.BAD_GATEWAY);
+    when(mockTableApi.updateTableV1(anyString(), anyString(), any()))
+        .thenReturn(Mono.error(exception50x));
+    Assertions.assertThrows(
+        CommitStateUnknownException.class, () -> openHouseTableOperations.doCommit(base, metadata));
   }
 
   @Test
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
@@ -299,7 +299,8 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
       return Mono.error(
           new BadRequestException(
               casted, casted.getStatusCode().value() + " , " + casted.getResponseBodyAsString()));
-    } else if (e instanceof WebClientResponseException) {
+    } else if (e instanceof WebClientResponseException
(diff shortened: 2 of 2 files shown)


integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -301,8 +301,6 @@ static Mono<GetTableResponseBody> handleCreateUpdateHttpError(
               casted, casted.getStatusCode().value() + " , " + casted.getResponseBodyAsString()));
     } else if (e instanceof WebClientResponseException) {
       return Mono.error(new WebClientResponseWithMessageException((WebClientResponseException) e));
-    } else if (e instanceof WebClientRequestException) {
```
sumedhsakdeo, line 304: Let's double click on this as well while we are at it. 
WebClientResponseException

sumedhsakdeo: Why is that not CSUE?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -301,8 +301,6 @@ static Mono<GetTableResponseBody> handleCreateUpdateHttpError(
...
-      return Mono.error(new WebClientRequestWithMessageException((WebClientRequestException) e));
     } else {
       /**
        * This serves as a catch-all for any unexpected exceptions that could occur during doCommit,
```
maluchari, line 306: Can we also a comment to explicitly mention webClientRequestException here?

autumnust (changes requested):
Shall we also get some understanding on why error code 0 from ambassador --> `WebClientRequestException` and what's the nature of `WebClientRequestException` ?  

Alternatively I think there should be some way to preserve the original exception that get translated into CSUE. The ultimate goal is NOT to throw CSUE all the time as it "works", it works until it doesn't. the ultimate goal is to throw CSUE only when there's legit case for CSUE.

sumedhsakdeo:
> Shall we also get some understanding on why error code 0 from ambassador --> `WebClientRequestException` and what's the nature of `WebClientRequestException` ?  
> 
> 
> 
> Alternatively I think there should be some way to preserve the original exception that get translated into CSUE. The ultimate goal is NOT to throw CSUE all the time as it "works", it works until it doesn't. the ultimate goal is to throw CSUE only when there's legit case for CSUE. 

I disagree here. There is no reliable way to classify if a case where you would deem it is legit to throw CSUE. Network can fail in N different ways. We know why ambassador throws 0, we debugged it when it had first happened. There should be an internal slack thread on this.

I am ok with not removing webrequestexception if javadoc for it says it's guaranteed to be thrown before a network call is made.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
HotSushi: added

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
maluchari: We are converting 5xx errors from WebClientResponseException to CSUE above. This branch covers mainly other errors that could be 4xx.

sumedhsakdeo (approved):

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
autumnust: `NOT_IMPLEMENTED` this one should be safe to not categorized as CSUE, wdyt ?

teamurko: we have a branch above that is not needed due to catch all branch below..
seems like we can recognize non-CSUE when we get specific error code sent by the server which is safe, e.g. 4xx or not_implemented, all other scenarios should be CSUE because there is no response from server that guarantees failed commit

HotSushi: > NOT_IMPLEMENTED this one should be safe to not categorized as CSUE

it is a 50x error, with ambassador in the middle, we cannot completely rule out if it might get surfaced in some way. I feel safer categorizing it a CSUE and running analysis on what gets dumped there.  

> we have a branch above that is not needed due to catch all branch below.

Thats true, I prefer we still keep the branch above. Its helping me with readability of expected CSUE and unexpected CSUE

integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
```diff
@@ -108,13 +109,21 @@ public void testCreateUpdateTableErrorHandle() {
     when(mockTableApi.updateTableV1(anyString(), anyString(), any()))
         .thenReturn(Mono.error(mock(WebClientRequestException.class)));
```
autumnust, line 110: Let's have some comment regarding what does `WebClientRequestException` mean? 

Response from Claude: 

```


WebClientRequestException is thrown for network-level failures like:


Connection refused
DNS resolution failures
Timeouts
SSL/TLS issues


WebClientResponseException is thrown when you get a valid HTTP response but with error status codes:


4xx client errors (400, 401, 403, 404, etc.)
5xx server errors (500, 502, 503, 504, etc.)
```

autumnust (approved):
minor suggestion, nothing blocking. thanks @HotSushi

teamurko (approved):

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
abhisheknath2011: So we have covered all the cases 4xx and 5xx and for all unknown scenarios throwing CSUE although it is thrown for specific known cases and that should be fine to prevent data loss.

abhisheknath2011 (approved):

integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
HotSushi: let me know post merge. I'll create a separate PR.

PR #269, opened 2024-12-11
Introduce a separate GHA workflow for Pull Request Validation to validate the correct commit hashes

Will-Lo:
## Summary

This PR adds a new workflow which largely reuses the existing workflow for building/creating images, except that it runs on pull requests but also uses the correct hashes. The PR workflow also has some optimizations to cancel the previous test run on the same PR if a new commit is pushed since that test result is outdated.

pull request in 2-3 lines.

The workflow on the main branch appears to be using the merge commit SHA for tagging purposes: https://github.com/anothrNick/github-tag-action

However this causes issues where it can use outdated hashes of pull requests due to using `github.event.pull_request.merge_commit_sha` which does not seem consistent with the Pull Request HEAD SHA.

Documentation on the default behavior here that should fix it https://github.com/actions/checkout/blob/main/README.md

This can cause PRs to have breaking commits but due to this workflow file can pass the CI tests, making it difficult to work in the repository. Another side effect of the PR is that committers branches that have multiple commits may have difficulty debugging their failing workflows locally since the CI tests are running on outdated files.

See testing below for replication of the issue.

## Changes

- [ ] Internal API Changes
- [x] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

Tested locally on my repo
https://github.com/Will-Lo/openhouse/pull/1 shows a PR with the old behavior which has a clearly breaking change, but is passing tests since the commit that opened the PR passed tests

https://github.com/Will-Lo/openhouse/pull/3 shows a PR that used to pass with the commit that opened it, but now fails with the same breaking change.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


autumnust (changes requested):
Thanks for looking into this problem ! 
I may need some education on how this thing work after the change. left a comment

.github/workflows/build-test-pr.yml
```diff
@@ -0,0 +1,51 @@
+name: Gradle Build and Test Pull Requests to OpenHouse
```
dxichen, line 1: Could you help me understand why this needs to be separated in a different workflow? Is it because of the latency of the work queue for the build-tag-publish workflow?

Will-Lo: The main motivation is a better separation of concerns.

The original workflow's primary purpose is to test the branch, build the images and push them to docker and JFrog. However, in PRs it would skip the docker image generation and publishing step since that would only apply on the main branch.

My reasoning for splitting it up is that the logic of the two workflows seem to diverge quite a bit. For checkout on the main branch its recommended to use the merge commit hash I believe in scenarios where the commits are not squashed-then-merged (e.g. a rebase merge I believe) so that the merge commit is consistent for a single PR.

It's possible to write an if/else logic in the workflow and reuse them for both PRs and main branch but I find there are other benefits of splitting them up to, such as defining a concurrency policy that makes sense for PRs which would not make sense for a main branch push.

.github/workflows/build-test-pr.yml
```diff
@@ -0,0 +1,51 @@
...
+  group: ci-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-gradle-project:
+    name: Build Latest commit
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout project sources using merge commit hash
+        uses: actions/checkout@v4
```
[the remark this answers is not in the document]
Will-Lo: 1. Yes unless the main branch (the merge branch) gets pushed then the merge commit SHA is consistent which we don't want, we want it to checkout the PR Head SHA which is default.
2. Yes the original workflow will only run post-merge, I validated this on my own repository https://github.com/Will-Lo/openhouse/actions. Keeping it the same since there was some behavior so that the image tagging is consistent.

.github/workflows/build-test-pr.yml
dxichen: Thanks for the clarification, if the intention is the separate these out while they are still using mostly the same steps, lets modularize these actions (see https://devblogs.microsoft.com/ise/three-ways-to-simplify-cicd-pipelines-on-github-actions/) that way when we change how we want to build/test we only have to change it in a single place, while the build-tag-publish could leverage the testing flow and finish with artifactory publish steps

Will-Lo: Good catch I'll try that

dxichen (approved):
Cleanup looks great, thanks!

PR #262, opened 2024-12-04
Adds SQL support for Configurable Table Snapshot History

Will-Lo:
## Summary

Adds Spark SQL support for configurable table snapshots, which controls the versioning of the Openhouse tables.
Syntax is similar to retention but is instead defined as `HISTORY`.

History configuration supports both `MAX_AGE` and `VERSIONS`, where we retain all table snapshots that live within `MAX_AGE` and within `VERSIONS`.

Example: A table with `MAX_AGE = 1d` will retain all snapshots that are within 1 day of when the snapshot retention job last ran.
A table with `VERSIONS = 5` will retain the last 5 snapshots of the table without considering the age of the snapshots
If both `MAX_AGE = 1d` and `VERSIONS = 5` is defined, keep the last 5 snapshots within the last day. Note: If there are less than 5 snapshots, then there were less than 5 commits done in the past day.

`MAX_AGE` and `VERSIONS` cannot be defined as less than 1.
The default maximums of `MAX_AGE` and `VERSIONS` defined in https://github.com/linkedin/openhouse/pull/259 are 3 days and 100 versions respectively.

Examples:
```
ALTER TABLE <db>.<table> SET POLICY (HISTORY MAX_AGE=24H VERSIONS=10)
ALTER TABLE <db>.<table> SET POLICY (HISTORY MAX_AGE=3D)
ALTER TABLE <db>.<table> SET POLICY (HISTORY VERSIONS=5)
```

## Changes

- [x] Client-facing API Changes
- [x] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Performance Improvements
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

For all the boxes checked, please include additional details of the changes made in this pull request.  

## Testing Done

- [x] Manually Tested on local docker setup. Please include commands ran, and their output.
- [x] Added new tests for the changes made.
- [x] Updated existing tests to reflect the changes made.
- [ ] No tests added or updated. Please explain why. If unsure, please feel free to ask for help.
- [ ] Some other form of testing like staging or soak time in production. Please explain.

Tested on local docker running spark:
Tested setting both policies
```
scala> spark.sql("ALTER TABLE openhouse.db.tb2 SET POLICY ( HISTORY MAX_AGE=3d VERSIONS=3 )").show
++
||
++
++


scala> spark.sql("SHOW TBLPROPERTIES openhouse.db.tb2 (policies)").show(truncate=false)
[...]

PR #259, opened 2024-11-28
Add validators for History / Snapshot Retention Policy

Will-Lo:
## Summary

Supports defining a policy to control the versions of snapshots tables to retain.
The policy will support 2 types of configurations:
1. Time-based (e.g. 3 days)
2. Count-based (e.g. 10 versions)

It can also support a combination of the 2 policies, in which case it will retain versions that fall under both categories (e.g. only keep versions newer than 3 days AND within 10 versions).

This PR also refactors the PolicySpecValidator to accurately depict that it's referencing to partition-based retention of tables. In the future, we should refactor these policy classes to follow some `Policy` interface for better standardization for future policy support (duplicated interfaces between Replication, Partition Retention, Retain Snapshot).

This PR also defines the maximums for snapshot retention defaults (TODO: make this configurable/static?) 

**The current maximums defined is 3 days and 100 versions**

## Changes
- [ ] Client-facing API Changes
- [ ] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Performance Improvements
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

For all the boxes checked, please include additional details of the changes made in this pull request.  

## Testing Done

- [ ] Manually Tested on local docker setup. Please include commands ran, and their output.
- [ ] Added new tests for the changes made.
- [ ] Updated existing tests to reflect the changes made.
- [ ] No tests added or updated. Please explain why. If unsure, please feel free to ask for help.
- [ ] Some other form of testing like staging or soak time in production. Please explain.

For all the boxes checked, include a detailed description of the testing done for the changes made in this pull request.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations
- [ ] Large PR broken into smaller PRs, and PR plan linked in the description.

For all the boxes checked, include additional details of the changes made in this pull request.

PR #274, opened 2024-12-24
Support Table History Policy in Snapshot Expiration Job

Will-Lo:
## Summary

[[Issue](https://github.com/linkedin/openhouse/issues/#nnn)] Briefly discuss the summary of the changes made in this 
pull request in 2-3 lines.

Following up from https://github.com/linkedin/openhouse/pull/262 and https://github.com/linkedin/openhouse/pull/259
this PR adds support for snapshot expiration table maintenance job to use the history policies defined.

Most notably snapshot expiration will follow the settings of `maxAge`, `granularity`, and `versions` as follows:
1. If maxAge is provided, it will expire snapshots older than maxAge in granularity timeunit. 
2. If versions is provided, it will retain the last versions snapshots regardless of their age.
3. If both are provided, it will prioritize maxAge; only retain up to versions number of snapshots younger than the maxAge. This is done by pruning the snapshots older than maxAge, and then running a second expiration to keeping N versions after that. 

Note: If versions are defined and there are less than N versions in the history, then there were not enough commits (within that timespan if defined). Snapshot expiration will always keep at least 1 version.

The default behavior of snapshot expiration job will remain the same, keep snapshots within the last 3 days.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
@@ -208,9 +208,9 @@ public final class Operations implements AutoCloseable {
 
   /**
    * Expire snapshots on a given {@link Table}. If maxAge is provided, it will expire snapshots
-   * older than maxAge milliseconds. If versions is provided, it will retain the last versions
-   * snapshots. If both are provided, it will prioritize maxAge - only retain up to versions number
-   * of snapshots younger than the maxAge
+   * older than maxAge in granularity timeunit. If versions is provided, it will retain the last
+   * versions snapshots. If both are provided, it will prioritize maxAge; only retain up to versions
+   * number of snapshots younger than the maxAge
    */
   public void expireSnapshots(Table table, int maxAge, String granularity, int versions) {
     ExpireSnapshots expireSnapshotsCommand = table.expireSnapshots().cleanExpiredFiles(false);
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
@@ -201,16 +201,17 @@ public final class Operations implements AutoCloseable {
   }
 
   /** Expire snapshots on a given fully-qualified table name. */
-  public void expireSnapshots(String fqtn, long expireBeforeTimestampMs) {
-    expireSnapshots(getTable(fqtn), expireBeforeTimestampMs);
+  public void expireSnapshots(String fqtn, long expireBeforeTimestampMs, int minVersions) {
+    expireSnapshots(getTable(fqtn), expireBeforeTimestampMs, minVersions);
   }
 
   /** Expire snapshots on a given {@link Table}. */
-  public void expireSnapshots(Table table, long expireBeforeTimestampMs) {
+  public void expireSnapshots(Table table, long expireBeforeTimestampMs, int minVersions) {
     table
         .expireSnapshots()
         .cleanExpiredFiles(false)
         .expireOlderThan(expireBeforeTimestampMs)
+        .retainLast(minVersions)
         .commit();
   }
 
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
@@ -1,6 +1,5 @@
 package com.linkedin.openhouse.jobs.spark;
 
-import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import com.linkedin.openhouse.common.stats.model.IcebergTableStats;
 import com.linkedin.openhouse.jobs.util.SparkJobUtil;
@@ -21,6 +20,7 @@ import org.apache.hadoop.fs.LocatedFileStatus;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.RemoteIterator;
 import org.apache.iceberg.CatalogUtil;
+import org.apache.iceberg.ExpireSnapshots;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.StructLike;
 import org.apache.iceberg.Table;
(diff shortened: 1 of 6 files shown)


apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
```diff
@@ -201,17 +202,36 @@ public void deleteStagedOrphanDirectory(
...
+          System.currentTimeMillis() - timeUnitGranularity.toMillis(maxAge);
+      log.info("Expiring snapshots for table: {} older than {}ms", table, expireBeforeTimestampMs);
+      expireSnapshotsCommand.expireOlderThan(expireBeforeTimestampMs).commit();
+    }
+    if (versions > 0) {
+      log.info("Expiring snapshots for table: {} retaining last {} versions", table, versions);
+      // Note: retainLast keeps the last N snapshots that WOULD be expired, hence expireOlderThan
+      // currentTime
+      expireSnapshotsCommand
+          .expireOlderThan(System.currentTimeMillis())
```
autumnust, line 231: this is a bit unintuitive? For my knowledge: What will happen if older-than-now is not explicitly specified?

apps/spark/src/test/java/com/linkedin/openhouse/jobs/spark/OperationsTest.java
```diff
@@ -286,14 +322,81 @@ public void testSnapshotsExpirationJavaAPI() throws Exception {
...
+          snapshotIds.size(),
+          String.format("There must be %d snapshot(s) after inserts", numInserts));
+      Table table = ops.getTable(tableName);
+      log.info("Loaded table {}, location {}", table.name(), table.location());
+
+      ops.expireSnapshots(table, maxAge, timeGranularity, versionsToKeep);
       // verify that table object snapshots are updated
-      checkSnapshots(table, snapshotIds.subList(2, snapshotIds.size()));
+      checkSnapshots(
+          table, snapshotIds.subList(snapshotIds.size() - versionsToKeep, snapshotIds.size()));
```
autumnust, line 391: how does it test the semantics of favoring maxAge when both spec coexists? I assume we need to create N + M  snapshots in total, with N > versionToKeep and all snapshots among N are oldThanMaxAge, and after calling expiration job we will have all N expired.

autumnust (changes requested):

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
Will-Lo: If it's not explicitly specified, then no snapshots would be deleted by this SQL statement. From the javadoc:
```
[ExpireSnapshots](https://javadoc.io/static/org.apache.iceberg/iceberg-api/1.7.1/org/apache/iceberg/ExpireSnapshots.html) retainLast​(int numSnapshots)
Retains the most recent ancestors of the current snapshot.
If a snapshot would be expired because it is older than the expiration timestamp, but is one of the numSnapshots most recent ancestors of the current state, it will be retained. This will not cause snapshots explicitly identified by id from expiring.

This may keep more than numSnapshots ancestors if snapshots are added concurrently. This may keep less than numSnapshots ancestors if the current table state does not have that many.

Parameters:
numSnapshots - the number of snapshots to retain
Returns:
this for method chaining
```
https://javadoc.io/doc/org.apache.iceberg/iceberg-api/latest/org/apache/iceberg/ExpireSnapshots.html#retainLast(int)

I agree it's not intuitive but it seems that's how the java API is set up, in the official iceberg documents the alternative is to set write properties to forcefully maintain a certain number of versions of the table, but I think we wanted to not affect writers. e.g. `write.metadata.previous-versions-max`

apps/spark/src/test/java/com/linkedin/openhouse/jobs/spark/OperationsTest.java
Will-Lo: Thanks, I added some tests around this and found a bug where `retainLast` would throw an exception if the `versionToKeep > N`

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/SnapshotsExpirationSparkApp.java
```diff
@@ -18,41 +17,68 @@
...
 public class SnapshotsExpirationSparkApp extends BaseTableSparkApp {
   private final String granularity;
-  private final int count;
+  private final int maxAge;
+  private final int versions;
+
+  public static class DEFAULT_CONFIGURATION {
+    public static final String MAX_AGE = "3";
+    public static final String GRANULARITY = "DAYS";
+    public static final String VERSIONS = "0";
```
jiang95-dev, line 26: We can simply make `MAX_AGE` and `VERSIONS` int type to avoid type conversion.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/Operations.java
jiang95-dev: After looking into the iceberg code I found it not unintuitive.

Iceberg follows an **OR** strategy: Snapshots can either satisfy the maxAge or the versions restrictions. So when `expireOlderThan` is not explicitly defined, the maxAge is Infinite, and so all snapshots will be kept. When there are more snapshots in the versions than maxAge, those snapshots will also be kept.

On the contrary, Openhouse follows an **AND** strategy: Snapshots must satisfy both the maxAge and the versions restrictions. So it needs to call the iceberg API twice.

jiang95-dev (changes requested):
One small change request.

autumnust (approved):
thanks for adding the test cases for both configuration provided

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/SnapshotsExpirationSparkApp.java
```diff
@@ -18,41 +17,68 @@
...
+      String jobId,
+      StateManager stateManager,
+      String fqtn,
+      int maxAge,
+      String granularity,
+      int versions) {
     super(jobId, stateManager, fqtn);
-    this.granularity = granularity;
-    this.count = count;
+    if (maxAge == 0 && versions == 0) {
```
abhisheknath2011, line 42: What happens if maxAge and versions both are 0? Should we apply default policy in such case like expire 3 days worth of snapshot?

abhisheknath2011: I see that default configuration is set to 3 days. Sounds good.

abhisheknath2011 (approved):
Thanks @Will-Lo

jiang95-dev (approved):
LGTM

PR #284, opened 2025-02-04
Generate partition scope DLO strategy and persist to DLO table

jiang95-dev:
## Summary
Generate partition scope DLO strategy and persist to the partition-level DLO table. The new DLO table contains 2 new columns: partitionId and partitionColumns. The latter will be used only for analysis purpose and won't be used as filters in the execution app. At this stage, we generate both table-level and partition-level strategies for each table.

## Changes

- [ ] Client-facing API Changes
- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Performance Improvements
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

- [x] Manually Tested on local docker setup. Please include commands ran, and their output.
- [x] Added new tests for the changes made.
- [x] Updated existing tests to reflect the changes made.
- [ ] No tests added or updated. Please explain why. If unsure, please feel free to ask for help.
- [ ] Some other form of testing like staging or soak time in production. Please explain.

Tested on the test cluster. The dlo_partition_strategies table was created and has 1 row for each partition. The file reduction count for each partition is 1, and the partition_id and partition_columns had been correctly set.
```
scala> spark.sql("use openhouse")

scala> spark.sql("create table u_openhouse.dlo_run (ts timestamp, id int, data string) partitioned by (days(ts), id)").show()

scala> spark.sql("insert into u_openhouse.dlo_run values (current_timestamp(), 1, 'data')").show()
scala> spark.sql("insert into u_openhouse.dlo_run values (current_timestamp(), 1, 'data')").show()

scala> spark.sql("insert into u_openhouse.dlo_run values (date_add(current_timestamp(), 1), 1, 'data')").show()
scala> spark.sql("insert into u_openhouse.dlo_run values (date_add(current_timestamp(), 1), 1, 'data')").show()


scala> spark.sql("show tables in u_openhouse").show(100, false)
+-----------+------------------------+-----------+
|namespace  |tableName               |isTemporary|
+-----------+------------------------+-----------+
|u_openhouse|dlo_run                 |false      |
|u_openhouse|dlo_strategies          |false      |
|u_openhouse|dlo_partition_strategies|false      |
+-----------+------------------------+-----------+


scala> spark.sql("select * from u_openhouse.dlo_partition_strategies").show(false)
+-------------------+-------------+-----------------+----------------------+----------------------+------------------------------+----------------------+
[...]

PR #195, opened 2024-09-17
Downgrade hadoop to 2.8 for docker setup

teamurko:
## Summary
Maintenance apps happened to work on hadoop3 docker cluster, DLO apps didn't work due to incompatibility between source that depends on 2.10 and runtime jars on the cluster. DLO app example error:
```
2024-08-08 09:07:22 2024-08-08 16:07:22,005 INFO utils.LineBufferedStream: 2024-08-08 16:07:22,004 WARN scheduler.TaskSetManager: Lost task 0.0 in stage 0.0 (TID 0) (172.20.0.8 executor 0): java.lang.IllegalAccessError: class org.apache.hadoop.hdfs.web.HftpFileSystem cannot access its superinterface org.apache.hadoop.hdfs.web.TokenAspect$TokenManagementDelegator
2024-08-08 09:07:22 2024-08-08 16:07:22,005 INFO utils.LineBufferedStream:      at java.lang.ClassLoader.defineClass1(Native Method)
2024-08-08 09:07:22 2024-08-08 16:07:22,005 INFO utils.LineBufferedStream:      at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
2024-08-08 09:07:22 2024-08-08 16:07:22,005 INFO utils.LineBufferedStream:      at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
2024-08-08 09:07:22 2024-08-08 16:07:22,006 INFO utils.LineBufferedStream:      at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
2024-08-08 09:07:22 2024-08-08 16:07:22,006 INFO utils.LineBufferedStream:      at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
2024-08-08 09:07:22 2024-08-08 16:07:22,006 INFO utils.LineBufferedStream:      at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
2024-08-08 09:07:22 2024-08-08 16:07:22,006 INFO utils.LineBufferedStream:      at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
2024-08-08 09:07:22 2024-08-08 16:07:22,006 INFO utils.LineBufferedStream:      at java.security.AccessController.doPrivileged(Native Method)
2024-08-08 09:07:22 2024-08-08 16:07:22,006 INFO utils.LineBufferedStream:      at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
[...]

PR #287, opened 2025-02-15
Generate delete file statistics for DLO strategy

cbb330:
## Summary
**problem**: delete files are accumulating across tables due to trino default write behavior for UPDATE/MERGE/DELET statements

**solution**: 
1. (this PR) - record delete file count/size per table and persist to table ✅ 
2. let the DLO strategy execute for many days to accumulate data 🕐 
3. analyze the optimization algorithm and update the model to discover and act on tables who would benefit most from compacting their delete files 🕐 
4. (future PR) - codify the algorithm updates into DLO strategy 🕐 

**outcome**: databases accumulating delete files will be discovered and compacted appropriately to prevent degraded query impact


✨ **bonus content** ✨  :

I noticed there is a gap in DLO testing because manual validation must be done to verify the final form and contents of the DLO table, e.g. https://github.com/linkedin/openhouse/pull/284

so:
1. will a functional test into IntegrationTest by moving the DLO output table logic into StrategiesDAOInternal with save/load methods, and validated DLO finishes and outputs a table with contents and shape as expected 🕐 
2. will add an acceptance test in internal app to validate schema is evolved correctly on the actual tables 🕐 

this will help automate the testing for future changes and prevent bad changes from occuring due to validation being manually done

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [X] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

- [X] Manually Tested on local docker setup. Please include commands ran, and their output.

tested in local docker: 
setup commands
```bash
➜ docker compose -f infra/recipes/docker-compose/oh-hadoop-spark/docker-compose.yml up --build
➜ docker exec -it local.spark-master /bin/bash
bin/spark-shell --packages org.apache.iceberg:iceberg-spark-runtime-3.1_2.12:1.2.0 \
  --jars openhouse-spark-runtime_2.12-*-all.jar  \
  --conf spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions,com.linkedin.openhouse.spark.extensions.OpenhouseSparkSessionExtensions   \
  --conf spark.sql.catalog.openhouse=org.apache.iceberg.spark.SparkCatalog   \
  --conf spark.sql.catalog.openhouse.catalog-impl=com.linkedin.openhouse.spark.OpenHouseCatalog     \
  --conf spark.sql.catalog.openhouse.metrics-reporter-impl=com.linkedin.openhouse.javaclient.OpenHouseMetricsReporter    \
  --conf spark.sql.catalog.openhouse.uri=http://openhouse-tables:8080   \
  --conf spark.sql.catalog.openhouse.auth-token=$(cat /var/config/$(whoami).token) \
  --conf spark.sql.catalog.openhouse.cluster=LocalHadoopCluster
spark.sql("use openhouse");
spark.sql("drop table if exists u_openhouse.dlo_run_pre")
spark.sql("drop table if exists u_openhouse.dlo_run")

spark.sql("create table u_openhouse.dlo_run (ts timestamp, id int, data string) partitioned by (days(ts), id)").show()
spark.sql("INSERT INTO u_openhouse.dlo_run (ts, id, data) VALUES (current_timestamp(), 1, 'data'), (current_timestamp(), 2, 'data'), (current_timestamp(), 3, 'data'), (current_timestamp(), 1, 'data'), (current_timestamp(), 2, 'data'), (current_timestamp(), 3, 'data');")


spark.sql("select content, file_path, file_size_in_bytes, record_count, partition from u_openhouse.dlo_run.files").show(200, false)

spark.sql("show tables in u_openhouse").show(100, false)
docker compose -f infra/recipes/docker-compose/oh-hadoop-spark/docker-compose.yml --profile with_jobs_scheduler run openhouse-jobs-scheduler - --type DATA_LAYOUT_STRATEGY_GENERATION --cluster local --tablesURL http://openhouse-tables:8080 --jobsURL http://openhouse-jobs:8080 --tableMinAgeThresholdHours 0

spark.sql("show tblproperties u_openhouse.dlo_run ('write.data-layout.strategies')").show(200, false)
spark.sql("select * from u_openhouse.dlo_strategies").show(80,false)
```bash


```bash
scala> spark.sql("show tblproperties u_openhouse.dlo_run ('write.data-layout.strategies')").show(200, false)
+----------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|key                         |value                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
+----------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|write.data-layout.strategies|[{"score":9.999985377015197,"entropy":2.77080843971934816E17,"cost":0.5000007311503093,"gain":5.0,"config":{"targetByteSize":526385152,"minByteSizeRatio":0.75,"maxByteSizeRatio":10.0,"minInputFiles":5,"maxConcurrentFileGroupRewrites":5,"partialProgressEnabled":true,"partialProgressMaxCommits":1,"maxFileGroupSizeBytes":107374182400,"deleteFileThreshold":2147483647},"posDeleteFileCount":0,"eqDeleteFileCount":0,"posDeleteFileBytes":0,"eqDeleteFileBytes":0,"posDeleteRecordCount":0,"eqDeleteRecordCount":0}]|
+----------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+


scala> spark.sql("select * from u_openhouse.dlo_strategies").show(80,false)
+------------------------------------+-----------------------+----------------------+------------------------------+----------------------+---------------------+--------------------+---------------------+--------------------+-----------------------+----------------------+
|fqtn                                |timestamp              |estimated_compute_cost|estimated_file_count_reduction|file_size_entropy     |pos_delete_file_count|eq_delete_file_count|pos_delete_file_bytes|eq_delete_file_bytes|pos_delete_record_count|eq_delete_record_count|
+------------------------------------+-----------------------+----------------------+------------------------------+----------------------+---------------------+--------------------+---------------------+--------------------+-----------------------+----------------------+
|u_openhouse.dlo_partition_strategies|2025-02-27 08:03:30.822|0.500001              |1.0                           |2.77079875424947968E17|0                    |0                   |0                    |0                   |0                      |0                     |
|u_openhouse.dlo_run                 |2025-02-27 07:11:10.589|0.5                   |3.0                           |2.77080843971934848E17|0                    |0                   |0                    |0                   |0                      |0                     |
|u_openhouse.dlo_run                 |2025-02-27 08:03:03.454|0.500001              |5.0                           |2.77080843971934816E17|0                    |0                   |0                    |0                   |0                      |0                     |
|u_openhouse.dlo_strategies          |2025-02-27 08:03:54.471|0.5                   |0.0                           |2.7708015546118544E17 |0                    |0                   |0                    |0                   |0                      |0                     |
+------------------------------------+-----------------------+----------------------+------------------------------+----------------------+---------------------+--------------------+---------------------+--------------------+-----------------------+----------------------+
```

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkApp.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkApp.java
@@ -80,7 +80,7 @@ public class DataLayoutStrategyGeneratorSparkApp extends BaseTableSparkApp {
         if (isPartitionScope) {
           rows.add(
               String.format(
-                  "('%s', '%s', '%s', current_timestamp(), %f, %f, %f, %d, %d, %d, %d)",
+                  "('%s', '%s', '%s', current_timestamp(), %f, %f, %f, %d, %d, %d, %d, %d, %d)",
                   fqtn,
                   strategy.getPartitionId(),
                   strategy.getPartitionColumns(),
@@ -90,11 +90,13 @@ public class DataLayoutStrategyGeneratorSparkApp extends BaseTableSparkApp {
                   strategy.getPosDeleteFileCount(),
                   strategy.getEqDeleteFileCount(),
                   strategy.getPosDeleteFileBytes(),
-                  strategy.getEqDeleteFileBytes()));
+                  strategy.getEqDeleteFileBytes(),
+                  strategy.getPosDeleteRecordCount(),
+                  strategy.getEqDeleteRecordCount()));
         } else {
           rows.add(
               String.format(
-                  "('%s', current_timestamp(), %f, %f, %f, %d, %d, %d, %d)",
+                  "('%s', current_timestamp(), %f, %f, %f, %d, %d, %d, %d, %d, %d)",
                   fqtn,
                   strategy.getCost(),
                   strategy.getGain(),
@@ -102,7 +104,9 @@ public class DataLayoutStrategyGeneratorSparkApp extends BaseTableSparkApp {
                   strategy.getPosDeleteFileCount(),
                   strategy.getEqDeleteFileCount(),
                   strategy.getPosDeleteFileBytes(),
-                  strategy.getEqDeleteFileBytes()));
+                  strategy.getEqDeleteFileBytes(),
+                  strategy.getPosDeleteRecordCount(),
+                  strategy.getEqDeleteRecordCount()));
         }
       }
       String strategiesInsertStmt =
@@ -128,7 +132,9 @@ public class DataLayoutStrategyGeneratorSparkApp extends BaseTableSparkApp {
                   + "pos_delete_file_count LONG, "
                   + "eq_delete_file_count LONG, "
                   + "pos_delete_file_bytes LONG, "
-                  + "eq_delete_file_bytes LONG"
+                  + "eq_delete_file_bytes LONG,"
+                  + "pos_delete_record_count LONG, "
+                  + "eq_delete_record_count LONG"
                   + ") "
                   + "PARTITIONED BY (days(timestamp))",
               outputFqtn));
@@ -144,7 +150,9 @@ public class DataLayoutStrategyGeneratorSparkApp extends BaseTableSparkApp {
                   + "pos_delete_file_count LONG, "
                   + "eq_delete_file_count LONG, "
                   + "pos_delete_file_bytes LONG, "
-                  + "eq_delete_file_bytes LONG "
+                  + "eq_delete_file_bytes LONG,"
+                  + "pos_delete_record_count LONG, "
+                  + "eq_delete_record_count LONG"
                   + ") "
                   + "PARTITIONED BY (days(timestamp))",
               outputFqtn));
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkApp.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkApp.java
@@ -80,18 +80,29 @@ public class DataLayoutStrategyGeneratorSparkApp extends BaseTableSparkApp {
         if (isPartitionScope) {
           rows.add(
               String.format(
-                  "('%s', '%s', '%s', current_timestamp(), %f, %f, %f)",
+                  "('%s', '%s', '%s', current_timestamp(), %f, %f, %f, %d, %d, %d, %d)",
                   fqtn,
                   strategy.getPartitionId(),
                   strategy.getPartitionColumns(),
                   strategy.getCost(),
                   strategy.getGain(),
-                  strategy.getEntropy()));
+                  strategy.getEntropy(),
+                  strategy.getPosDeleteFileCount(),
+                  strategy.getEqDeleteFileCount(),
+                  strategy.getPosDeleteFileBytes(),
+                  strategy.getEqDeleteFileBytes()));
         } else {
           rows.add(
               String.format(
-                  "('%s', current_timestamp(), %f, %f, %f)",
-                  fqtn, strategy.getCost(), strategy.getGain(), strategy.getEntropy()));
+                  "('%s', current_timestamp(), %f, %f, %f, %d, %d, %d, %d)",
+                  fqtn,
+                  strategy.getCost(),
+                  strategy.getGain(),
+                  strategy.getEntropy(),
+                  strategy.getPosDeleteFileCount(),
+                  strategy.getEqDeleteFileCount(),
+                  strategy.getPosDeleteFileBytes(),
+                  strategy.getEqDeleteFileBytes()));
         }
       }
       String strategiesInsertStmt =
@@ -113,7 +124,11 @@ public class DataLayoutStrategyGeneratorSparkApp extends BaseTableSparkApp {
                   + "timestamp TIMESTAMP, "
                   + "estimated_compute_cost DOUBLE, "
                   + "estimated_file_count_reduction DOUBLE, "
-                  + "file_size_entropy DOUBLE "
+                  + "file_size_entropy DOUBLE, "
+                  + "pos_delete_file_count LONG, "
+                  + "eq_delete_file_count LONG, "
+                  + "pos_delete_file_bytes LONG, "
+                  + "eq_delete_file_bytes LONG"
                   + ") "
                   + "PARTITIONED BY (days(timestamp))",
               outputFqtn));
@@ -125,7 +140,11 @@ public class DataLayoutStrategyGeneratorSparkApp extends BaseTableSparkApp {
                   + "timestamp TIMESTAMP, "
                   + "estimated_compute_cost DOUBLE, "
                   + "estimated_file_count_reduction DOUBLE, "
-                  + "file_size_entropy DOUBLE "
+                  + "file_size_entropy DOUBLE, "
+                  + "pos_delete_file_count LONG, "
+                  + "eq_delete_file_count LONG, "
+                  + "pos_delete_file_bytes LONG, "
+                  + "eq_delete_file_bytes LONG "
                   + ") "
                   + "PARTITIONED BY (days(timestamp))",
               outputFqtn));
--- a/libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/FileStat.java
+++ b/libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/FileStat.java
@@ -16,5 +16,6 @@ public final class FileStat {
   private FileContent content;
   private String path;
   private long sizeInBytes;
+  private long recordCount;
   private List<String> partitionValues;
 }
--- a/libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/FileStat.java
+++ b/libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/FileStat.java
@@ -5,6 +5,7 @@ import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+import org.apache.iceberg.FileContent;
 
 /** Represents the statistics of a file. */
 @Data
@@ -12,7 +13,8 @@ import lombok.NoArgsConstructor;
 @NoArgsConstructor
 @AllArgsConstructor
 public final class FileStat {
+  private FileContent content;
   private String path;
-  private long size;
+  private long sizeInBytes;
   private List<String> partitionValues;
 }
--- a/libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/TableFileStats.java
+++ b/libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/TableFileStats.java
@@ -27,14 +27,14 @@ public class TableFileStats implements DataSource<FileStat> {
       return spark
           .sql(
               String.format(
-                  "SELECT content, file_path, file_size_in_bytes, partition FROM %s.data_files",
+                  "SELECT content, file_path, file_size_in_bytes, record_count, partition FROM %s.data_files",
                   tableName))
           .map(new FileStatMapper(), Encoders.bean(FileStat.class));
     } catch (IllegalArgumentException e) {
       return spark
           .sql(
               String.format(
-                  "SELECT content, file_path, file_size_in_bytes, null FROM %s.data_files",
+                  "SELECT content, file_path, file_size_in_bytes, record_count, null FROM %s.data_files",
                   tableName))
           .map(new FileStatMapper(), Encoders.bean(FileStat.class));
     }
@@ -44,7 +44,7 @@ public class TableFileStats implements DataSource<FileStat> {
     @Override
     public FileStat call(Row row) {
       List<String> partitionValues = new ArrayList<>();
-      Row partition = row.getStruct(3);
+      Row partition = row.getStruct(4);
       if (partition != null) {
         for (int i = 0; i < partition.size(); i++) {
           partitionValues.add(Objects.toString(partition.get(i)));
@@ -54,6 +54,7 @@ public class TableFileStats implements DataSource<FileStat> {
           .content(fromId(row.getInt(0)))
           .path(row.getString(1))
           .sizeInBytes(row.getLong(2))
+          .recordCount(row.getLong(3))
           .partitionValues(partitionValues)
           .build();
     }
--- a/libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/TableFileStats.java
+++ b/libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/TableFileStats.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import lombok.Builder;
+import org.apache.iceberg.FileContent;
 import org.apache.spark.api.java.function.MapFunction;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Encoders;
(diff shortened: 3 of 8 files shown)


apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkApp.java
```diff
@@ -80,18 +80,29 @@ private void appendToDloStrategiesTable(
...
                   strategy.getPartitionId(),
                   strategy.getPartitionColumns(),
                   strategy.getCost(),
                   strategy.getGain(),
-                  strategy.getEntropy()));
+                  strategy.getEntropy(),
+                  strategy.getPosDeleteFileCount(),
+                  strategy.getEqDeleteFileCount(),
+                  strategy.getPosDeleteFileBytes(),
+                  strategy.getEqDeleteFileBytes()));
```
teamurko, line 93: Should it be a nested struct?

apps/spark/src/test/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkAppTest.java
```diff
@@ -0,0 +1,37 @@
+package com.linkedin.openhouse.jobs.spark;
+
+import com.linkedin.openhouse.jobs.util.OtelConfig;
+import com.linkedin.openhouse.tablestest.OpenHouseSparkITest;
+import io.opentelemetry.api.metrics.Meter;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Test;
+
+@Slf4j
+public class DataLayoutStrategyGeneratorSparkAppTest extends OpenHouseSparkITest {
```
teamurko, line 10: +1 on having a test, should this be in OperationsTest.java?

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/FileStat.java
```diff
@@ -12,7 +12,8 @@
 @NoArgsConstructor
 @AllArgsConstructor
 public final class FileStat {
+  private int content;
```
teamurko, line 15: can we use content enum instead of id?

teamurko (changes requested):
Thank you, Christian. Looks good! Left a few questions

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/FileStat.java
cbb330: good idea done

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkApp.java
cbb330: I see that these values are related, but a con is that the a struct would make ser/de have more steps. 

is there another pro?

apps/spark/src/test/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkAppTest.java
cbb330: I took a chunk of the business logic from the DLO SparkApp related to serializing the strategy into the openhouse output table, and pushed it into the /libs folder via StrategiesDAOInternal save and load methods.

so now all of the schemas and save/load methods are in /libs module together. result is that it only required slightly more code added to the existing unittests which validates these new changes, and doesn't require adding to `OperationsTest.java` or ``DataLayoutStrategyGeneratorSparkAppTest.java`

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkApp.java
```diff
@@ -57,79 +58,28 @@ private void runInnerTableScope(
...
-                  + "fqtn STRING, "
-                  + "timestamp TIMESTAMP, "
-                  + "estimated_compute_cost DOUBLE, "
-                  + "estimated_file_count_reduction DOUBLE, "
-                  + "file_size_entropy DOUBLE "
-                  + ") "
-                  + "PARTITIONED BY (days(timestamp))",
-              outputFqtn));
-    }
+    // TODO: is StrategiesDaoTableProps supposed to be here?
```
cbb330, line 75: @jiang95-dev did we forget to call StrategiesDaoTableProps.save in the partitioned table condition here? or was it intentional

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/persistence/StrategiesDaoInternal.java
```diff
@@ -0,0 +1,131 @@
+package com.linkedin.openhouse.datalayout.persistence;
```
sumedhsakdeo, line 1: Can we do DAO refactor into a separate PR and not mix with the delete file stats PR?

sumedhsakdeo: In fact let's do the delete file stats PR first, and then do DAO refactor, will need to think more about it and will take longer for that to land.

sumedhsakdeo (changes requested):
Reduce scope of this PR to not include DAO refactor.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/persistence/StrategiesDaoInternal.java
cbb330: ok, will take this out of this PR,

> will need to think more about it and will take longer for that to land.
I'd like to follow up with the refactor shortly after this PR deploys and works. whats the concern? the refactor is intended to not change behavior

sumedhsakdeo: No major concern.

apps/spark/src/test/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkAppTest.java
cbb330: while this test worked in this PR, i'll take the test and add it in a future PR

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkApp.java
cbb330: discussed offline, it was intentional.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkApp.java
```diff
@@ -80,18 +80,29 @@ private void appendToDloStrategiesTable(
...
+                  strategy.getPosDeleteFileCount(),
+                  strategy.getEqDeleteFileCount(),
+                  strategy.getPosDeleteFileBytes(),
+                  strategy.getEqDeleteFileBytes()));
         } else {
           rows.add(
               String.format(
-                  "('%s', current_timestamp(), %f, %f, %f)",
-                  fqtn, strategy.getCost(), strategy.getGain(), strategy.getEntropy()));
+                  "('%s', current_timestamp(), %f, %f, %f, %d, %d, %d, %d)",
```
sumedhsakdeo, line 97: %d for bytes? Do you see a risk of overflow?

cbb330: good question, no it should not. 

initial answers say %d is sufficient for Long.MAX_VALUE

diving deeper is not a clear code/doc answer. so I tested it a range of very large numbers:
https://replit.com/join/xeprpdjctd-chrisbush747

sumedhsakdeo: cool, make sure you test the sql execution as well.

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/FileStat.java
```diff
@@ -5,14 +5,16 @@
...
 /** Represents the statistics of a file. */
 @Data
 @Builder
 @NoArgsConstructor
 @AllArgsConstructor
 public final class FileStat {
+  private FileContent content;
   private String path;
-  private long size;
+  private long sizeInBytes;
```
sumedhsakdeo, line 18: Is this serialized as size into tblproperty?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/DataLayoutStrategyGeneratorSparkApp.java
cbb330: sure thing! 

this code path isn't tested in main

i added [unittests](https://github.com/linkedin/openhouse/pull/287/commits/5a0ec3c53b840a81ad1d684c35c5364a10f689ca#diff-0b6d3a9ac0905fecf8a43e01a3390f52b15cd8488c5f927305a5c653db9b4ceeL86), they were green, but it required a bit of refactor, which we chose to remove in this PR

I will follow up with the tests again shortly after this PR is deployed

in the follow up, i will also ensure the LONG.MAX_VALUE is passed from pojo, string format, sql, and read back correctly

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/FileStat.java
cbb330: no, this value is only used in intermediate processing before being computed for DataLayoutStrategy (DLS) object. the DLS object is then serialized into tblproperty 

here's the intermediate processing: https://github.com/linkedin/openhouse/pull/287/files#diff-454fe015ffc81fffb4f558b25e865396a81ebca0f3b2444e422d3379cf1ccf29R149

and here is where DLS is serialized into tblproperty: https://github.com/linkedin/openhouse/blob/5a0ec3c53b840a81ad1d684c35c5364a10f689ca/libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/persistence/StrategiesDaoTableProps.java#L32

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/TableFileStats.java
```diff
@@ -41,17 +44,27 @@ static class FileStatMapper implements MapFunction<Row, FileStat> {
...
-          .path(row.getString(0))
-          .size(row.getLong(1))
+          .content(fromId(row.getInt(0)))
+          .path(row.getString(1))
+          .sizeInBytes(row.getLong(2))
           .partitionValues(partitionValues)
           .build();
     }
+
+    public static FileContent fromId(int id) {
```
teamurko, line 61:
```suggestion
    private static FileContent fromId(int id) {
```

libs/datalayout/src/test/java/com/linkedin/openhouse/datalayout/e2e/IntegrationTest.java
```diff
@@ -30,8 +30,9 @@ public void testCompactionStrategyGenerationWithPersistencePartitioned() throws
               .tableFileStats(tableFileStats)
               .tablePartitionStats(tablePartitionStats)
               .build();
-      List<DataLayoutStrategy> strategies = strategyGenerator.generateTableLevelStrategies();
-      Assertions.assertEquals(1, strategies.size());
+      List<DataLayoutStrategy> strategies = strategyGenerator.generatePartitionLevelStrategies();
+      Assertions.assertEquals(10, strategies.size());
+
```
teamurko, line 35: why is this change?

teamurko (commented):
Thank you, looks clean, just one more question

libs/datalayout/src/test/java/com/linkedin/openhouse/datalayout/e2e/IntegrationTest.java
cbb330: doing a bit of side cleaning, since this is a bug

1. the test is for a partitioned table
2. the strategy generated at the end is a non-partitioned table

so i'm fixing it to generate a strategy for partitioned table

teamurko (approved):

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/datasource/FileStat.java
sumedhsakdeo: We might want to make this class in future PR https://github.com/linkedin/openhouse/blob/d8ecafe498e2b27c9fb4ff2b7571c60c3c1f6644/libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/strategy/DataLayoutStrategy.java#L10 implement Serializable 
fyi.. @teamurko

sumedhsakdeo:
@cbb330 I don't see description include local docker testing. Is it possible to have delete files be generated in local docker and show that data is DLO strategy generator shows the right values?

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutStrategyGenerator.java
```diff
@@ -142,9 +146,16 @@ private Optional<DataLayoutStrategy> buildDataLayoutStrategy(
...
             .score(reducedFileCountPerComputeGbHr)
-            .entropy(computeEntropy(fileSizes))
+            .entropy(
+                computeEntropy(
+                    dataFiles.map(
+                        (MapFunction<FileStat, Long>) FileStat::getSizeInBytes, Encoders.LONG())))
             .partitionId(partitionValues)
             .partitionColumns(partitionColumns)
+            .posDeleteFileBytes(posDeleteStats._1)
+            .eqDeleteFileBytes(eqDeleteStats._1)
```
sumedhsakdeo, line 156: I don't think position and equality delete file bytes is super interesting stat, by nature. I would remove it and rather capture delete file `record_count` here. This would be good signal for DLO to know how many deleted records we are sitting on that need to be compacted and go away.
CC: @teamurko thoughts?

sumedhsakdeo (commented):
💯 getting there! Left a couple of comments.

cbb330:
> @cbb330 I don't see description include local docker testing. Is it possible to have delete files be generated in local docker and show that data is DLO strategy generator shows the right values?

@sumedhsakdeo i updated description -- but i was unable to create delete files using spark3.1. all we see now is 0 for these fields

libs/datalayout/src/main/java/com/linkedin/openhouse/datalayout/generator/OpenHouseDataLayoutStrategyGenerator.java
teamurko: I agree. For the purpose of estimating # removed rows, eq-delete file size doesn't tell anything, pos-delete row count is more useful than byte count.

cbb330:
@sumedhsakdeo spark3.5 doesn't ship with hadoop 2.8 https://archive.apache.org/dist/spark/spark-3.5.2/

DLO requires hadoop 2.8 to test, ref: https://github.com/linkedin/openhouse/pull/195

I can solve this with both spark3.1 and spark3.5 e2e validation in a new PR, including:
- test table with delete files
- jobs scheduler/app start up
- DLO target the test table
- Inspect DLO afterwords

by writing code into here: https://github.com/linkedin/openhouse/commit/5867db2948cf71b254054a0d73da03ac438fafed#diff-4a38c63e12296739521a9524b4d527521023ab3b49302a5e77d5af1441b2d4c9R16

either before or after merging this one

sumedhsakdeo:
Got it. Thanks.

sumedhsakdeo (approved):
Tests should also check for non-zero delete stats. But I am fine doing it in a followup PR

PR #299, opened 2025-03-23
Adding new API endpoint for lock table feature

rohitkum2506:
## Summary

Introducing API for locking a table.
```
API:
POST  /v1/tables/{tableId}/lock
Optional Request Params are message, creationTime, expirationTimeInDays: 
{
   locked: <boolean>
   message: <string>,
   creationTime: <long>,
   expirationTimeInDays: <int>
}
```
Lock state will be persisted as part of policies on table and can be read as `tblProperties`
New table will be in unlocked state by default. lock state is not stored in table policies on a newly created table.

Maintains consistency with existing format where policies on a table is null unless explicitly added.

Next PR:
1. Add delete lock API
2. Add Permission model on lock API which should allow only specific  user to perform locking/unlocking
3. Add permission model to reserve read/write to predefined user groups when table is locked.
4. Add SQL support

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done
Added unit tests
Local Docker testing

Patch LOCK api usage:
Screenshot 2025-03-26 at 2 04 14 PM

Get table shows lock state in policies

Screenshot 2025-03-26 at 2 04 05 PM

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/TablesApiHandler.java
```diff
@@ -108,4 +109,17 @@ ApiResponse<GetAclPoliciesResponseBody> getAclPolicies(
...
       String databaseId, String tableId, String actingPrincipal, String userPrincipal);
+
+  /**
+   * @param databaseId
+   * @param tableId
+   * @param updateLockRequestBody
+   * @param tableCreatorUpdator
+   * @return empty body on successful update
+   */
+  ApiResponse<Void> updateLock(
```
abhisheknath2011, line 120: I believe have a common update lock method for locking and unlocking. As there are two states how about `toggleLock`?

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseTablesApiHandler.java
```diff
@@ -140,4 +141,15 @@ public ApiResponse<GetAclPoliciesResponseBody> getAclPoliciesForUserPrincipal(
                 .build())
         .build();
   }
+
+  @Override
+  public ApiResponse<Void> updateLock(
```
abhisheknath2011, line 146:
```suggestion
  public ApiResponse<Void> toggleLock(
```

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/UpdateLockRequestBody.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+@Builder
+@EqualsAndHashCode
+@Getter
+@AllArgsConstructor(access = AccessLevel.PROTECTED)
+@NoArgsConstructor(access = AccessLevel.PROTECTED)
+public class UpdateLockRequestBody {
```
abhisheknath2011, line 19:
```suggestion
public class LockStateRequestBody {
```

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/UpdateLockRequestBody.java
```diff
@@ -0,0 +1,31 @@
...
+  @Schema(description = "locked state to be updated for table.")
+  @NotEmpty(message = "locked cannot be cannot be empty")
+  @NotNull(message = "locked State cannot be null")
+  boolean locked;
```
abhisheknath2011, line 23: Should this be isLocked?

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/TablesController.java
```diff
@@ -291,4 +292,39 @@ public ResponseEntity<GetAclPoliciesResponseBody> getAclPoliciesForUserPrincipal
...
+        @ApiResponse(responseCode = "403", description = "lock PATCH: FORBIDDEN"),
+        @ApiResponse(responseCode = "404", description = "lock PATCH: TABLE_NOT_FOUND")
+      })
+  @PatchMapping(
+      value = {
+        "/v0/databases/{databaseId}/tables/{tableId}/lock",
+        "/v1/databases/{databaseId}/tables/{tableId}/lock"
+      },
+      produces = {"application/json"})
+  public ResponseEntity<Void> updateLock(
```
abhisheknath2011, line 314:
```suggestion
  public ResponseEntity<Void> toggleLock(
```
Toggle or flip lock both works.

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/TablesController.java
```diff
@@ -291,4 +292,39 @@ public ResponseEntity<GetAclPoliciesResponseBody> getAclPoliciesForUserPrincipal
...
+      value = {
+        @ApiResponse(responseCode = "204", description = "lock PATCH: NO_CONTENT"),
+        @ApiResponse(responseCode = "400", description = "lock PATCH: BAD_REQUEST"),
+        @ApiResponse(responseCode = "401", description = "lock PATCH: UNAUTHORIZED"),
+        @ApiResponse(responseCode = "403", description = "lock PATCH: FORBIDDEN"),
+        @ApiResponse(responseCode = "404", description = "lock PATCH: TABLE_NOT_FOUND")
+      })
+  @PatchMapping(
+      value = {
+        "/v0/databases/{databaseId}/tables/{tableId}/lock",
```
abhisheknath2011, line 310: As this is new API shall we remove `V0` version?

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/TablesController.java
```diff
@@ -291,4 +292,39 @@ public ResponseEntity<GetAclPoliciesResponseBody> getAclPoliciesForUserPrincipal
     return new ResponseEntity<>(
         apiResponse.getResponseBody(), apiResponse.getHttpHeaders(), apiResponse.getHttpStatus());
   }
+
+  @Operation(
+      summary = "Update locked state for Table",
+      description = "Updates locked state for a table identified by databaseId and tableId",
```
abhisheknath2011, line 298: Description needs to be updated as per API name.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/TablesApiHandler.java
rohitkum2506: Agree with the idea here, but I think we should have common method in service layer to maintain consistency with MVC API layering. Since there are `update` and `delete` lock APIs, it would be good to follow same naming in controller and handler layers, we can sync it in service layer with common logic to toggle.

services/tables/src/main/java/com/linkedin/openhouse/tables/dto/mapper/iceberg/PoliciesSpecMapper.java
```diff
@@ -101,6 +102,10 @@ public Policies mapPolicies(Policies policies) {
               .replication(mapReplicationPolicies(policies.getReplication()))
               .build();
     }
+    if (policies != null && policies.getLockState() == null) {
+      LockState lockState = LockState.builder().isLocked(false).message("default").build();
```
abhisheknath2011, line 106: Shall we have generic default message instead of just default?

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/UpdateLockRequestBody.java
rohitkum2506: Following convention with other APIs and request body, I tried to keep request object name consistent with API name.
`Update Lock` API -> request object is `UpdateLockRequestBody`

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/UpdateLockRequestBody.java
rohitkum2506: Yes, will update

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/TablesApiHandler.java
abhisheknath2011: As for the new table the lock state is false by default, why not use single API to update/toggle the state. In that case  we don't need to have delete API. It would be convenient for the user to use just a single API.

rohitkum2506: Also, `toggleLock` implies that anytime its called, it should change the state of lock  `true <-> false`, which is not the case with the API.

services/tables/src/main/java/com/linkedin/openhouse/tables/dto/mapper/iceberg/PoliciesSpecMapper.java
rohitkum2506: Ys, updated to:
`Default table lock state: unlocked`

services/tables/src/main/java/com/linkedin/openhouse/tables/services/TablesServiceImpl.java
```diff
@@ -211,6 +233,52 @@ public List<AclPolicy> getAclPolicies(
...
+          && tableDto.getPolicies().getLockState().isLocked() != updateLockRequestBody.isLocked()) {
+        // should allow updating lock on a table with different reason
+        policiesToSave = tableDto.getPolicies().toBuilder().lockState(lockState).build();
+      } else {
+        policiesToSave = policiesToSave.toBuilder().lockState(lockState).build();
+      }
+    } else {
+      policiesToSave = Policies.builder().lockState(lockState).build();
+    }
+    TableDto tableDtoToSave =
```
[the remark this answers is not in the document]
rohitkum2506: Right, we can avoid saving table if lock state is same.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/TablesApiHandler.java
abhisheknath2011: Going by the implementation, it is not just the state, but the reasoning is can be updated although the state can be same. So I am fine with keeping the API as update which can potentially handle toggling lock and update reason. Hence, we don't need another API for deleting lock.

abhisheknath2011: Discussed offline with @rohitkum2506. It would be good to capture the lock duration as well as optional parameter. Alternatively, we can have default lock duration. Lock duration can be a good use case going forward so that we can have auto unlock feature. Keeping the lock duration also helps to analyze how long a table was in lock state with what reason. Based on this we can have two APIs to start with `createLock` and `deleteLock` as you mentioned there could be different set of users who can lock and unlock a table. So in future, `updateLock` API might come if we want to extend the lock duration. Adding these points so that you can think more to see how we can serve current set of use cases and future requirements as well. Also thinking if it is possible to cover all these usecases with just two APIs, if not mentioned APIs looks good.

rohitkum2506: `Hence, we don't need another API for deleting lock`
Agree that we update API can handle the delete process as well. 
Reason for coming up with delete API is for cleaner handling of  authorization policies for OH as well for easy comprehension of users. 
Also, following REST principals, it would be better to use create lock as `post API` to create a lock on table and `delete API` to unlock. 
Create should be idempotent.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/UpdateLockRequestBody.java
rohitkum2506: I went back to keep it `locked` as json serde drops the `is` part for boolean names and adds to confusion. It needs to be handled explicitly if we want to use `isLocked`. It's better to avoid it IMO.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/TablesApiHandler.java
rohitkum2506: Updated API to be `createLock`. 
Delete should be handled by `deleteLock` API.

By default table policies would not hold any state for Lock on a table. This translates table being in unlocked state by default.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockState.java
```diff
@@ -0,0 +1,36 @@
...
+  String message = "Default";
+
+  @Schema(
+      description = "lock creation time measured in UTC milliseconds for a table",
+      example = "1651002318265")
+  @Builder.Default
+  long creationTime = System.currentTimeMillis();
+
+  @Schema(
+      description = "lock expiration time measured in UTC milliseconds for a table",
```
abhisheknath2011, line 32: Update the description?

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/CreateUpdateLockRequestBody.java
```diff
@@ -0,0 +1,40 @@
...
+  String message;
+
+  @Schema(
+      description = "lock creation time measured in UTC milliseconds for a table",
+      example = "1651002318265")
+  @Builder.Default
+  long creationTime = System.currentTimeMillis();
+
+  @Schema(description = "lock expiration time measured in int `n` days", example = "3")
+  int expirationInDays = 0;
```
abhisheknath2011, line 35: I believe n days makes it easier for user to enter number of days as API input. Are we planning to store this as epoch time as a table property for easier backend processing? Also it would be in line with the creation time.

abhisheknath2011: Lets say in future we provide update API to extend expiration. How user input would be considered from API perspective, from creation time or extend by n days from expiration time. Lets say initial expiration day is 14 days and user wants to extends by 7 days. User can also enter 21 days. So the API can have two different interpretations. How about just accepting expiration date or just epoch time, which will always provide exact date, so no dual interpretation with respect to update. wdyt?

rohitkum2506: Fo all policies update process does not account for previous values and is a full override. I think update should also be just required to provide new `daysToExpire` from `creationTime`. 
It'll be confusing to use update as an extension on current expiration in days.

Also, expirationInDays as `int` is for easy interpretation.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockState.java
rohitkum2506: Done

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/CreateUpdateLockRequestBody.java
rohitkum2506: On storing in different format to match creation date and ease of processing, I think its better to keep policy as user inputs as it saves from serde to days format when user tries to `get` i, which could be erroneous as well.
Processing can be handled at consumer layer as suitable.

abhisheknath2011: Yes agree on keeping the same input format as stored format. But do think it would help if we take just epoch time or date as input from backend processing as well like no computation needed? If we consider expiration in days then the update API in future should just take days provided by user. So we might need to enforce some validation for update API if we go with days like updated day should always be greater then the current expiry day.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockState.java
```diff
@@ -0,0 +1,36 @@
...
+      description = "lock creation time measured in UTC milliseconds for a table",
+      example = "1651002318265")
+  @Builder.Default
+  long creationTime = System.currentTimeMillis();
+
+  @Schema(
+      description = "lock expiration time measured in days from creationTime for a table",
+      example = "3")
+  @Builder.Default
+  int expirationInDays = -1;
```
abhisheknath2011, line 35: Is -1 means no expiry set by default?

services/tables/src/main/java/com/linkedin/openhouse/tables/api/validator/impl/OpenHouseTablesApiValidator.java
```diff
@@ -249,6 +250,31 @@ public void validateGetAclPolicies(String databaseId, String tableId) {
...
+    List<String> validationFailures = new ArrayList<>();
+    validateGetTable(databaseId, tableId);
+    if (!createUpdateLockRequestBody.isLocked()) {
+      String errMsg =
+          String.format(
+              "Locked state on databaseId: %s tableId: %s should be set to true. Provided: false",
+              databaseId, tableId);
+      validationFailures.add(errMsg);
+    }
+    if (createUpdateLockRequestBody.getExpirationInDays() < 0) {
```
abhisheknath2011, line 265: Are we planning to mandate expiration time? If so we can have a default expiry (14 days) and let user override the value as part of create. We can keep the validation for negative expiration days. As we are enforcing expiry days what happens after expiry as we don't have a mechanism to unlock automatically (lets say by some offline jobs)? May be we will have to come up with offline jobs to auto unlock the tables. Other option is to start with non expiration of lock and keep this field as optional.

abhisheknath2011: But yeah keeping the tables locked for longer duration would cause more read and write issues. So auto unlock feature is something we might need sooner to avoid such issues.

rohitkum2506: Yes, current strategy is to keep expiration optional and not enforced at tables API layer. 
We don't have a use case right now to set a default values so avoiding default values.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockState.java
rohitkum2506: Yes, -1 indicates no value is set

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/CreateUpdateLockRequestBody.java
rohitkum2506: I think both ways are valid and equally relevant validation will need to be done with days as well as epoch time.
Major criteria is it should be > currentTime and lock creation time.

I don't have strong inclination to either, but advantage I see with int days format is ease of consumption from user side. Let me know if you strongly align with epoch time. I can update it.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/validator/impl/OpenHouseTablesApiValidator.java
abhisheknath2011: Sounds good. But will this validation is enforced (expirationInDays is defaulted to -1) for every create call, so making this mandatory?

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/CreateUpdateLockRequestBody.java
abhisheknath2011: Yeah lets go with expiration in days, we have discussed the pros and cons. It would be good from API usage perspective.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/validator/impl/OpenHouseTablesApiValidator.java
rohitkum2506: Yes, the validation is enforced for every createCall and is Done on `CreateLockStateRequest` where the default value is 0. 
-1 was set on `LockState` object which is in policies. I see the confusion, updated the default to 0 in lockState as well. Can't have default value as -1 since users can also set it that which will violate the expiration time semantics.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/CreateUpdateLockRequestBody.java
```diff
@@ -0,0 +1,40 @@
...
+import lombok.NoArgsConstructor;
+
+@Builder
+@EqualsAndHashCode
+@Getter
+@AllArgsConstructor(access = AccessLevel.PROTECTED)
+@NoArgsConstructor(access = AccessLevel.PROTECTED)
+public class CreateUpdateLockRequestBody {
+  @Schema(description = "locked state to be created/updated for table.")
+  @NotEmpty(message = "locked cannot be cannot be empty")
```
abhisheknath2011, line 21:
```suggestion
  @NotEmpty(message = "locked cannot be empty")
```

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockState.java
```diff
@@ -0,0 +1,36 @@
...
+  String message = "Default";
+
+  @Schema(
+      description = "lock creation time measured in UTC milliseconds for a table",
+      example = "1651002318265")
+  @Builder.Default
+  long creationTime = System.currentTimeMillis();
+
+  @Schema(
+      description = "lock expiration time measured in days from creationTime for a table",
```
abhisheknath2011, line 32:
```suggestion
      description = "lock expiration time for a table is `n` days from creationTime",
```

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockState.java
```diff
@@ -0,0 +1,36 @@
...
+  @Schema(description = "locked state of a table, example = false, true")
+  @Builder.Default
+  boolean locked = false;
+
+  @Schema(description = "message for locking", example = "")
+  @Builder.Default
+  String message = "Default";
+
+  @Schema(
+      description = "lock creation time measured in UTC milliseconds for a table",
```
abhisheknath2011, line 26:
```suggestion
      description = "lock creation time for a table is in UTC milliseconds",
```

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/CreateUpdateLockRequestBody.java
```diff
@@ -0,0 +1,40 @@
...
+  @NotNull(message = "locked value cannot be null")
+  boolean locked;
+
+  @Schema(description = "reason for creating/updating the lock on table")
+  String message;
+
+  @Schema(
+      description = "lock creation time measured in UTC milliseconds for a table",
```
abhisheknath2011, line 29:
```suggestion
      description = "lock creation time for a table is in UTC milliseconds",
```

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/CreateUpdateLockRequestBody.java
```diff
@@ -0,0 +1,40 @@
...
+      example = "1651002318265")
+  @Builder.Default
+  long creationTime = System.currentTimeMillis();
+
+  @Schema(description = "lock expiration time measured in int `n` days", example = "3")
```
abhisheknath2011, line 34:
```suggestion
  @Schema(description = "lock expiration time for a table is `n` days from creationTime", example = "3")
```

services/tables/src/main/java/com/linkedin/openhouse/tables/api/validator/impl/OpenHouseTablesApiValidator.java
```diff
@@ -249,6 +250,31 @@ public void validateGetAclPolicies(String databaseId, String tableId) {
...
+      String errMsg =
+          String.format(
+              "Lock creation time cannot be less that expiration time. Creation Time: %s, expiration time: %sdays",
```
abhisheknath2011, line 268:
```suggestion
              "Lock creation time cannot be less that expiration time. Creation Time: %s, expiration time: %s days",
```

services/tables/src/main/java/com/linkedin/openhouse/tables/services/TablesServiceImpl.java
```diff
@@ -145,6 +154,19 @@ public Pair<TableDto, Boolean> putTable(
     }
   }
 
+  private void checkIfLockPoliciesUpdated(
+      TableDto tableDto, CreateUpdateTableRequestBody requestBody) {
+    if (requestBody.getPolicies() != null
+        && requestBody.getPolicies().getLockState() != null
+        && requestBody.getPolicies().getLockState().isLocked()
+            != tableDto.getPolicies().getLockState().isLocked()) {
+      throw new IllegalArgumentException(
```
abhisheknath2011, line 163: I believe once we have update API then we will support updating lock state.

abhisheknath2011 (commented):
Thanks @rohitkum2506 for addressing the comments!  Looking great. Added some minor comments.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/TablesServiceImpl.java
rohitkum2506: Yes, Updating lock status should be part of updateAPI

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/CreateUpdateLockRequestBody.java
rohitkum2506: Done

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockState.java
rohitkum2506: Previous description was according to other timestamp related fields in GetTableResponse body. Better to keep consistency. Updated to match accordingly.
Same for `CreateUpdateLockReuqestBody`.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockState.java
rohitkum2506: Done

abhisheknath2011 (approved):
Thanks @rohitkum2506!

PR #304, opened 2025-04-01
Handle interrupted exception on doCommit call

rohitkum2506:
## Summary
Problem: Table commit operation exceptions are handled for various kinds of exceptions raised from service API calls at client. Since webFlux reactive Mono is blocked to wait for api to resolve, there are occasions when the `Mono.block` call can itself be interrupted. e.g: executing engine runs out of resource, Thread is terminated etc. This leads to unhandled exception at client side which results in partial commits. In some cases, this would result in table corruptions, when the metadata state has been updated in HTS but spark executors fail to write corresponding state to Filesystem successfully. Specifically these steps can occur:
1. Client makes commit call for a table
2. Server received the commit, performs DB update, table state update and responds with success
3. Client undergoes some error and fails to ack the success response(fails while waiting on Mono.block()), this error goes unhandled at TableOperations layer

For such cases, we need to handle the interrupted exception instances in commit operation and handle it such that transaction can be rolled back and table state is maintained.

## Changes

- [ ] Internal API Changes
- [x] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
@@ -133,6 +133,50 @@ public class OpenHouseTableOperationsTest {
   }
 
   @Test
+  public void testCreateUpdateTableInterruptedErrorHandle() throws InterruptedException {
+    TableIdentifier id = TableIdentifier.of("a", "b");
+    FileIO mockFileIO = mock(FileIO.class);
+    TableApi mockTableApi = mock(TableApi.class);
+    SnapshotApi mockSnapshotApi = mock(SnapshotApi.class);
+    OpenHouseTableOperationsForTest openHouseTableOperations =
+        new OpenHouseTableOperationsForTest(
+            id, mockFileIO, mockTableApi, mockSnapshotApi, "cluster");
+
+    TableMetadata metadata = mock(TableMetadata.class);
+    TableMetadata base = mock(TableMetadata.class);
+
+    // ensure the metadata-comparison triggers
+    Schema mockSchemaX = mock(Schema.class);
+    Schema mockSchemaY = mock(Schema.class);
+    when(metadata.schema()).thenReturn(mockSchemaX);
+    when(base.schema()).thenReturn(mockSchemaY);
+    Map<String, String> propsBase = ImmutableMap.of();
+    Map<String, String> propsMeta = ImmutableMap.of("a", "b");
+    when(metadata.properties()).thenReturn(propsMeta);
+    when(base.properties()).thenReturn(propsBase);
+
+    // ensure this is a snapshot change
+    List<Snapshot> snapshotList = Lists.newArrayList();
+    List<Snapshot> snapshotList1 = Lists.newArrayList();
+    when(metadata.snapshots()).thenReturn(snapshotList);
+    when(base.snapshots()).thenReturn(snapshotList1);
+
+    // Simulates a long-running operation
+    when(mockTableApi.updateTableV1(anyString(), anyString(), any())).thenReturn(Mono.never());
+
+    // Interrupt the current thread before calling .block()
+    // When .block() is called on response mono from snapshotApi, it detects that the thread is
+    // interrupted
+    // and throws a RuntimeException wrapping an InterruptedException
+    Thread.currentThread().interrupt();
+
+    Assertions.assertThrows(
+        CommitStateUnknownException.class, () -> openHouseTableOperations.doCommit(base, metadata));
+    // ensure that the thread interrupt is cleared for other tests
+    Assertions.assertFalse(Thread.interrupted());
+  }
+
+  @Test
   public void testNoPoliciesInMetadata() {
     TableMetadata metadata = mock(TableMetadata.class);
     when(metadata.properties()).thenReturn(Collections.emptyMap());
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
@@ -104,10 +104,23 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
   public void doCommit(TableMetadata base, TableMetadata metadata) {
     log.info("Calling doCommit for table: {}", tableName());
     boolean metadataUpdated = isMetadataUpdated(base, metadata);
-    if (areSnapshotsUpdated(base, metadata)) {
-      putSnapshots(base, metadata);
-    } else if (metadataUpdated) {
-      createUpdateTable(base, metadata);
+    try {
+      if (areSnapshotsUpdated(base, metadata)) {
+        putSnapshots(base, metadata);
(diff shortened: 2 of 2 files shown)


integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
```diff
@@ -132,6 +132,50 @@ public void testCreateUpdateTableErrorHandle() {
...
+    when(base.snapshots()).thenReturn(snapshotList1);
+
+    // Simulates a long-running operation
+    when(mockTableApi.updateTableV1(anyString(), anyString(), any())).thenReturn(Mono.never());
+
+    // Interrupt the current thread before calling .block()
+    // When .block() is called on response mono from snapshotApi, it detects that the thread is
+    // interrupted
+    // and throws a RuntimeException wrapping an InterruptedException
+    Thread.currentThread().interrupt();
```
cbb330, line 171: we want to simulate:
1. doCommit
2. mono block
3. interrupt

but it seems here that interrupt happens before doCommit. is this correct?

cbb330 (changes requested):

cbb330:
everything LGTM on first pass. nice! just one comment on the unit test.

also, adding some context in PR,

* here is a discussion around CSUE and why opting for conservative approach of NOT cleaning up files is preferred: https://github.com/apache/iceberg/pull/10373

* for future readers, the `GlueTableOperations`  implementation of `BaseMetastoreTableOperations` takes a similar approach to this change

integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
```diff
@@ -132,6 +132,50 @@ public void testCreateUpdateTableErrorHandle() {
...
+
+    Assertions.assertThrows(
+        CommitStateUnknownException.class, () -> openHouseTableOperations.doCommit(base, metadata));
+    // ensure that the thread interrupt is cleared for other tests
+    Assertions.assertFalse(Thread.interrupted());
```
cbb330, line 176: can we also validate that the files aren't cleaned up?

cbb330: actually, since this will be executed in a mock environment,

 its more useful to validate this integrity of files in integration docker testing

integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
rohitkum2506: Yes, it does not simulate the sequence rather ensures that Mono.block encounters interrupt when table API is executing or finished execution. 
When openHouseTableOperations.doCommit(base, metadata) is executed, it eventually calls .block() on the Mono returned by updateTableV1.
Since the thread is already interrupted, .block() immediately throws a RuntimeException wrapping an InterruptedException instead of waiting for the Mono to complete.

integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
rohitkum2506:
```
can we also validate that the files aren't cleaned up
``` 
I'm not too sure if the files are cleaned as part of failed OPS here. Have you verified that the error was because files were deleted or files were not created as per table metadata?

sumedhsakdeo:
Thanks for PR. It looks great. 

Are there other kinds of RuntimeException we need to handle?
Also look at exception handling in this Iceberg PR solving similar issue by Amogh 
https://github.com/apache/iceberg/pull/10373
See if we miss any cases.

integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
cbb330: only the compute engine e.g. spark will clean up. the error was because files were deleted. it may require something like a sleep on the catalog after metadata commit and before response , and a ctrl+c on the docker spark client.

sumedhsakdeo (approved):

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -104,10 +104,23 @@ public void doRefresh() {
...
-      putSnapshots(base, metadata);
-    } else if (metadataUpdated) {
-      createUpdateTable(base, metadata);
+    try {
+      if (areSnapshotsUpdated(base, metadata)) {
+        putSnapshots(base, metadata);
+      } else if (metadataUpdated) {
+        createUpdateTable(base, metadata);
+      }
+    } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
abhisheknath2011: But yeah we will treat non runtime exception as it is unless we want to handle any specific ones. If that is the case RuntimeException should also suffice. wdyt?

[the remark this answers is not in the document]
rohitkum2506: The catch block here will basically be handling Unchecked exceptions after the API call has processed. CheckedExceptions will anyway bubble to iceberg.

integrations/java/iceberg-1.2/openhouse-java-itest/src/test/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperationsTest.java
rohitkum2506: Putting sleep on catalog after metadata commit and before response is not in our control, since it happens as part of Iceberg OPS and tableAPI. We can add a timed sleep but that makes tests flaky. 

Ctrl+c on docker client does not simulate an interrupt exactly. It throws IOException which does not lead to any corruption.

cbb330: gotcha, in that case we already have enough confidence to ship and get deployment started 👍 

assuming that you have extra time to continue testing, perhaps InterruptedException can be triggered in the spark process by sending SIGINT signal to the PID after exec'ing into the docker spark docker container, e.g. `kill -SIGINT <java_process_pid>`

I agree we should not do it in unittest, but rather part of one time e2e testing.

rohitkum2506: Yes, I tried exactly that for local testing. Not sure if it's with the timing to kill the process but I keep getting IOException with that.

PR #319, opened 2025-05-07
Fix jobs scheduler to prevent burst of submission after timeout

jiang95-dev:
#### Bug description
There have been observations that the job scheduler submits all remaining waiting jobs after timeout. This creates bursts of job submissions to the engines which causes heavy pressure. The hypothesis is that after we cancel each tasks in the loop, one more task will be polled from the blocking queue and call the jobs service before it is cancelled. So we need to drain the queue before we start to cancel tasks.

#### ShutdownNow vs Queue
`shutdownNow` drains the queue, and in addition it interrupts all workers. However, from logging perspective, we cannot use `shutdownNow` because all active tasks will be marked as `Done` and thus we cannot distinguish between a finished job and an interrupted job. This will pollute the stats of cancelled jobs and succeeded jobs.

#### Next step
Currently tasks only handle interrupt during sleep. We should wrapped it around `call()` so that it will cancel the job whenever scheduler signals a kill.

Skipped count is still inaccurate in that tasks can return empty state for any unhandled failures. We can enrich the `JobState` or use `TaskState` for better metrics.

## Changes

- [ ] Internal API Changes
- [x] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

Need to test it in production.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
@@ -33,9 +33,9 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import lombok.extern.slf4j.Slf4j;
@@ -94,15 +94,15 @@ public class JobsScheduler {
   private static final String SUPPORTED_OPERATIONS_STRING =
       String.join(",", OPERATIONS_REGISTRY.keySet());
 
-  private final ExecutorService executorService;
+  private final ThreadPoolExecutor executors;
   private final OperationTaskFactory<? extends OperationTask> taskFactory;
   private final TablesClient tablesClient;
 
   public JobsScheduler(
-      ExecutorService executorService,
+      ThreadPoolExecutor executors,
       OperationTaskFactory<? extends OperationTask> taskFactory,
       TablesClient tablesClient) {
-    this.executorService = executorService;
+    this.executors = executors;
     this.taskFactory = taskFactory;
     this.tablesClient = tablesClient;
   }
@@ -124,11 +124,14 @@ public class JobsScheduler {
                 OperationTask.POLL_INTERVAL_MS_DEFAULT),
             NumberUtils.toLong(
                 cmdLine.getOptionValue("taskTimeoutMs"), OperationTask.TIMEOUT_MS_DEFAULT));
-    JobsScheduler app =
-        new JobsScheduler(
-            Executors.newFixedThreadPool(getNumParallelJobs(cmdLine)),
-            tasksFactory,
-            tablesClientFactory.create());
+    ThreadPoolExecutor executors =
+        new ThreadPoolExecutor(
+            getNumParallelJobs(cmdLine),
+            getNumParallelJobs(cmdLine),
+            0L,
+            TimeUnit.MILLISECONDS,
+            new LinkedBlockingQueue<Runnable>());
+    JobsScheduler app = new JobsScheduler(executors, tasksFactory, tablesClientFactory.create());
     app.run(
         operationType,
         operationTaskCls.toString(),
@@ -162,7 +165,7 @@ public class JobsScheduler {
     log.info("Submitting and running {} jobs based on the job type: {}", taskList.size(), jobType);
     List<Future<Optional<JobState>>> taskFutures = new ArrayList<>();
     for (OperationTask<?> operationTask : taskList) {
-      taskFutures.add(executorService.submit(operationTask));
+      taskFutures.add(executors.submit(operationTask));
     }
 
     int emptyStateJobCount = 0;
@@ -184,6 +187,7 @@ public class JobsScheduler {
       } catch (InterruptedException e) {
         throw new RuntimeException("Scheduler thread is interrupted, shutting down", e);
       } catch (TimeoutException e) {
+        executors.getQueue().clear();
         if (!taskFuture.isDone()) {
           log.warn(
               "Attempting to cancel job for {} because of timeout of {} hours",
@@ -211,7 +215,7 @@ public class JobsScheduler {
         jobStateCountMap.get(JobState.CANCELLED),
         jobStateCountMap.get(JobState.FAILED),
         emptyStateJobCount);
-    executorService.shutdown();
+    executors.shutdown();
     METER.counterBuilder("scheduler_end_count").build().add(1);
     reportMetrics(jobStateCountMap, taskType, startTimeMillis);
   }
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
@@ -176,10 +176,6 @@ public class JobsScheduler {
       try {
         long passedTimeMillis = System.currentTimeMillis() - startTimeMillis;
         long remainingTimeMillis = TimeUnit.HOURS.toMillis(tasksWaitHours) - passedTimeMillis;
-        if (remainingTimeMillis <= 0) {
-          // treat as a global timeout case similar to future.get timeout
-          throw new TimeoutException();
-        }
         jobState = taskFuture.get(remainingTimeMillis, TimeUnit.MILLISECONDS);
       } catch (ExecutionException e) {
         log.error(String.format("Operation for %s failed with exception", task), e);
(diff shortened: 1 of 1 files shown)


apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
```diff
@@ -184,6 +187,7 @@ protected void run(
       } catch (InterruptedException e) {
         throw new RuntimeException("Scheduler thread is interrupted, shutting down", e);
       } catch (TimeoutException e) {
+        executors.getQueue().clear();
```
cbb330, line 190: add log.info(queue.size is being cleared)

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
```diff
@@ -211,7 +215,7 @@ protected void run(
         jobStateCountMap.get(JobState.CANCELLED),
         jobStateCountMap.get(JobState.FAILED),
         emptyStateJobCount);
-    executorService.shutdown();
+    executors.shutdown();
```
cbb330, line 218: add log.info(executors shutting down) and any attributes of the executors that is worth printing e.g. current threads

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
```diff
@@ -124,11 +124,14 @@ public static void main(String[] args) {
...
-            Executors.newFixedThreadPool(getNumParallelJobs(cmdLine)),
-            tasksFactory,
-            tablesClientFactory.create());
+    ThreadPoolExecutor executors =
+        new ThreadPoolExecutor(
+            getNumParallelJobs(cmdLine),
+            getNumParallelJobs(cmdLine),
+            0L,
+            TimeUnit.MILLISECONDS,
+            new LinkedBlockingQueue<Runnable>());
```
cbb330, line 133: is this FIFO? 

what if head() runs for 6 hours?

cbb330 (changes requested):

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
teamurko: would executorService.shutdownNow() achieve the same?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
jiang95-dev: It is FIFO. Head will be picked by one thread. It won't block other threads.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
abhisheknath2011: +1 can we check if shutdownNow works here?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
```diff
@@ -124,11 +124,14 @@ public static void main(String[] args) {
...
-    JobsScheduler app =
-        new JobsScheduler(
-            Executors.newFixedThreadPool(getNumParallelJobs(cmdLine)),
-            tasksFactory,
-            tablesClientFactory.create());
+    ThreadPoolExecutor executors =
+        new ThreadPoolExecutor(
+            getNumParallelJobs(cmdLine),
+            getNumParallelJobs(cmdLine),
+            0L,
```
abhisheknath2011, line 131: I think we should have some higher value for keep live time (60 sec or more) as it determines how long a thread can be idle before terminating.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
abhisheknath2011, line 218: Shutdown now drains the queue.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
abhisheknath2011: We can adjust it to higher value when we switch to batch discovery and submission. For now 0L should be good as default ExecutorSevice default implementation with fixedThreadPool internally uses 0L and creates ThreadPoolExecutor. If shutdownNow on executorService works then we probably don't need to explicitly create this.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
jiang95-dev: It works but we need to change the code a bit. shutdownNow() would mark all running futures as done, so it won't go into the if block.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
jiang95-dev: `keepAliveTime` doesn't matter when `core threads` == `max threads`.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
jiang95-dev: However. `showdown() `should work since we are explicitly cancelling the running tasks.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
jiang95-dev: I changed the implementation, so no need to print the queue info.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
```diff
@@ -184,6 +184,7 @@ protected void run(
       } catch (InterruptedException e) {
         throw new RuntimeException("Scheduler thread is interrupted, shutting down", e);
       } catch (TimeoutException e) {
+        executorService.shutdown();
```
abhisheknath2011, line 187: Looked at ThreadPoolExecutor code. Did not see that [executor.shutdown](https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/concurrent/ThreadPoolExecutor.java#L1351) is draining the queue. Instead [shutdownNow](https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/concurrent/ThreadPoolExecutor.java#L1390) drains the queue and returns the queued tasks. Then we can continue on the outer loop and cancel on the future tasks. Essentially shutdownNow should interrupt the running threads.

abhisheknath2011: We are already handling interrupted exception in the OperationTask thread itself.

abhisheknath2011: With shutdown option we might have to provide some grace period for cleaner shutdown and need to use waitForTermination with timeout before actually calling shutdownNow. Lets say our job SLA is 12 hour and we provide grace period of another 30 mins for the existing jobs to complete. This can be another option. But we can follow strict SLA of of 12 hours with shutdownNow.

abhisheknath2011: Other option is increasing the job run SLA itself so that job submission percentage increases and then shutdown will help terminate as per SLA.

teamurko: we need shutdownNow() to cancel queued tasks

teamurko (approved):

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
```diff
@@ -173,31 +176,33 @@ protected void run(
       try {
         long passedTimeMillis = System.currentTimeMillis() - startTimeMillis;
         long remainingTimeMillis = TimeUnit.HOURS.toMillis(tasksWaitHours) - passedTimeMillis;
-        if (remainingTimeMillis <= 0) {
-          // treat as a global timeout case similar to future.get timeout
-          throw new TimeoutException();
-        }
         jobState = taskFuture.get(remainingTimeMillis, TimeUnit.MILLISECONDS);
```
abhisheknath2011, line 179: Does taskFuture.get with negative value returns TimeoutException or other exception? Better to verify if it waits longer. Usually positive value is provided.

jiang95-dev: Given negative value, it will return jobState if it is exists, otherwise TimeoutException.

abhisheknath2011: Ok nice.

abhisheknath2011 (approved):
Thanks @jiang95-dev exploring on clean shutdown! As a next step we could consider adding pre SLA check (may be 15 - 30 mins before SLA) to clear the queued jobs so that there is no new submission and get status cleanly for the running jobs (already submitted to Setu) before shutting down completely.

cbb330 (approved):
LGTM and thanks for adding meaningful logs

PR #324, opened 2025-05-19
OpenHouse server accept client provided schema as-it-is

autumnust:
## Summary

Supporting evolution pattern like Schema's Field Ordering for OpenHouse. 
The data flow for the schema evolution is like below: 

1. User-side, which is mostly engine's catalog layer, provided a evolved schema,
2. Such schema gets serialized and passes over the wire. There's a specific segment `schema` in OH's REST API to capture that "newSchema", if there's evolution happening on the client side, 
3. OpenHouse server simply does a `unionByNameWith` operation using Iceberg's Table API. 

Note that this is also how `merge-schema` works in Spark's client side, see https://github.com/apache/iceberg/blob/31e31fd819c846f49d2bd459b8bfadfdc3c2bc3a/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkWriteBuilder.java#L172 as a reference. 

On a high level, OpenHouse's internalCatalog simulate the client-side behavior of Iceberg's catalog and "replay" the operation using the high-level Iceberg's Table API, like `UpdateProperties`, `table.updateSchema()` etc. This approach doesn't work well in support column reordering, as OH server side has no direct access to the information about how reordering is conducted on the client side, which is necessary if sticking to the higher level API like `moveBefore` and `moveAfter` exposed by `table.updateSchema()`. 

The new approach here is to
- Trust the `newSchema` provided by client side to the be source-of-truth after simple validation (e.g. Field-id not drifted, no columns being dropped, this behavior can be extended based on specific needs), 
- Use `newSchema` to assemble the new `TableMetadata` object to be persisted as part of commit on the server side, 
- To do so, we use a similar method like snapshot-json, to let service-level constructs `services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java` passing `newSchema` through table properties, and for catalog-level constructs `iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java` deserialize it and process. 

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

Adding a new test cases to demonstrate how the client-side resolved schema will be persisted on the server side. 
Other wise, passing all existing schema evolution test cases since the core logic is being changed on the server side as well. 

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/apps/spark/src/test/java/com/linkedin/openhouse/catalog/e2e/SparkSchemaEvolutionTest.java
+++ b/apps/spark/src/test/java/com/linkedin/openhouse/catalog/e2e/SparkSchemaEvolutionTest.java
@@ -1,0 +1,86 @@
+package com.linkedin.openhouse.catalog.e2e;
+
+import com.linkedin.openhouse.jobs.spark.Operations;
+import com.linkedin.openhouse.tablestest.OpenHouseSparkITest;
+import java.util.Arrays;
+import java.util.List;
+import org.apache.iceberg.Schema;
+import org.apache.iceberg.Table;
+import org.apache.iceberg.spark.SparkSchemaUtil;
+import org.apache.iceberg.types.Types;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.RowFactory;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.Metadata;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class SparkSchemaEvolutionTest extends OpenHouseSparkITest {
+
+  @Test
+  void testEndUserSchemaEvolution() throws Exception {
+    SparkSession spark = null;
+    try {
+      spark = getSparkSession();
+      spark.sql("CREATE TABLE openhouse.d1.t1 (name string, id bigint)");
+      spark.sql("INSERT INTO openhouse.d1.t1 VALUES ('Alice', 1)");
+      spark.sql("INSERT INTO openhouse.d1.t1 VALUES ('Bob', 2), ('Charlie', 3)");
+
+      // Inspect the schema object first, establish the baseline
+      Dataset<Row> tableDF = spark.table("openhouse.d1.t1");
+
+      StructType oldSchema = tableDF.schema();
+
+      // Create new dummy dataframe with reordered columns
+      StructType schema =
+          new StructType(
+              new StructField[] {
+                new StructField("dt", DataTypes.StringType, false, Metadata.empty()),
+                new StructField("id", DataTypes.LongType, false, Metadata.empty()),
+                new StructField("name", DataTypes.StringType, false, Metadata.empty())
+              });
+
+      List<Row> data =
+          Arrays.asList(
+              RowFactory.create("2025-04-05", 1L, "John Doe"),
+              RowFactory.create("2025-04-06", 2L, "Jane Smith"));
+
+      // Remember the column ordering here is dt, id, name,
+      // the idea table schema's column ordering will be name, id, dt (as long as dt is at the end,
+      // it should be working)
+      Dataset<Row> dummyDF = spark.createDataFrame(data, schema);
+
+      // Manually evolve to mimic client-side explicit evolution, before a write can be issued
+      Operations operations = Operations.withCatalog(spark, null);
+      Table table = operations.getTable("d1.t1");
+      int colLen = table.schema().columns().size();
+      String lastColName = table.schema().columns().get(colLen - 1).name();
+      table
+          .updateSchema()
+          .unionByNameWith(SparkSchemaUtil.convert(schema))
+          .moveAfter("dt", lastColName)
+          .commit();
+
+      Schema newSchema = operations.getTable("d1.t1").schema();
+      List<Types.NestedField> newSchemaCols = newSchema.columns();
+      Assertions.assertEquals(newSchemaCols.size(), 3);
+      Assertions.assertEquals(newSchemaCols.get(2).name(), "dt");
+
+      // This is necessary to ensure Spark not caching the previous state.
+      spark.sql("REFRESH TABLE openhouse.d1.t1");
+
+      dummyDF.write().mode("append").format("parquet").saveAsTable("openhouse.d1.t1");
+
+      tableDF = spark.table("openhouse.d1.t1");
+      Assertions.assertEquals(tableDF.count(), 5);
+    } finally {
+      if (spark != null) {
+        spark.sql("DROP TABLE openhouse.d1.t1");
+      }
+    }
+  }
+}
--- a/apps/spark/src/test/java/com/linkedin/openhouse/catalog/e2e/SparkSchemaEvolutionTest.java
+++ b/apps/spark/src/test/java/com/linkedin/openhouse/catalog/e2e/SparkSchemaEvolutionTest.java
@@ -75,8 +75,24 @@ public class SparkSchemaEvolutionTest extends OpenHouseSparkITest {
 
       dummyDF.write().mode("append").format("parquet").saveAsTable("openhouse.d1.t1");
 
+      // Validating write go through successfully
       tableDF = spark.table("openhouse.d1.t1");
       Assertions.assertEquals(tableDF.count(), 5);
+
+      // another schema evolution
+      StructType schema2 =
+          new StructType(
+              new StructField[] {
+                new StructField("dt", DataTypes.StringType, false, Metadata.empty()),
+                new StructField("zipcode", DataTypes.StringType, false, Metadata.empty()),
+                new StructField("id", DataTypes.LongType, false, Metadata.empty()),
+                new StructField("name", DataTypes.StringType, false, Metadata.empty())
+              });
+      table = operations.getTable("d1.t1");
+      table.updateSchema().unionByNameWith(SparkSchemaUtil.convert(schema2)).commit();
+      newSchema = operations.getTable("d1.t1").schema();
+      newSchemaCols = newSchema.columns();
+      Assertions.assertEquals(newSchemaCols.size(), 4);
     } finally {
       if (spark != null) {
         spark.sql("DROP TABLE openhouse.d1.t1");
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
@@ -15,7 +15,7 @@ public final class CatalogConstants {
   /** Used to uniquely identify an update towards a table from user side. */
   public static final String COMMIT_KEY = "commitKey";
 
-  public static final String EVOLVED_SCHEMA_KEY = "evolvedSchema";
+  public static final String EVOLVED_SCHEMA_KEY = "evolved.table.schema";
 
   static final String FEATURE_TOGGLE_STOP_CREATE = "stop_create";
 
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
@@ -15,6 +15,8 @@ public final class CatalogConstants {
   /** Used to uniquely identify an update towards a table from user side. */
   public static final String COMMIT_KEY = "commitKey";
 
+  public static final String EVOLVED_SCHEMA_KEY = "EvolvedSchema";
+
   static final String FEATURE_TOGGLE_STOP_CREATE = "stop_create";
 
   static final String CLIENT_TABLE_SCHEMA = "client.table.schema";
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
@@ -15,7 +15,7 @@ public final class CatalogConstants {
   /** Used to uniquely identify an update towards a table from user side. */
   public static final String COMMIT_KEY = "commitKey";
 
-  public static final String EVOLVED_SCHEMA_KEY = "EvolvedSchema";
+  public static final String EVOLVED_SCHEMA_KEY = "evolvedSchema";
 
   static final String FEATURE_TOGGLE_STOP_CREATE = "stop_create";
 
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
@@ -175,14 +175,23 @@ public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperatio
   }
 
   /** An internal helper method to rebuild the {@link TableMetadata} object. */
-  private TableMetadata rebuildTblMetaWithSchema(TableMetadata newMetadata, String schemaKey) {
+  private TableMetadata rebuildTblMetaWithSchema(
+      TableMetadata newMetadata, String schemaKey, boolean reuseMetadata) {
     Schema writerSchema = SchemaParser.fromJson(newMetadata.properties().get(schemaKey));
-    return TableMetadata.buildFrom(newMetadata)
-        .setCurrentSchema(writerSchema, writerSchema.highestFieldId())
-        .addPartitionSpec(
-            rebuildPartitionSpec(newMetadata.spec(), newMetadata.schema(), writerSchema))
-        .addSortOrder(rebuildSortOrder(newMetadata.sortOrder(), writerSchema))
-        .build();
+    if (reuseMetadata) {
+      return TableMetadata.buildFrom(newMetadata)
+          .setCurrentSchema(writerSchema, writerSchema.highestFieldId())
+          .build();
+    } else {
+      return TableMetadata.buildFromEmpty()
+          .setLocation(newMetadata.location())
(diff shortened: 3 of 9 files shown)


services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/TablesControllerTest.java
```diff
@@ -215,45 +215,49 @@ public void testCrudTables() throws Exception {
...
-        mvc, storageManager, evolveDummySchema(mvcResultT1d1), tableLocation);
-    RequestAndValidateHelper.updateTableAndValidateResponse(
-        mvc, storageManager, evolveDummySchema(mvcResultT2d1), tableSameDbLocation);
-    RequestAndValidateHelper.updateTableAndValidateResponse(
-        mvc, storageManager, evolveDummySchema(mvcResultT1d2), tableDiffDbLocation);
-
-    RequestAndValidateHelper.deleteTableAndValidateResponse(mvc, GET_TABLE_RESPONSE_BODY);
-    RequestAndValidateHelper.deleteTableAndValidateResponse(mvc, GET_TABLE_RESPONSE_BODY_SAME_DB);
-    RequestAndValidateHelper.deleteTableAndValidateResponse(mvc, GET_TABLE_RESPONSE_BODY_DIFF_DB);
+    try {
```
autumnust, line 218: There are no changes other than a finally block to ensure no side effects left even there are testing/execution failures in the code below

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -184,17 +196,11 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
-              .setCurrentSchema(clientSchema, metadata.lastColumnId())
-              .addPartitionSpec(
-                  rebuildPartitionSpec(metadata.spec(), metadata.schema(), clientSchema))
-              .addSortOrder(rebuildSortOrder(metadata.sortOrder(), clientSchema))
-              .setProperties(metadata.properties())
-              .build();
+      metadata = rebuildTblMetaWithSchema(metadata, CatalogConstants.CLIENT_TABLE_SCHEMA);
+    }
+
+    if (metadata.properties().get(CatalogConstants.EVOLVED_SCHEMA_KEY) != null) {
```
maluchari, line 202: For my understanding: why do we rebuild this twice when line 198 is true?

apps/spark/src/test/java/com/linkedin/openhouse/catalog/e2e/SparkSchemaEvolutionTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.jupiter.api.Test;
+
+public class SparkSchemaEvolutionTest extends OpenHouseSparkITest {
+
+  @Test
+  void testEndUserSchemaEvolution() throws Exception {
+    SparkSession spark = null;
+    try {
+      spark = getSparkSession();
+      spark.sql("CREATE TABLE openhouse.d1.t1 (name string, id bigint)");
```
sumedhsakdeo, line 29: maybe you want add a test-case that starts with name, id, dt as the schema, and then wants to evolve to name, id, desc, dt?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
```diff
@@ -15,6 +15,8 @@ public final class CatalogConstants {
   /** Used to uniquely identify an update towards a table from user side. */
   public static final String COMMIT_KEY = "commitKey";
 
+  public static final String EVOLVED_SCHEMA_KEY = "evolvedSchema";
```
sumedhsakdeo, line 18: Not for this PR, but do you think we eventually need to bring the evolved schema to top level spec, calling it writerSchema?

sumedhsakdeo (commented):
Thanks @autumnust , overall the PR looks great. I left some minor comments. Ptal.

apps/spark/src/test/java/com/linkedin/openhouse/catalog/e2e/SparkSchemaEvolutionTest.java
autumnust: It's technically the same as the current case, where `schema` at line 39 is intentionally putting `dt` in the very beginning. Client needs to do things at Line 62 - 66 to make sure it presents the expected schema where `dt` will be placed at the end.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
autumnust: with this change, the `schema` in the top-level API semantically means `writeSchema`. It's "take this schema as-it-is after this update"

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -174,6 +174,18 @@ public void commit(TableMetadata base, TableMetadata metadata) {
...
 
+  /** An internal helper method to rebuild the {@link TableMetadata} object. */
+  private TableMetadata rebuildTblMetaWithSchema(TableMetadata newMetadata, String schemaKey) {
+    Schema writerSchema = SchemaParser.fromJson(newMetadata.properties().get(schemaKey));
+    return TableMetadata.buildFrom(newMetadata)
+        .setCurrentSchema(writerSchema, writerSchema.highestFieldId())
+        .addPartitionSpec(
+            rebuildPartitionSpec(newMetadata.spec(), newMetadata.schema(), writerSchema))
+        .addSortOrder(rebuildSortOrder(newMetadata.sortOrder(), writerSchema))
+        .setProperties(newMetadata.properties())
```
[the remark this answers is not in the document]
autumnust: good catch. yes it's redundant, will remove.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BaseIcebergSchemaValidator.java
```diff
@@ -16,32 +16,63 @@
...
-  public void validateWriteSchema(
-      Schema oldSchema, Schema newSchema, Schema updatedSchema, String tableUri)
+  public void validateWriteSchema(Schema oldSchema, Schema newSchema, String tableUri)
       throws InvalidSchemaEvolutionException, IllegalArgumentException {
-    // TODO: Detect and resolve rename and reorder
-    // This implementation assumes both old and new schemas have field ids internally assigned by
-    // iceberg rather than specified by the users
+    validateFields(oldSchema, newSchema, tableUri);
     enforceDeleteColumnFailure(oldSchema, newSchema, tableUri);
-    TypeUtil.validateWriteSchema(
```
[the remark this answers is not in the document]
autumnust: I actually get stuck on this one for a while due to the existing unit tests. The previous usage for `TypeUtil.validateWriteSchema` is actually a bit problematic. 

- This method is designed to be called by users / catalog client. 
- There are in total 3 schemas accessible by catalog client: 1. old schema, 2. attempted write-schema, 3. merged write schema. 
- This method is validating between 2 and 3. 
All these can be demonstrated by the usage of `TypeUtil.validateWriteSchema` in https://github.com/apache/iceberg/blob/31e31fd819c846f49d2bd459b8bfadfdc3c2bc3a/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkWriteBuilder.java#L172 


In OpenHouse's case, we only have 1 & 3, so it doesn't really make sense to run validation and it will actually fail some of the valid existing test cases. The correct thing here is to run this validation on client/engine side.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -220,6 +226,9 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
               getCanonicalFieldName("tableLocation"), CatalogConstants.INITIAL_VERSION));
       properties.put(getCanonicalFieldName("tableLocation"), newMetadataLocation);
 
+      if (properties.containsKey(CatalogConstants.EVOLVED_SCHEMA_KEY)) {
```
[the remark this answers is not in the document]
autumnust: or we can print this in the log for debuggability ?
there are lineage of schema preserved in the `metadata.json`

apps/spark/src/test/java/com/linkedin/openhouse/catalog/e2e/SparkSchemaEvolutionTest.java
sumedhsakdeo: Confused, the client wants the schema to be dt, id, name, but the test is written in a way that it becomes id, name, and dt?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
sumedhsakdeo: I see. then evolved schema key is more for messaging passing in internal table operations.

sumedhsakdeo, line 18: nit: evolved.table.schema

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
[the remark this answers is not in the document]
sumedhsakdeo: true

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BaseIcebergSchemaValidator.java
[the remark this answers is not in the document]
sumedhsakdeo: But we can't guarantee that client's would do this validation. For example, a non-engine client trying to update the schema using Iceberg APIs. We have some tools doing that. We would need to catch it on the server side that they are not trying to make an optional field, as required.

Which line of code in the current PR would help catch this issue?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
```diff
@@ -15,6 +15,8 @@ public final class CatalogConstants {
...
+
   static final String FEATURE_TOGGLE_STOP_CREATE = "stop_create";
 
   static final String CLIENT_TABLE_SCHEMA = "client.table.schema";
```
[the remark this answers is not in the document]
sumedhsakdeo: ok sounds good. just a nit. `evolved.table.schema`

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/RepositoryTest.java
```diff
@@ -501,14 +501,30 @@ void testSchemaEvolutionBasic() {
...
-            validator.validateWriteSchema(
-                oldSchema, newSchema, update.apply(), createDto.getTableUri()));
+        () -> validator.validateWriteSchema(oldSchema, newSchema, createDto.getTableUri()));
     TableDtoPrimaryKey primaryKey = getPrimaryKey(TABLE_DTO);
     openHouseInternalRepository.deleteById(primaryKey);
     Assertions.assertFalse(openHouseInternalRepository.existsById(primaryKey));
   }
 
+  @Test
+  void testSchemaEvolutionWithMismatchedFieldId() {
```
aastha25, line 511: can you also add another test which has 2 subsequent evolutions. That will help validate that the evolved schema from 1st evolution is not corrupted.

apps/spark/src/test/java/com/linkedin/openhouse/catalog/e2e/SparkSchemaEvolutionTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.spark.sql.types.Metadata;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class SparkSchemaEvolutionTest extends OpenHouseSparkITest {
+
+  @Test
+  void testEndUserSchemaEvolution() throws Exception {
```
aastha25, line 25: can you also add another test which has 2 subsequent evolutions. That will help validate that the evolved schema from 1st evolution is not corrupted.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BaseIcebergSchemaValidator.java
```diff
@@ -16,32 +16,63 @@
...
+   *
+   * <p>Note that Iceberg by itself is case-sensitive by default, the casing info can be lost
+   * through the SQL engine layer. Here OH is honoring the default case sensitivity for column
+   * name-based resolution.
+   *
+   * @param oldSchema existing schema in table catalog
+   * @param newSchema user-provided schema that may contain evolution
+   * @param tableUri tableUri
+   */
+  protected void validateFields(Schema oldSchema, Schema newSchema, String tableUri) {
```
aastha25, line 56: this seems to be checking all fields from oldSchema are present in newSchema with same casing & fieldids. 
but do we also need (1) nullability matching between the fields of the two schemas? (2) do we enforce somewhere that the new top level fields added cannot be non-nullable.

sumedhsakdeo (changes requested):
Will wait for extra validations we want to add validateFields

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BaseIcebergSchemaValidator.java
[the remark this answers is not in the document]
autumnust: I had discussed with Raymond, fixing some of the unit test issues. Now the `validateWriteSchema` is preserved.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BaseIcebergSchemaValidator.java
autumnust: I am adding back the Iceberg validation method. That should cover the nullability case you mentioned.

apps/spark/src/test/java/com/linkedin/openhouse/catalog/e2e/SparkSchemaEvolutionTest.java
autumnust: I added a subsequent step to do one more evolution, although I am not sure if that's very useful, since the purpose for the change is to let OH server accept schema "as-it-is" instead of honoring some of the contracts, which I believe is the one you wanted to validate with by "That will help validate that the evolved schema from 1st evolution is not corrupted."

sumedhsakdeo (dismissed):
💯

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -174,6 +174,18 @@ public void commit(TableMetadata base, TableMetadata metadata) {
     }
   }
 
+  /** An internal helper method to rebuild the {@link TableMetadata} object. */
+  private TableMetadata rebuildTblMetaWithSchema(TableMetadata newMetadata, String schemaKey) {
+    Schema writerSchema = SchemaParser.fromJson(newMetadata.properties().get(schemaKey));
+    return TableMetadata.buildFrom(newMetadata)
+        .setCurrentSchema(writerSchema, writerSchema.highestFieldId())
```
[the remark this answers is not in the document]
Will-Lo: No this wouldn't, this specific function call takes the following signature and has this impl:
```
    public Builder setCurrentSchema(Schema newSchema, int newLastColumnId) {
      setCurrentSchema(addSchemaInternal(newSchema, newLastColumnId));
      return this;
    }
```
where the return value of addSchemaInternal adds to its own internal schema ID, afaik we need to incrementally evolve the schema ID of the replica table. It's been fixed for the setup job but I still need to fix it for multi schema evolutions in a single commit.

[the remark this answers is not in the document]
Will-Lo: I think it will fix the schema ID not being found for replication (given that the writerSchema can be found from the src Metadata, otherwise it will over increment the schema ID and cause a different problem). I'm wondering if there are any other side effects if the metadata is built directly from `newMetadata` instead of fromEmpty, which was the old code, took a look at the TableMetadata internals and I don't believe so

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -174,6 +174,26 @@ public void commit(TableMetadata base, TableMetadata metadata) {
     }
   }
 
+  /** An internal helper method to rebuild the {@link TableMetadata} object. */
+  private TableMetadata rebuildTblMetaWithSchema(
+      TableMetadata newMetadata, String schemaKey, boolean reuseMetadata) {
+    Schema writerSchema = SchemaParser.fromJson(newMetadata.properties().get(schemaKey));
+    if (reuseMetadata) {
```
autumnust, line 181: placeholder

sumedhsakdeo (approved):

PR #334, opened 2025-06-11
Add timeout for queued tasks

jiang95-dev:
## Summary
In the past, we kept the queued tasks for 3 hours. This PR reduced it to 10 minutes, and renames variables to prevent confusions. The reason behind the change is that, we found most of the queued tasks are not due to engine-side queuing, but due to an issue we called "orphan tasks". The root cause is that user apps are never launched because of out-of-quota issue in the user space, and so the state will never be updated resulting in thread spinning for 3 hours. This type of launch failure ends in minutes, but we want to make the killing mechanism more generic, so I'm setting the timeout to be 10 minutes. For running jobs, we will keep them running even after they hit the SLA of the scheduler, but only let Task exit status polling. This will prevent the waste of parallelism and also increase the job success rate. For more info, check this [doc](https://docs.google.com/document/d/1uJWIfYchhTNquFFrO9XVXr9CDOZJjIgvD2b3isJOZDI/edit?tab=t.0#heading=h.ki3l4jeu6j04).

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [x] Refactoring
- [ ] Documentation
- [ ] Tests

1. Stop cancelling queued or running jobs. Instead, only let Task exit status polling. Note, even the scheduler will only cancel the task instead of the submitted jobs after timeout. That being said, the SOT of `jobState` will only be mysql. The log and metrics reported by the scheduler only reflects the snapshot when it is killed.
2. Added a new parameter `queuedTimeoutMs`, so need to update the constructors for all Tasks.

## Testing Done

Will verify the results in prod.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
@@ -159,7 +159,7 @@ public class JobsScheduler {
                 cmdLine.getOptionValue("taskPollIntervalMs"),
                 OperationTask.POLL_INTERVAL_MS_DEFAULT),
             NumberUtils.toLong(
-                cmdLine.getOptionValue("taskTimeoutMs"), OperationTask.TIMEOUT_MS_DEFAULT));
+                cmdLine.getOptionValue("taskTimeoutMs"), OperationTask.QUEUED_TIMEOUT_MS_DEFAULT));
     ThreadPoolExecutor jobExecutors = null;
     ThreadPoolExecutor statusExecutors = null;
     if (isMultiOperationMode(cmdLine)) {
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
@@ -159,7 +159,8 @@ public class JobsScheduler {
                 cmdLine.getOptionValue("taskPollIntervalMs"),
                 OperationTask.POLL_INTERVAL_MS_DEFAULT),
             NumberUtils.toLong(
-                cmdLine.getOptionValue("taskTimeoutMs"), OperationTask.QUEUED_TIMEOUT_MS_DEFAULT));
+                cmdLine.getOptionValue("taskQueuedTimeoutMs"),
+                OperationTask.QUEUED_TIMEOUT_MS_DEFAULT));
     ThreadPoolExecutor jobExecutors = null;
     ThreadPoolExecutor statusExecutors = null;
     if (isMultiOperationMode(cmdLine)) {
@@ -491,7 +492,7 @@ public class JobsScheduler {
         Option.builder(null)
             .required(false)
             .hasArg()
-            .longOpt("taskTimeoutMs")
+            .longOpt("taskQueuedTimeoutMs")
             .desc("Timeout in milliseconds for an individual task")
             .build());
     // TODO: move these to ODD specific config
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
@@ -160,7 +160,9 @@ public class JobsScheduler {
                 OperationTask.POLL_INTERVAL_MS_DEFAULT),
             NumberUtils.toLong(
                 cmdLine.getOptionValue("taskQueuedTimeoutMs"),
-                OperationTask.QUEUED_TIMEOUT_MS_DEFAULT));
+                OperationTask.QUEUED_TIMEOUT_MS_DEFAULT),
+            NumberUtils.toLong(
+                cmdLine.getOptionValue("taskTimeoutMs"), OperationTask.TASK_TIMEOUT_MS_DEFAULT));
     ThreadPoolExecutor jobExecutors = null;
     ThreadPoolExecutor statusExecutors = null;
     if (isMultiOperationMode(cmdLine)) {
@@ -357,11 +359,13 @@ public class JobsScheduler {
     }
 
     log.info(
-        "Finishing scheduler for job type {}, tasks stats: {} created, {} succeeded,"
-            + " {} cancelled (timeout), {} failed, {} skipped (no state)",
+        "Finishing scheduler for job type {}, tasks stats: {} created, {} succeeded, {} running, {} queued"
+            + " {} cancelled (scheduler timeout), {} failed, {} skipped (no state)",
         jobType,
         taskList.size(),
         jobStateCountMap.get(JobState.SUCCEEDED),
+        jobStateCountMap.get(JobState.RUNNING),
+        jobStateCountMap.get(JobState.QUEUED),
         jobStateCountMap.get(JobState.CANCELLED),
         jobStateCountMap.get(JobState.FAILED),
         jobStateCountMap.get(JobState.SKIPPED));
@@ -385,10 +389,14 @@ public class JobsScheduler {
     LongCounter failedJobCounter = METER.counterBuilder(AppConstants.FAILED_JOB_COUNT).build();
     LongCounter cancelledJobCounter =
         METER.counterBuilder(AppConstants.CANCELLED_JOB_COUNT).build();
+    LongCounter runningJobCounter = METER.counterBuilder(AppConstants.RUNNING_JOB_COUNT).build();
+    LongCounter queuedJobCounter = METER.counterBuilder(AppConstants.QUEUED_JOB_COUNT).build();
     Attributes attributes = Attributes.of(AttributeKey.stringKey(AppConstants.TYPE), taskType);
     successfulJobCounter.add(jobStateCountMap.get(JobState.SUCCEEDED), attributes);
     failedJobCounter.add(jobStateCountMap.get(JobState.FAILED), attributes);
     cancelledJobCounter.add(jobStateCountMap.get(JobState.CANCELLED), attributes);
+    runningJobCounter.add(jobStateCountMap.get(JobState.RUNNING), attributes);
+    queuedJobCounter.add(jobStateCountMap.get(JobState.QUEUED), attributes);
     METER
         .gaugeBuilder(AppConstants.RUN_DURATION_SCHEDULER)
         .ofLongs()
@@ -493,7 +501,14 @@ public class JobsScheduler {
             .required(false)
             .hasArg()
             .longOpt("taskQueuedTimeoutMs")
-            .desc("Timeout in milliseconds for an individual task")
+            .desc("Timeout in milliseconds for an individual task in queued state")
+            .build());
+    options.addOption(
+        Option.builder(null)
+            .required(false)
+            .hasArg()
+            .longOpt("taskTimeoutMs")
+            .desc("Timeout in milliseconds for an individual task to poll jobs status")
             .build());
     // TODO: move these to ODD specific config
     options.addOption(
@@ -1035,15 +1050,18 @@ public class JobsScheduler {
           executors.getQueue().clear();
         }
         log.warn(
-            "Attempting to cancel job for {} because of timeout of {} hours", task, tasksWaitHours);
+            "Attempting to cancel task for {} because of timeout of {} hours",
+            task,
+            tasksWaitHours);
         if (taskFuture.cancel(true)) {
-          log.warn("Cancelled job for {} because of timeout of {} hours", task, tasksWaitHours);
+          log.warn("Cancelled task for {} because of timeout of {} hours", task, tasksWaitHours);
         }
       } finally {
         if (!skipStateCountUpdate) {
           if (jobState.isPresent()) {
             jobStateCountMap.put(jobState.get(), jobStateCountMap.get(jobState.get()) + 1);
           } else if (taskFuture.isCancelled()) {
+            // Even though the jobs are reported as cancelled, they might be queued or running.
             jobStateCountMap.put(JobState.CANCELLED, jobStateCountMap.get(JobState.CANCELLED) + 1);
           } else {
             // Jobs that are skipped due to replica or missing retention policy, etc.
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
@@ -1,5 +1,6 @@
 package com.linkedin.openhouse.jobs.scheduler.tasks;
 
+import com.google.common.base.Preconditions;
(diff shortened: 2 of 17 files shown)


apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
```diff
@@ -142,12 +142,13 @@ private Optional<JobState> pollJobStatus(Attributes typeAttributes) {
...
+      // Cancel job if it is queued for more than queuedTimeoutMs.
+      // Otherwise, if it is running, then keep it running until scheduler cancels it.
+      if (elapsedTime > queuedTimeoutMs) {
         Optional<JobState> job = jobsClient.getState(jobId);
-        // Do not cancel job if it is still in running state after 3 hours
-        if (job.isPresent() && !job.get().equals(JobState.RUNNING)) {
+        if (job.isPresent() && job.get().equals(JobState.QUEUED)) {
           log.info(
-              "Cancelling job: {} due to timeout for {}: jobState: {}",
+              "Cancelling job: {} due to queued timeout for {}: jobState: {}",
```
abhisheknath2011: Queued job can be successful after long queued time. I have collected some data for two weeks and it seems max queued time is 75 mins and after that job started and succeeded. I think we should just skip checking status instead of cancelling the queued tasks.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
```diff
@@ -142,12 +142,13 @@ private Optional<JobState> pollJobStatus(Attributes typeAttributes) {
     long startTime = System.currentTimeMillis();
     while (!jobFinished()) {
       long elapsedTime = System.currentTimeMillis() - startTime;
-      if (elapsedTime > timeoutMs) {
+      // Cancel job if it is queued for more than queuedTimeoutMs.
+      // Otherwise, if it is running, then keep it running until scheduler cancels it.
+      if (elapsedTime > queuedTimeoutMs) {
         Optional<JobState> job = jobsClient.getState(jobId);
-        // Do not cancel job if it is still in running state after 3 hours
-        if (job.isPresent() && !job.get().equals(JobState.RUNNING)) {
```
abhisheknath2011, line 156: On the other hand job can run for longer duration like OFD or compaction. Compaction can take more than an hour. Should we make the running job timeout configurable? For snapshot expiration we can set very less timeout. For compaction we can keep 30 min or 1 hour. If configurable we can set different timeout for different job types.  Post the timeout if the job still in running state we can just exit without cancelling the job. So we should have timeout for other job state as well not just queued ones.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
jiang95-dev: If you check the dashboard (app queue time), you will find the p99 for most of the jobs are 2min, while for OFD it is 5min. 75 mins is an extreme case and can be ignored.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
jiang95-dev: We never cancelled the running jobs, unless the scheduler reaches 12 hour SLA. I didn't change the logic here, but just shorten the wait time for **QUEUED** jobs. So there's no need for additional configuration.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
jiang95-dev: If we `just skip checking status instead of cancelling the queued tasks`, there's no gain in the throughput. Those threads will be hanging for 12 hours instead of 3 hours. Now I'm making it 10 mins.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
```diff
@@ -483,8 +487,15 @@ protected static CommandLine parseArgs(String[] args) {
...
-            .longOpt("taskTimeoutMs")
-            .desc("Timeout in milliseconds for an individual task")
+            .longOpt("taskQueuedTimeoutMs")
+            .desc("Timeout in milliseconds for an individual task in queued state")
+            .build());
+    options.addOption(
+        Option.builder(null)
+            .required(false)
+            .hasArg()
+            .longOpt("taskStatusCheckTimeoutMs")
```
teamurko: can we preserve naming for consistency

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
```diff
@@ -1021,9 +1032,9 @@ private void updateJobStateFromTaskFutures(
           executors.getQueue().clear();
         }
         log.warn(
-            "Attempting to cancel job for {} because of timeout of {} hours", task, tasksWaitHours);
+            "Attempting to stop task for {} because of timeout of {} hours", task, tasksWaitHours);
```
teamurko:
```suggestion
            "Attempting to cancel task for {} because of timeout of {} hours", task, tasksWaitHours);
```

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
```diff
@@ -47,7 +48,10 @@ public abstract class OperationTask<T extends Metadata> implements Callable<Opti
   private final long pollIntervalMs;
 
   @Getter(AccessLevel.NONE)
-  private final long timeoutMs;
+  private final long queuedTimeoutMs;
+
+  @Getter(AccessLevel.NONE)
+  private final long taskTimeoutMs; // should be larger than queuedTimeoutMs
```
teamurko, line 55: should we add precondition

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
```diff
@@ -140,42 +152,43 @@ private Optional<JobState> submitJob(Attributes typeAttributes) {
...
-        if (!jobsClient.cancelJob(jobId)) {
-          log.error("Could not cancel job {} for {}", getType(), metadata);
-          return Optional.empty();
-        }
       }
+    } catch (InterruptedException e) {
+      // Exit status check if scheduler send out an interrupt signal.
+      log.warn(
+          String.format(
+              "Exit status check: Interrupted status polling for job %s for %s.",
```
teamurko:
```suggestion
              "Exiting status check, interrupted status polling for job %s for %s.",
```

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
```diff
@@ -140,42 +152,43 @@ private Optional<JobState> submitJob(Attributes typeAttributes) {
...
+              getType(), metadata),
+          e);
     }
     Optional<JobResponseBody> ret = jobsClient.getJob(jobId);
     if (ret.isPresent()) {
-      log.info(
-          "Job: {} for {} has reached terminal state: {}", jobId, metadata, ret.get().getState());
+      log.info("Exiting: Job: {} for {} has state: {}", jobId, metadata, ret.get().getState());
```
teamurko:
```suggestion
      log.info("Finished job with id {}, state {} for {}", jobId, ret.get().getState(), metadata);
```

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
```diff
@@ -140,42 +152,43 @@ private Optional<JobState> submitJob(Attributes typeAttributes) {
 
   private Optional<JobState> pollJobStatus(Attributes typeAttributes) {
     long startTime = System.currentTimeMillis();
-    while (!jobFinished()) {
-      long elapsedTime = System.currentTimeMillis() - startTime;
-      if (elapsedTime > timeoutMs) {
+    try {
+      while (!jobFinished()) {
+        long elapsedTime = System.currentTimeMillis() - startTime;
         Optional<JobState> job = jobsClient.getState(jobId);
```
teamurko:
```suggestion
        Optional<JobState> jobState = jobsClient.getState(jobId);
```

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
```diff
@@ -140,42 +152,43 @@ private Optional<JobState> submitJob(Attributes typeAttributes) {
...
-        // Do not cancel job if it is still in running state after 3 hours
-        if (job.isPresent() && !job.get().equals(JobState.RUNNING)) {
+        // Exit status check if a job is queued for more than queuedTimeoutMs.
+        if (elapsedTime > queuedTimeoutMs) {
+          if (job.isPresent() && job.get().equals(JobState.QUEUED)) {
+            log.info(
+                "Exit status check: {} due to queued timeout for {}: jobState: {}",
```
teamurko:
```suggestion
                "Exiting status check for job {} for {} due to timeout",
```

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
abhisheknath2011: jobFinished() method is also invoking jobsClient.getState. So we should avoid invoking twice. We can refactor the code and change the while loop to do while and has finished can be checked later based on already retrieved jobState.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
```diff
@@ -140,42 +152,43 @@ private Optional<JobState> submitJob(Attributes typeAttributes) {
...
-              "Cancelling job: {} due to timeout for {}: jobState: {}",
+              "Exit status check: {} due to task timeout for {}: jobState: {}",
               getType(),
               metadata,
-              job.get());
-          if (!jobsClient.cancelJob(jobId)) {
-            log.error("Could not cancel job {} for {}", getType(), metadata);
-            return Optional.empty();
-          }
+              job.orElse(null));
```
abhisheknath2011:
```suggestion
              job.orElse(JobState.SKIPPED));
```

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
abhisheknath2011: taskStatusPollTimeoutMs or just taskTimeoutMs?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
```diff
@@ -140,42 +152,43 @@ private Optional<JobState> submitJob(Attributes typeAttributes) {
...
+                metadata,
+                job.get());
+            break;
+          }
+        }
+        // Exit status check if a job is running for more than taskTimeoutMs.
+        if (elapsedTime > taskTimeoutMs) {
           log.info(
-              "Cancelling job: {} due to timeout for {}: jobState: {}",
+              "Exit status check: {} due to task timeout for {}: jobState: {}",
```
abhisheknath2011: We need to update observability/dashboard as well if we change the log lines.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
```diff
@@ -140,42 +152,43 @@ private Optional<JobState> submitJob(Attributes typeAttributes) {
...
-          return Optional.empty();
-        }
       }
+    } catch (InterruptedException e) {
+      // Exit status check if scheduler send out an interrupt signal.
+      log.warn(
+          String.format(
+              "Exit status check: Interrupted status polling for job %s for %s.",
+              getType(), metadata),
+          e);
```
abhisheknath2011, line 195: Break the loop here with explicit break.

jiang95-dev: It's not in the loop.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
jiang95-dev: Yes. It's also necessary since we are not cancelling jobs now.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
jiang95-dev: sure.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
jiang95-dev: Yes.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/JobsScheduler.java
jiang95-dev: Sure just using the old one.

teamurko (approved):

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTask.java
```diff
@@ -148,42 +166,36 @@ private Optional<JobState> submitJob(Attributes typeAttributes) {
...
-            String.format(
-                "Interrupted status polling for job %s for %s. Cancelling the job",
-                getType(), metadata),
-            e);
-        if (!jobsClient.cancelJob(jobId)) {
-          log.error("Could not cancel job {} for {}", getType(), metadata);
-          return Optional.empty();
-        }
-      }
+      } while (jobFinished(jobState));
```
abhisheknath2011, line 188:
```suggestion
      } while (!jobFinished(jobState));
```

abhisheknath2011: The loop will not exit early on terminal state.

PR #338, opened 2025-06-12
Fix multiple instances of tables and jobs client creation issue

abhisheknath2011:
## Summary

pull request in 2-3 lines.

In the existing codebase for every operation task creation (every job submission thread) a new TablesClient and JobsClient instances are created. That results multiple such clients creation and effectively not taking the advantage of http connection pooling. Instead default connection pool (pool size is 500) is created for each new client created which will effectively exhaust all the resources on the client side. One job submission thread makes just one call to job service to launch job so JobClient with pool size of 500 is waster of resources. This is a major issue and was unnoticed so far due to the nature of sequential submit which slows down over a period of time and sequential/low connection usage. We often times get netty issues in the job scheduler logs such a resource exhaustion, netty closing connection etc. when we adjust parallelism or regular cadence execution, but the actual root cause was not noticed so far. 
In the new multi-mode operation model, jobs submission frequency is higher and submission and status check happens in parallel and hence more connection usage. So with the multi mode operation started seeing more connection issues, more frequent DNS lookup failures after couple of hours of execution. Eventually netty starts closing the connections and netty runs out of resources. So the fix done here is to use single instance of TablesClient and JobsClient with connection pooling.

## Changes

- [ ] Internal API Changes
- [x] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTaskFactory.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTaskFactory.java
@@ -1,9 +1,7 @@
 package com.linkedin.openhouse.jobs.scheduler.tasks;
 
 import com.linkedin.openhouse.jobs.client.JobsClient;
-import com.linkedin.openhouse.jobs.client.JobsClientFactory;
 import com.linkedin.openhouse.jobs.client.TablesClient;
-import com.linkedin.openhouse.jobs.client.TablesClientFactory;
 import com.linkedin.openhouse.jobs.util.Metadata;
 import java.lang.reflect.InvocationTargetException;
 import lombok.AllArgsConstructor;
@@ -12,8 +10,8 @@ import lombok.AllArgsConstructor;
 @AllArgsConstructor
 public class OperationTaskFactory<T extends OperationTask<?>> {
   private Class<T> cls;
-  private JobsClientFactory jobsClientFactory;
-  private TablesClientFactory tablesClientFactory;
+  private JobsClient jobsClient;
+  private TablesClient tablesClient;
   private long pollIntervalMs;
   private long timeoutMs;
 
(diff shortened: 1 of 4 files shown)


apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTaskFactory.java
```diff
@@ -22,11 +20,6 @@ public <S extends Metadata> T create(S metadata)
           IllegalAccessException, IllegalStateException {
     return cls.getDeclaredConstructor(
             JobsClient.class, TablesClient.class, metadata.getClass(), long.class, long.class)
-        .newInstance(
```
abhisheknath2011, line 25: This was creating new instance of table and jobs client for every operation task creation.

cbb330:
Looks like clients are stateless and use thread-safe components so no issue on sharing between threads. 

do you have a way to monitor connection usage post deploy? so we see both metric improvement and saturation

will the timeouts need tuning for shared clients? is the client properly cleaned up post-shutdown?

abhisheknath2011:
> Looks like clients are stateless and use thread-safe components so no issue on sharing between threads.
> 
> do you have a way to monitor connection usage post deploy? so we see both metric improvement and saturation
> 
> will the timeouts need tuning for shared clients? is the client properly cleaned up post-shutdown?

We should use single instance of tables and jobs client as they come with default connection pooling. If we create multiple instances there will be more such connection pool and hence leads to issues. There should not be an issues sharing single instance between multiple threads. As an example, parallel metadata fetch (with multiple threads) is using single instance of TablesClient and that works as expected. For multiple instance of clients the recommended option is not to create JobsClient or tableClient with default connection pooling instead go with create new connection strategy. But connection http pooling comes with advantages of reusing connections and hence recommended to use in the client side.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/OperationTaskFactory.java
teamurko: Nice catch!

teamurko (approved):

PR #345, opened 2025-06-25
Soft Delete functionality for HouseTableService 

Will-Lo:
## Summary
This PR has the backend changes needed to support soft deleting tables, recovery, purging soft deleted tables, and listing them

[Issue](https://github.com/linkedin/openhouse/issues/#nnn) Briefly discuss the summary of the changes made in this 
pull request in 2-3 lines.

## Changes

- [x] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

The soft delete functionality will be a new API provided by OpenHouse, augmenting its delete API.
This PR defines endpoints on the HouseTableService to support soft delete related operations, which are the following 4 APIs:
1. Soft Delete User Table (using existing deleteTable API with softDelete param)
2. Recover User Table
3. Purge User Table
4. List soft deleted user tables

Soft deleted tables will be stored in a separate MySQL table, for better separation of concerns with active user tables.
Soft deleted tables will have the same schema as user tables, but will also have an additional 2 columns:
1. deleted_at_ms (Long)
2. purge_after_ms (Long)
For an initial implementation, **purge_after_ms will be hardcoded to 7 days from the deletion time** This will be reworked to become configurable based on database level configurations.
Soft deleted tables will have a primary key of `(database_id, table_id, deleted_at_ms)`

When a table is soft deleted, it will insert a new entry into soft deleted mysql table, and delete the corresponding row from user tables.
This operation will be atomic so that any failures during insert/delete will be rolled back and the failure will propagate to the user.

When a table is recovered, it will do the reverse and delete the entry from the soft deleted mysql table and insert into the user tables.

List soft deleted user tables and purge soft deleted user tables will be used by the purger job that will run asynchronously to clean up soft deleted user tables. They will also be available to users through table service for users to gain the ability to manually restore any version (instead of popping from the top of the dropped table stack) and clean their own resources.

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/handler/OpenHouseSoftDeletedUserTableHtsApiHandler.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/handler/OpenHouseSoftDeletedUserTableHtsApiHandler.java
@@ -58,10 +58,10 @@ public class OpenHouseSoftDeletedUserTableHtsApiHandler
   }
 
   @Override
-  public ApiResponse<Void> deleteEntities(String databaseId, String tableId, Long purgeFromMs) {
+  public ApiResponse<Void> deleteEntities(String databaseId, String tableId, Long purgeAfterMs) {
     userTablesHtsApiValidator.validateDeleteEntity(
         UserTableKey.builder().databaseId(databaseId).tableId(tableId).build());
-    userTableService.purgeSoftDeletedUserTables(databaseId, tableId, purgeFromMs);
+    userTableService.purgeSoftDeletedUserTables(databaseId, tableId, purgeAfterMs);
     return ApiResponse.<Void>builder().httpStatus(HttpStatus.NO_CONTENT).build();
   }
 
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/model/SoftDeletedUserTableKey.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/model/SoftDeletedUserTableKey.java
@@ -11,7 +11,7 @@ import javax.validation.constraints.Pattern;
 import lombok.Builder;
 import lombok.Value;
 
-/** The key type for the House table storing user soft deleted tables. */
+/** The key type for the House table for storing soft deleted user tables. */
 @Builder
 @Value
 public class SoftDeletedUserTableKey {
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
@@ -173,7 +173,7 @@ public class UserHouseTablesController {
       description =
           "Delete a User House Table entry identified by databaseID and tableId. "
               + "Soft delete will remove the User House Table entry and store it in Soft Deleted User Table. with additional metadata deletedAtMs and "
-              + "purgeFromMs. Tables soft deleted will be configured with a TTL (purgeAfterMs) to be permanently deleted unless restored.",
+              + "purgeAfterMs. Tables soft deleted will be configured with a TTL (purgeAfterMs) to be permanently deleted unless restored.",
       tags = {"UserTable"})
   @ApiResponses(
       value = {
@@ -288,9 +288,9 @@ public class UserHouseTablesController {
   }
 
   @Operation(
-      summary = "Purge Soft Deleted User Tables older than purgeFromMs",
+      summary = "Purge Soft Deleted User Tables older than purgeAfterMs",
       description =
-          "Permanently deletes existing Soft Deleted User Tables older than purgeFromMs that match "
+          "Permanently deletes existing Soft Deleted User Tables older than purgeAfterMs that match "
               + "databaseId and tableId. If purgeAfterMs is not provided, it will delete all soft deleted tables for the given table id",
       tags = {"UserTable"})
   @ApiResponses(
@@ -303,9 +303,9 @@ public class UserHouseTablesController {
   public ResponseEntity<Void> purgeSoftDeletedUserTables(
       @RequestParam(value = "databaseId") String databaseId,
       @RequestParam(value = "tableId") String tableId,
-      @RequestParam(value = "purgeFromMs", required = false) Long purgeFromMs) {
+      @RequestParam(value = "purgeAfterMs", required = false) Long purgeAfterMs) {
     com.linkedin.openhouse.common.api.spec.ApiResponse<Void> apiResponse;
-    apiResponse = softDeletedTablesHtsApiHandler.deleteEntities(databaseId, tableId, purgeFromMs);
+    apiResponse = softDeletedTablesHtsApiHandler.deleteEntities(databaseId, tableId, purgeAfterMs);
     return new ResponseEntity<>(
         apiResponse.getResponseBody(), apiResponse.getHttpHeaders(), apiResponse.getHttpStatus());
   }
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/dto/mapper/SoftDeletedUserTablesMapper.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/dto/mapper/SoftDeletedUserTablesMapper.java
@@ -11,6 +11,10 @@ import org.mapstruct.MappingTarget;
 @Mapper(componentModel = "spring")
 public interface SoftDeletedUserTablesMapper {
 
+  // TODO: This should be a configurable value based off of database-level configurations
+  // Default to 7 days in seconds
+  static final int DEFAULT_PURGE_AFTER_SECONDS = 7 * 24 * 60 * 60;
+
   SoftDeletedUserTableRow toSoftDeletedUserTableRow(UserTableRow userTableRow);
 
   UserTableDto toUserTableDto(SoftDeletedUserTableRow softDeletedUserTableRow);
@@ -22,8 +26,7 @@ public interface SoftDeletedUserTablesMapper {
     Instant deleteTime = Instant.now();
     softDeletedUserTableRowBuilder.deletedAtMs(deleteTime.toEpochMilli());
 
-    // TODO: This should be a configurable value based off of database-level configurations
     softDeletedUserTableRowBuilder.purgeAfterMs(
-        deleteTime.plus(7, java.time.temporal.ChronoUnit.DAYS).toEpochMilli());
+        deleteTime.plusSeconds(DEFAULT_PURGE_AFTER_SECONDS).toEpochMilli());
   }
 }
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/model/SoftDeletedUserTableRowPrimaryKey.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/model/SoftDeletedUserTableRowPrimaryKey.java
@@ -7,19 +7,16 @@ import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
+import lombok.ToString;
 
 @Builder
 @EqualsAndHashCode
 @Getter
+@ToString
 @NoArgsConstructor
 @AllArgsConstructor(access = AccessLevel.PROTECTED)
 public class SoftDeletedUserTableRowPrimaryKey implements Serializable {
   private String tableId;
   private String databaseId;
   private Long deletedAtMs;
-
-  @Override
-  public String toString() {
-    return this.databaseId + ":" + this.tableId + ":" + this.deletedAtMs;
-  }
 }
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/SoftDeletedUserTableHtsJdbcRepository.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/SoftDeletedUserTableHtsJdbcRepository.java
@@ -76,20 +76,20 @@ public interface SoftDeletedUserTableHtsJdbcRepository
 
   /**
    * Delete all soft deleted table rows for a given databaseId and tableId that have a purgeAfterMs
-   * that is before purgeFromMs
+   * that is before purgeAfterMs
    *
    * @param databaseId
    * @param tableId
-   * @param purgeFromMs
+   * @param purgeAfterMs
    */
   @Transactional
   @Query(
       "DELETE FROM SoftDeletedUserTableRow u"
           + " WHERE lower(u.databaseId) = lower(:databaseId)"
           + " AND lower(u.tableId) = lower(:tableId)"
-          + " AND u.purgeAfterMs < :purgeFromMs")
+          + " AND u.purgeAfterMs < :purgeAfterMs")
   @Modifying
-  void deleteByDatabaseIdTableIdPurgeAfterMs(String databaseId, String tableId, Long purgeFromMs);
+  void deleteByDatabaseIdTableIdPurgeAfterMs(String databaseId, String tableId, Long purgeAfterMs);
 
   /**
    * Delete all soft deleted table rows for a given databaseId and tableId
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesService.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesService.java
@@ -81,13 +81,13 @@ public interface UserTablesService {
 
   /**
    * Delete soft deleted user tables given a databaseId, tableId that are older than the given
-   * purgeFromMs
+   * purgeAfterMs
    *
    * @param databaseId
    * @param tableId
-   * @param purgeFromMs
+   * @param purgeAfterMs
    */
-  void purgeSoftDeletedUserTables(String databaseId, String tableId, Long purgeFromMs);
+  void purgeSoftDeletedUserTables(String databaseId, String tableId, Long purgeAfterMs);
 
   /**
    * Get all soft deleted tables by filters.
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
@@ -214,21 +214,21 @@ public class UserTablesServiceImpl implements UserTablesService {
 
   /**
    * Deletes all soft deleted user tables for a given databaseId and tableId that have a
-   * purgeAfterMs earlier than purgeFromMs.
+   * purgeAfterMs earlier than purgeAfterMs.
    *
    * @param databaseId The database ID of the soft deleted user table.
    * @param tableId The table ID of the soft deleted user table.
-   * @param purgeFromMs The timestamp in milliseconds after which all soft deleted user tables
+   * @param purgeAfterMs The timestamp in milliseconds after which all soft deleted user tables
    *     should be deleted. If null, all soft deleted user tables for the given databaseId and
    *     tableId will be deleted.
    */
   @Override
-  public void purgeSoftDeletedUserTables(String databaseId, String tableId, Long purgeFromMs) {
-    if (purgeFromMs == null) {
+  public void purgeSoftDeletedUserTables(String databaseId, String tableId, Long purgeAfterMs) {
+    if (purgeAfterMs == null) {
       softDeletedHtsJdbcRepository.deleteAllByDatabaseIdTableId(databaseId, tableId);
     } else {
       softDeletedHtsJdbcRepository.deleteByDatabaseIdTableIdPurgeAfterMs(
-          databaseId, tableId, purgeFromMs);
+          databaseId, tableId, purgeAfterMs);
     }
   }
 
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/handler/SoftDeletedUserTableHtsApiHandler.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/handler/SoftDeletedUserTableHtsApiHandler.java
@@ -8,7 +8,7 @@ import com.linkedin.openhouse.housetables.api.spec.response.EntityResponseBody;
 /**
  * {@link SoftDeletedUserTableHtsApiHandler} is the API handler for soft deleted user tables. It
  * provides additional methods to restore soft deleted user tables and bulk delete them based on
- * purgeFromMs.
+ * purgeAfterMs.
  */
 public interface SoftDeletedUserTableHtsApiHandler
     extends HouseTablesApiHandler<SoftDeletedUserTableKey, UserTable> {
@@ -24,12 +24,12 @@ public interface SoftDeletedUserTableHtsApiHandler
 
   /**
    * Deletes all soft deleted user tables that match the given tableId and databaseId, and are older
-   * than purgeFromMs
+   * than purgeAfterMs
    *
    * @param databaseId
    * @param tableId
-   * @param purgeFromMs
+   * @param purgeAfterMs
    * @return
    */
-  ApiResponse<Void> deleteEntities(String databaseId, String tableId, Long purgeFromMs);
+  ApiResponse<Void> deleteEntities(String databaseId, String tableId, Long purgeAfterMs);
 }
--- a/services/housetables/src/test/java/com/linkedin/openhouse/housetables/e2e/usertable/HtsControllerTest.java
+++ b/services/housetables/src/test/java/com/linkedin/openhouse/housetables/e2e/usertable/HtsControllerTest.java
@@ -701,7 +701,7 @@ public class HtsControllerTest {
             MockMvcRequestBuilders.delete("/hts/tables/purge")
                 .param("databaseId", TEST_DB_ID)
                 .param("tableId", TEST_TABLE_ID)
-                .param("purgeFromMs", String.valueOf(purgeAfterMs + 1)))
+                .param("purgeAfterMs", String.valueOf(purgeAfterMs + 1)))
         .andExpect(status().isNoContent());
 
     // Verify it's no longer in soft deleted tables
--- a/services/housetables/src/test/java/com/linkedin/openhouse/housetables/e2e/usertable/SoftDeletedHtsRepositoryTest.java
+++ b/services/housetables/src/test/java/com/linkedin/openhouse/housetables/e2e/usertable/SoftDeletedHtsRepositoryTest.java
@@ -91,7 +91,7 @@ public class SoftDeletedHtsRepositoryTest {
             .deletedAtMs(deletedTimestamp)
             .build();
 
-    // Does not delete any soft deleted user tables because purgeFromMs
+    // Does not delete any soft deleted user tables because purgeAfterMs
     Assertions.assertDoesNotThrow(
         () ->
             softDeletedUserTableHtsJdbcRepository.deleteByDatabaseIdTableIdPurgeAfterMs(
@@ -102,7 +102,7 @@ public class SoftDeletedHtsRepositoryTest {
         softDeletedUserTableHtsJdbcRepository
             .findAllByFilters(TEST_TUPLE_1_1.getDatabaseId(), null, null, Pageable.ofSize(10))
             .getTotalElements());
-    // Does not delete any soft deleted user tables because purgeFromMs
+    // Does not delete any soft deleted user tables because purgeAfterMs
     Assertions.assertDoesNotThrow(
         () ->
             softDeletedUserTableHtsJdbcRepository.deleteByDatabaseIdTableIdPurgeAfterMs(


services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -195,7 +232,7 @@ private List<UserTableDto> listTables(UserTable userTable) {
         () ->
             StreamSupport.stream(
                     htsJdbcRepository
-                        .findAllByDatabaseIdIgnoreCase(userTable.getDatabaseId())
+                        .findAllByDatabaseIdIgnoreCaseAndDeletedIsFalse(userTable.getDatabaseId())
```
abhisheknath2011: Is the intention here to list soft deleted tables as well?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
```diff
@@ -42,19 +42,20 @@ Optional<UserTableRow> findByDatabaseIdIgnoreCaseAndTableIdIgnoreCase(
   @Query("SELECT DISTINCT databaseId FROM UserTableRow")
   Iterable<String> findAllDistinctDatabaseIds();
 
-  Iterable<UserTableRow> findAllByDatabaseIdIgnoreCase(String databaseId);
+  Iterable<UserTableRow> findAllByDatabaseIdIgnoreCaseAndDeletedIsFalse(String databaseId);
 
-  Iterable<UserTableRow> findAllByDatabaseIdAndTableIdLikeAllIgnoreCase(
+  Iterable<UserTableRow> findAllByDatabaseIdAndTableIdLikeAllIgnoreCaseAndDeletedIsFalse(
```
abhisheknath2011: So we will get all the tables by default if they are soft deleted?

abhisheknath2011: We want to filter out the soft deleted tables and how table will be handed?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
```diff
@@ -64,14 +65,16 @@ Page<UserTableRow> findAllByDatabaseIdAndTableIdLikeAllIgnoreCase(
           + "(:tableVersion IS NULL OR u.version = :tableVersion) AND "
           + "(:metadataLocation IS NULL OR u.metadataLocation = :metadataLocation) AND "
           + "(:storageType IS NULL OR u.storageType = :storageType) AND "
-          + "(:creationTime IS NULL OR u.creationTime = :creationTime)")
+          + "(:creationTime IS NULL OR u.creationTime = :creationTime) AND "
+          + "(:deleted IS FALSE OR u.deleted = :deleted)")
```
abhisheknath2011: This method seems to be filtering based on the deleted flag.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
abhisheknath2011: We might need to add query for these methods to exclude soft deleted ones.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
```diff
@@ -115,12 +120,13 @@ default void deleteById(UserTableRowPrimaryKey userTableRowPrimaryKey) {
   @Transactional
   @Modifying
   @Query(
-      "UPDATE UserTableRow table SET table.tableId = :toTableId, table.metadataLocation = :metadataLocation, table.databaseId = :toDatabaseId "
+      "UPDATE UserTableRow table SET table.tableId = :toTableId, table.metadataLocation = :metadataLocation, table.databaseId = :toDatabaseId, table.deleted = :deleted "
```
abhisheknath2011: This requires adding deleted column in mysql table as well right.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -37,6 +38,10 @@ public class UserTablesServiceImpl implements UserTablesService {
 
   @Autowired UserTablesMapper userTablesMapper;
 
+  // Max length for tableId is 128, minus 22 characters for the suffix "_deleted_" and current
```
abhisheknath2011: Some table name can be long. We can have `_del_` prefix and timestamp in sec to shorten the deleted table prefix.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -150,21 +155,53 @@ public void renameUserTable(
...
+  /**
+   * Since soft deletes require renaming a tableId internally, we want to ensure that the tableId
+   * does not exceed the column size This method truncates the tableId and appends a hash for
+   * uniqueness If the tableId does not exceed the limit, this is a noop
+   *
+   * @param tableId The original table ID.
+   * @return A unique hashed table ID for the deleted table.
+   */
+  private String shortenTableIdIfNecessary(String tableId) {
+    return tableId.length() >= MAX_TABLE_ID_LENGTH_DELETE
```
abhisheknath2011: How do we recover the original table id if needed?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
abhisheknath2011: I think list tables should also exclude the soft deleted tables.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -284,7 +322,8 @@ private List<UserTableDto> searchTables(UserTable userTable) {
                             userTable.getTableVersion(),
                             userTable.getMetadataLocation(),
                             userTable.getStorageType(),
-                            userTable.getCreationTime())
+                            userTable.getCreationTime(),
+                            false)
```
abhisheknath2011: Seems like by default deleted tables are searchable. Are we thinking of running maintenance jobs on the deleted tables as well?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
Will-Lo: This function does filter out soft deleted tables, softDeleted tables are indicated with `deleted=True`, I will update the PR description

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
Will-Lo: Ah sorry I fixed this in the paginated method but not the regular one, fixed

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
Will-Lo: `AndDeletedIsFalse` filters out soft deleted tables, since soft deleted tables are indicated with `deleted=True`

abhisheknath2011: For better readability it would be good to add the Query directly although Spring data JPA parses the method name and generates query.

rohitkum2506: +1 to to adding query via Spring. Will be easier to use the API as well.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/repository/HouseTableRepository.java
```diff
@@ -21,4 +21,6 @@ void rename(
       String toDatabaseId,
       String toTableId,
       String metadataLocation);
+
+  void softDeleteById(HouseTablePrimaryKey id);
```
rohitkum2506:
```suggestion
  void softDeleteById(HouseTablePrimaryKey houseTablePrimaryKey);
```

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/model/UserTable.java
```diff
@@ -61,6 +61,12 @@ public class UserTable {
   @JsonProperty(value = "creationTime")
   private Long creationTime;
 
+  @Schema(
+      description = "Indicates whether the table is soft deleted.",
+      example = "false",
+      defaultValue = "false")
+  private Boolean deleted;
```
rohitkum2506: A `deleted` field field may carry semantic language inference to it and requires contextual understanding to know it means soft delete.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -150,21 +155,53 @@ public void renameUserTable(
...
+      if (!htsJdbcRepository.existsById(
+          UserTableRowPrimaryKey.builder().databaseId(databaseId).tableId(tableId).build())) {
+        throw new NoSuchUserTableException(databaseId, tableId);
+      }
+      htsJdbcRepository.deleteById(
+          UserTableRowPrimaryKey.builder().databaseId(databaseId).tableId(tableId).build());
+    } else {
+      UserTableRow existingTable =
+          htsJdbcRepository
+              .findById(
```
rohitkum2506: What happens when `htsJdbcRepository.findById` fails to find table with given Id?
Should this whole method be wrapped with `htsJdbcRepository.existsById` at the top?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
Will-Lo: I was thinking the recover API would use the original table ID, and it would search through the deleted IDs that it maps to. Given the hash it should be extremely unlikely that there is a collision with a different table name.

Something like `recoverTable(databaseName, tableName, timestamp)` which performs the name calculation, finds the matching tables and then renames back to the original table name provided in the API.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
Will-Lo: it should throw if its failed to be found, works the same way as exists just different return types. Wrapping in exists is not always a safeguard in concurrent situations where it could be deleted between the exists and find check anyways

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
Will-Lo: timestamp would prefer in millis since that is more precise. Ultimately if the name is long we have to have some sort of conversion that is transparent to users, changing the prefix/suffix only affects the number of tables that need to go through the conversion.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
Will-Lo: Yes

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/model/HouseTable.java
```diff
@@ -47,4 +47,7 @@ public class HouseTable {
    * with this table.
    */
   private String storageType;
+
+  // Column to indicate if the table was soft deleted
+  private boolean deleted;
```
maluchari: markedForDeletion?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -150,21 +155,53 @@ public void renameUserTable(
...
-        UserTableRowPrimaryKey.builder().databaseId(databaseId).tableId(tableId).build())) {
-      throw new NoSuchUserTableException(databaseId, tableId);
+  public void deleteUserTable(String databaseId, String tableId, boolean isSoftDeleted) {
+    if (!isSoftDeleted) {
+      if (!htsJdbcRepository.existsById(
+          UserTableRowPrimaryKey.builder().databaseId(databaseId).tableId(tableId).build())) {
+        throw new NoSuchUserTableException(databaseId, tableId);
+      }
+      htsJdbcRepository.deleteById(
+          UserTableRowPrimaryKey.builder().databaseId(databaseId).tableId(tableId).build());
```
sumedhsakdeo: I thought we said we would rename the tableId to something like tableId@<deletion_timestamp> Is that not happening as part of soft-delete?
We need to allow for recreation of table with same name when previous table was soft-deleted.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/handler/OpenHouseSoftDeletedUserTableHtsApiHandler.java
```diff
@@ -0,0 +1,111 @@
...
+
+  @Override
+  public ApiResponse<GetAllEntityResponseBody<UserTable>> getEntities(UserTable userTable) {
+    userTablesHtsApiValidator.validateGetEntities(userTable);
+    return ApiResponse.<GetAllEntityResponseBody<UserTable>>builder()
+        .httpStatus(HttpStatus.OK)
+        .responseBody(
+            GetAllEntityResponseBody.<UserTable>builder()
+                .results(
+                    userTableService.getAllUserTables(userTable).stream()
```
abhisheknath2011:
```suggestion
                    userTableService. getAllSoftDeletedTables(userTable).stream()
```

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/handler/SoftDeletedTablesHtsApiHandler.java
```diff
@@ -0,0 +1,13 @@
+package com.linkedin.openhouse.housetables.api.handler;
+
+import com.linkedin.openhouse.common.api.spec.ApiResponse;
+import com.linkedin.openhouse.housetables.api.spec.model.SoftDeletedUserTableKey;
+import com.linkedin.openhouse.housetables.api.spec.model.UserTable;
+import com.linkedin.openhouse.housetables.api.spec.response.EntityResponseBody;
+
+public interface SoftDeletedTablesHtsApiHandler
```
abhisheknath2011:
```suggestion
public interface SoftDeletedUserTableHtsApiHandler
```
for consistency.

abhisheknath2011: Doc?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/handler/SoftDeletedTablesHtsApiHandler.java
```diff
@@ -0,0 +1,13 @@
...
+    extends HouseTablesApiHandler<SoftDeletedUserTableKey, UserTable> {
+
+  ApiResponse<EntityResponseBody<UserTable>> restoreEntity(
```
abhisheknath2011: doc?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -122,7 +162,32 @@ public ResponseEntity<Void> deleteTable(
...
+            UserTableKey.builder().tableId(tableId).databaseId(databaseId).build(), false);
+    return new ResponseEntity<>(
+        apiResponse.getResponseBody(), apiResponse.getHttpHeaders(), apiResponse.getHttpStatus());
+  }
+
+  @Operation(
+      summary = "Delete a User Table",
+      description =
+          "Delete a User House Table entry identified by databaseID and tableId. "
+              + "Soft delete will store the User House Table entry in a separate table and cleaned up at a later time unless restoreed.",
```
abhisheknath2011:
```suggestion
              + "Soft delete will store the User House Table entry in a separate table and cleaned up at a later time (configured TTL) unless restored.",
```

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -189,4 +254,61 @@ public ResponseEntity<Void> renameTable(
...
+        @ApiResponse(responseCode = "204", description = "User Table PUT: NO_CONTENT"),
+        @ApiResponse(responseCode = "400", description = "User Table PUT: BAD_REQUEST"),
+        @ApiResponse(responseCode = "404", description = "User Table PUT: TBL_DB_NOT_FOUND"),
+        @ApiResponse(responseCode = "409", description = "User Table PUT: CONFLICT")
+      })
+  @PutMapping(value = HTS_TABLES_RESTORE_ENDPOINT)
+  public ResponseEntity<EntityResponseBody<UserTable>> restoreUserTable(
+      @RequestParam(value = "databaseId") String databaseId,
+      @RequestParam(value = "tableId") String tableId,
+      @RequestParam(value = "deletedAtMs") Long deletedAtMs) {
```
abhisheknath2011: Can this be optional? if deletedAtMs is not specified can it restored from the last deleted version?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -189,4 +254,61 @@ public ResponseEntity<Void> renameTable(
...
+      value = {
+        @ApiResponse(responseCode = "204", description = "User Table PUT: NO_CONTENT"),
+        @ApiResponse(responseCode = "400", description = "User Table PUT: BAD_REQUEST"),
+        @ApiResponse(responseCode = "404", description = "User Table PUT: TBL_DB_NOT_FOUND")
+      })
+  @DeleteMapping(value = HTS_TABLES_PURGE_ENDPOINT)
+  public ResponseEntity<Void> purgeSoftDeletedUserTable(
+      @RequestParam(value = "databaseId") String databaseId,
+      @RequestParam(value = "tableId") String tableId,
+      @RequestParam(value = "deletedAtMs") Long deletedAtMs) {
```
abhisheknath2011: Do we need deletedAtMs here? Purge means should purge all the entries of soft deleted tables for the tableId and databaseId?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/SoftDeletedUserTableHtsJdbcRepository.java
```diff
@@ -0,0 +1,94 @@
...
+  Page<SoftDeletedUserTableRow> findAllByFilters(
+      String databaseId, String tableId, Long purgeAfterMs, Pageable pageable);
+
+  @Query(
+      "DELETE FROM SoftDeletedUserTableRow u"
+          + " WHERE lower(u.databaseId) = lower(:databaseId)"
+          + " AND lower(u.tableId) = lower(:tableId)"
+          + " AND u.deletedAtMs = :deletedAt")
+  @Modifying
+  void deleteByDatabaseIdTableIdDeletedAt(String databaseId, String tableId, Long deletedAt);
```
abhisheknath2011, line 75: deleteByDatabaseIdTableId is needed for purging all the rows for a table.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
Will-Lo: I was thinking of making it optional only for the client frontend on the Table Service API, HouseTableService endpoint can require this. So the list + fetch most recent can be done in the `OpenHouseInternalCatalog` instead of in HTS. It would cause more API requests from `OpenHouseInternalCatalog` is increased from 1 to 2 for restore.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
Will-Lo: I was thinking we change this to `purgeFromMs` to delete all table rows with `purgeAfterMs` < `purgeFromMs`, but let `purgeFromMs` be configurable. You're right deleting record by record isn't that useful here. Originally it was to make deletions between HDFS and MySQL more consistent but we discussed performing a mysql purge followed by a bulk HDFS purge of those folders instead, and letting ODD handle any dangling folders if HDFS deletion fails (which would have been needed anyways).

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/handler/OpenHouseSoftDeletedUserTableHtsApiHandler.java
Will-Lo: Realized this is for non-paginated API, will just not support non-paginated as I think only exposing the paginated API is sufficient

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/model/SoftDeletedUserTableKey.java
```diff
@@ -0,0 +1,44 @@
...
+      example = "my_database")
+  @JsonProperty(value = "databaseId")
+  @NotEmpty(message = "databaseId cannot be empty")
+  @Pattern(regexp = ALPHA_NUM_UNDERSCORE_REGEX, message = ALPHA_NUM_UNDERSCORE_ERROR_MSG)
+  private String databaseId;
+
+  @Schema(description = "Delete time in unix epoch milliseconds", example = "1651002318265")
+  @JsonProperty(value = "deletedAtMs")
+  @NotEmpty(message = "deletedAtMs cannot be empty")
+  private Long deletedAtMs;
```
vikrambohra: can this be the primitive long instead of object Long. This is to avoid the null case

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/model/SoftDeletedUserTableKey.java
```diff
@@ -0,0 +1,44 @@
...
+      example = "my_table")
+  @JsonProperty(value = "tableId")
+  @NotEmpty(message = "tableId cannot be empty")
+  @Pattern(regexp = ALPHA_NUM_UNDERSCORE_REGEX, message = ALPHA_NUM_UNDERSCORE_ERROR_MSG)
+  private String tableId;
+
+  @Schema(
+      description =
+          "Unique Resource identifier for the Database containing the Table. Together with tableID"
+              + " they form a composite primary key for a user table.",
```
vikrambohra: Does the description for databaseId need to change?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/response/GetAllEntityResponseBody.java
```diff
@@ -13,7 +13,7 @@
 public class GetAllEntityResponseBody<T> {
 
   @Schema(description = "List of user table objects in House table", example = "")
-  @JsonProperty(access = JsonProperty.Access.READ_ONLY)
+  @JsonProperty(access = JsonProperty.Access.READ_ONLY, defaultValue = "[]")
```
vikrambohra: Why is this change needed?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -105,6 +113,38 @@ public ResponseEntity<GetAllEntityResponseBody<UserTable>> getPaginatedUserTable
         apiResponse.getResponseBody(), apiResponse.getHttpHeaders(), apiResponse.getHttpStatus());
   }
 
+  @Operation(
+      summary = "Get Soft Deleted User Tables",
+      description =
+          "Returns paginated soft deleted user tables given databaseId, tableId, and/or purgeAfterMs. "
```
vikrambohra: nit: and/or tableId

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -105,6 +113,38 @@ public ResponseEntity<GetAllEntityResponseBody<UserTable>> getPaginatedUserTable
...
+            .build();
+    com.linkedin.openhouse.common.api.spec.ApiResponse<GetAllEntityResponseBody<UserTable>>
+        apiResponse = softDeletedTablesHtsApiHandler.getEntities(searchByTable, page, size, sortBy);
+    return new ResponseEntity<>(
+        apiResponse.getResponseBody(), apiResponse.getHttpHeaders(), apiResponse.getHttpStatus());
+  }
+
   @Operation(
       summary = "Delete a User Table",
       description = "Delete a User House Table entry identified by databaseID and tableId.",
```
vikrambohra: Please add about soft deletion to description

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -122,7 +162,33 @@ public ResponseEntity<Void> deleteTable(
...
+            UserTableKey.builder().tableId(tableId).databaseId(databaseId).build(), false);
+    return new ResponseEntity<>(
+        apiResponse.getResponseBody(), apiResponse.getHttpHeaders(), apiResponse.getHttpStatus());
+  }
+
+  @Operation(
+      summary = "Delete a User Table",
+      description =
+          "Delete a User House Table entry identified by databaseID and tableId. "
+              + "Soft delete will store the User House Table entry in a separate table and cleaned up at a later time "
```
vikrambohra: Soft delete will remove the entry from user table and store in a separate table with deleted time etc ...

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -189,4 +255,57 @@ public ResponseEntity<Void> renameTable(
     return new ResponseEntity<>(
         apiResponse.getResponseBody(), apiResponse.getHttpHeaders(), apiResponse.getHttpStatus());
   }
+
+  @Operation(
+      summary = "restore a Soft Deleted User Table",
```
vikrambohra: nit: Restore (uppercase r)

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -189,4 +255,57 @@ public ResponseEntity<Void> renameTable(
...
+                .databaseId(databaseId)
+                .tableId(tableId)
+                .deletedAtMs(deletedAtMs)
+                .build());
+    return new ResponseEntity<>(
+        apiResponse.getResponseBody(), apiResponse.getHttpHeaders(), apiResponse.getHttpStatus());
+  }
+
+  @Operation(
+      summary = "Purge a Soft Deleted User Tables older than purgeFromMs",
```
vikrambohra: nit: remove "a"

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -189,4 +255,57 @@ public ResponseEntity<Void> renameTable(
...
+      value = {
+        @ApiResponse(responseCode = "204", description = "User Table PUT: NO_CONTENT"),
+        @ApiResponse(responseCode = "400", description = "User Table PUT: BAD_REQUEST"),
+        @ApiResponse(responseCode = "404", description = "User Table PUT: TBL_DB_NOT_FOUND")
+      })
+  @DeleteMapping(value = HTS_TABLES_PURGE_ENDPOINT)
+  public ResponseEntity<Void> purgeSoftDeletedUserTable(
+      @RequestParam(value = "databaseId") String databaseId,
+      @RequestParam(value = "tableId") String tableId,
+      @RequestParam(value = "purgeFromMs") Long purgeFromMs) {
```
vikrambohra: Worth thinking if purgeFromMs can be made optional, if null then purge all soft deleted entries

services/housetables/src/main/java/com/linkedin/openhouse/housetables/dto/mapper/SoftDeletedUserTablesMapper.java
```diff
@@ -0,0 +1,29 @@
...
+  @AfterMapping
+  default void calculateDeleteTimeAndTimeToLive(
+      @MappingTarget
+          SoftDeletedUserTableRow.SoftDeletedUserTableRowBuilder softDeletedUserTableRowBuilder) {
+    Instant deleteTime = Instant.now();
+    softDeletedUserTableRowBuilder.deletedAtMs(deleteTime.toEpochMilli());
+
+    // TODO: This should be a configurable value based off of database-level configurations
+    softDeletedUserTableRowBuilder.purgeAfterMs(
+        deleteTime.plus(7, java.time.temporal.ChronoUnit.DAYS).toEpochMilli());
```
vikrambohra: make it cleaner by importing the Chronounit package.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/SoftDeletedUserTableHtsJdbcRepository.java
```diff
@@ -0,0 +1,104 @@
...
+ * JDBC-backed {@link HtsRepository} for CRUDing {@link UserTableRow}
+ *
+ * <p>This class gets configured in {@link
+ * com.linkedin.openhouse.housetables.config.db.DatabaseConfiguration} with @EnableJpaRepositories.
+ * The datasource for the Jpa repository is provided in {@link JdbcProviderConfiguration}.
+ */
+public interface SoftDeletedUserTableHtsJdbcRepository
+    extends HtsRepository<SoftDeletedUserTableRow, SoftDeletedUserTableRowPrimaryKey> {
+
+  @Query(
```
vikrambohra, line 35: Add comment

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/SoftDeletedUserTableHtsJdbcRepository.java
```diff
@@ -0,0 +1,104 @@
...
+  Page<SoftDeletedUserTableRow> findAllByFilters(
+      String databaseId, String tableId, Long purgeAfterMs, Pageable pageable);
+
+  @Query(
+      "DELETE FROM SoftDeletedUserTableRow u"
+          + " WHERE lower(u.databaseId) = lower(:databaseId)"
+          + " AND lower(u.tableId) = lower(:tableId)"
+          + " AND u.deletedAtMs = :deletedAt")
+  @Modifying
+  void deleteByDatabaseIdTableIdDeletedAt(String databaseId, String tableId, Long deletedAt);
```
vikrambohra, line 75: Add comment

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesService.java
```diff
@@ -40,8 +40,12 @@ public interface UserTablesService {
    */
   Page<UserTableDto> getAllUserTables(UserTable userTable, int page, int size, String sortBy);
 
-  /** Given a databaseId and tableId, delete the user table entry from the House Table. */
-  void deleteUserTable(String databaseId, String tableId);
+  /**
+   * Given a databaseId and tableId, delete the user table entry from the House Table. If a
+   * deletedAtMs is provided, it will be used to delete the entry in the softDeletedUserTable db
```
vikrambohra: there is no deleteAtMs in the method, then why is it mentioned in the javadoc?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesService.java
```diff
@@ -69,4 +73,37 @@ void renameUserTable(
       String toDatabaseId,
       String toTableId,
       String metadataLocation);
+
+  /**
+   * restore a soft-deleted user table identified by its databaseId, tableId, and deletedAtMs
```
vikrambohra: nit: Restore (uppercase r)

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/model/UserTable.java
```diff
@@ -61,6 +61,22 @@ public class UserTable {
   @JsonProperty(value = "creationTime")
   private Long creationTime;
 
+  @Schema(
```
vikrambohra, line 64: Why are these being added to UserTable? Can we extend this class and create a SoftDeletedUserTable with the 2 new fields?

Will-Lo:
Sorry for the long PR, it started off with only `softdelete` api but I kept adding the other endpoints as I wrote out the tests 😅 I will break it up more for the other PRs to keep within 10 files preferably.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/response/GetAllEntityResponseBody.java
Will-Lo: Sorry left that in during tests when I had purgeAfterMs as a datetimestamp, now no longer needed as it is stored as a long

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesService.java
Will-Lo: Oops, that was a LLM hallucination...fixed that. Thanks

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/model/UserTable.java
Will-Lo: I think you could go either way with this, I thought given that this entity is virtual it does not need to match 1-1 with what is stored on the DB side. When the client consumes this class it would need to map to a `HouseTable` entity, I don't think we should need to create a `SoftDeletedHouseTable` class as well unless we want 100% separation in code paths from soft deleted entities and active table entities?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
Will-Lo: Sure I'll implement that

abhisheknath2011: Yes optional purgeFromMs.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/dto/mapper/SoftDeletedUserTablesMapper.java
abhisheknath2011: Until we have database configuration may be move 7 days to constant or define at global level.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/model/SoftDeletedUserTableRowPrimaryKey.java
```diff
@@ -0,0 +1,25 @@
...
+@Getter
+@NoArgsConstructor
+@AllArgsConstructor(access = AccessLevel.PROTECTED)
+public class SoftDeletedUserTableRowPrimaryKey implements Serializable {
+  private String tableId;
+  private String databaseId;
+  private Long deletedAtMs;
+
+  @Override
+  public String toString() {
```
abhisheknath2011: Add lombok @ToString annotation instead?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -157,16 +166,69 @@ public void renameUserTable(
...
   }
 
+  @Override
+  @Transactional
+  public UserTableDto restoreUserTable(String databaseId, String tableId, Long deletedAt) {
+    SoftDeletedUserTableRowPrimaryKey softDeletedTableKey =
+        SoftDeletedUserTableRowPrimaryKey.builder()
+            .databaseId(databaseId)
+            .tableId(tableId)
+            .deletedAtMs(deletedAt)
```
abhisheknath2011, line 200: As we accept optional  deletedAt or restore latest we need to take care of this in the builder.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -157,16 +166,69 @@ public void renameUserTable(
...
+            .build();
+    SoftDeletedUserTableRow existingSoftDeletedTable =
+        softDeletedHtsJdbcRepository
+            .findById(softDeletedTableKey)
+            .orElseThrow(() -> new NoSuchUserTableException(databaseId, tableId));
+    try {
+      softDeletedHtsJdbcRepository.deleteById(softDeletedTableKey);
```
abhisheknath2011, line 207: Looks like restoring latest requires additional work to find the latest id and then delete.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/SoftDeletedUserTableHtsJdbcRepository.java
```diff
@@ -0,0 +1,104 @@
...
+          + " AND u.purgeAfterMs < :purgeFromMs")
+  @Modifying
+  void deleteByDatabaseIdTableIdPurgeAfterMs(String databaseId, String tableId, Long purgeFromMs);
+
+  @Query(
+      "SELECT CASE WHEN COUNT(u) > 0 THEN true ELSE false END FROM SoftDeletedUserTableRow u"
+          + " WHERE lower(u.databaseId) = lower(:databaseId)"
+          + " AND lower(u.tableId) = lower(:tableId)"
+          + " AND u.deletedAtMs = :deletedAt")
+  boolean existsByDatabaseIdTableIdDeletedAt(String databaseId, String tableId, Long deletedAt);
```
abhisheknath2011, line 122: We also have findByDatabaseIdTableIdDeletedAt. Do we need this?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
Will-Lo: For now on HTS side we just support deleted at a specific time, table service side can encapsulate this to grab latest on `OpenHouseInternalCatalog`

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
Will-Lo: Yes, restoring would be 2 api calls (1 get, 1 restore call). But it'd be bound to 2 calls, neither will be heavyweight because we can sort the get results in the query and . I think that is fine given that restore won't be called often. Let me know if you believe otherwise.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/SoftDeletedUserTableHtsJdbcRepository.java
Will-Lo: `existsById` is part of the interface for `HtsRepository`, I think it's fine to keep this query given that it can be used for row existence before operations

Will-Lo: Or for testing

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/model/SoftDeletedUserTableKey.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.gson.Gson;
+import io.swagger.v3.oas.annotations.media.Schema;
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.Pattern;
+import lombok.Builder;
+import lombok.Value;
+
+/** The key type for the House table storing user soft deleted tables. */
```
abhisheknath2011:
```suggestion
/** The key type for the House table for storing soft deleted user tables. */
```

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/model/UserTable.java
```diff
@@ -61,6 +61,22 @@ public class UserTable {
...
+      defaultValue = "null")
+  private Long deletedAtMs;
+
+  @Schema(
+      description =
+          "Unix timestamp in milliseconds indicating when the table will be deleted permanently by another process. "
+              + "If the current time is past the TTL, then the table could be arbitrarily deleted at any time.",
+      example = "1751907524",
+      defaultValue = "null")
+  private Long purgeAfterMs;
```
abhisheknath2011, line 78: Should this be `purgeAfterMs` or `purgeFromMs`? At API level `purgeFromMs` is used. So should be good to use one format consistently.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -105,9 +113,42 @@ public ResponseEntity<GetAllEntityResponseBody<UserTable>> getPaginatedUserTable
...
+      value = {
+        @ApiResponse(responseCode = "200", description = "User Table GET: OK"),
+        @ApiResponse(responseCode = "400", description = "User Table GET: BAD_REQUEST"),
+        @ApiResponse(responseCode = "404", description = "User Table GET: TBL_DB_NOT_FOUND")
+      })
+  @GetMapping(value = HTS_TABLES_QUERY_ENDPOINT_SOFT_DELETED)
+  public ResponseEntity<GetAllEntityResponseBody<UserTable>> getSoftDeletedUserTables(
+      @RequestParam String databaseId,
+      @RequestParam(required = false) String tableId,
+      @RequestParam(required = false) Long purgeAfterMs,
```
abhisheknath2011, line 132: Same comment here.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -189,4 +256,57 @@ public ResponseEntity<Void> renameTable(
...
+      value = {
+        @ApiResponse(responseCode = "204", description = "User Table PUT: NO_CONTENT"),
+        @ApiResponse(responseCode = "400", description = "User Table PUT: BAD_REQUEST"),
+        @ApiResponse(responseCode = "404", description = "User Table PUT: TBL_DB_NOT_FOUND")
+      })
+  @DeleteMapping(value = HTS_TABLES_PURGE_ENDPOINT)
+  public ResponseEntity<Void> purgeSoftDeletedUserTables(
+      @RequestParam(value = "databaseId") String databaseId,
+      @RequestParam(value = "tableId") String tableId,
+      @RequestParam(value = "purgeFromMs", required = false) Long purgeFromMs) {
```
abhisheknath2011: purgeFromMs/purgeAfterMs.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/spec/model/UserTable.java
Will-Lo: Good point, I think I will keep everything to `purgeAfterMs`, it's because on the API side it seems like I want to purge all datasets before <x> time, but at an entity level the entity describes itself to purge after <x> time. I think keeping it consistent is easier to reason about though.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
Will-Lo: Changed design to use an additional soft deleted table. In the old design it would call delete() but use rename under the hood on HTS, so that the audit events and logging would be consistent with a delete API.

abhisheknath2011 (approved):
Great work @Will-Lo! Thanks for iterating on the PR and addressing the comments.

PR #347, opened 2025-06-26
new endpoint for namespaceExists using database acl

cbb330:
## Summary
problem: namespaceExists is unimplemented which requires users like trino to implement client side hacks, e.g. always return true. 

solution: since OpenHouse has a table centric model, with namespace being a derived entity, we rely on our acl service to return presence of database. when openhouse implements a namespace object , this implementation will be changed to depend on that object's existence. the client API will remain the same. 

## Changes

- [X] Client-facing API Changes
- [ ] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done
new tests added

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseCatalog.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseCatalog.java
@@ -405,7 +405,11 @@ public class OpenHouseCatalog extends BaseMetastoreCatalog
                 e -> Mono.error(new WebClientRequestWithMessageException(e)))
             .blockOptional();
 
-    boolean exists = result.isPresent();
+    // Namespace exists only if we got a response AND there are ACL policies
+    boolean exists =
+        result.isPresent()
+            && result.get().getResults() != null
+            && !result.get().getResults().isEmpty();
     log.debug("Calling namespaceExists succeeded - namespace exists: {}", exists);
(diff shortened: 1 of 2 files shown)


integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseCatalog.java
```diff
@@ -384,7 +385,33 @@ public boolean removeProperties(Namespace namespace, Set<String> set)
 
   @Override
   public boolean namespaceExists(Namespace namespace) throws NoSuchNamespaceException {
-    throw new UnsupportedOperationException("Checking if database exists is not supported");
+    log.info("Calling namespaceExists with identifier: {}", namespace.toString());
```
maluchari, line 388: It is okay for this API to be not implemented in OSS until we have database as first class citizens. It is possible that tomorrow we have databases just with some properties defined without acls and that would need to be returned incase this method is invoked. I prefer to keep this stopgap solution in the extension of this class or at client rather than here.

cbb330: ack, since we implicitly assume every DB must have acl policy, which may not be the case for all consumers, will move this into a more specific LI impl of openhouse

PR #366, opened 2025-08-20
Support flags to disable maintenance

teamurko:
## Summary
This feature will be used to during migration in order to temporarily disable TMS writes.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/TableDataCompactionTask.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/TableDataCompactionTask.java
@@ -44,6 +44,8 @@ public class TableDataCompactionTask extends TableOperationTask<TableMetadata> {
 
   @Override
   protected boolean shouldRun() {
-    return metadata.isPrimary() && (metadata.isTimePartitioned() || metadata.isClustered());
+    return metadata.isPrimary()
+        && (metadata.isTimePartitioned() || metadata.isClustered())
+        && !metadata.isMaintenanceJobDisabled(OPERATION_TYPE);
   }
 }
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/TableMetadata.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/TableMetadata.java
@@ -30,9 +30,8 @@ public class TableMetadata extends Metadata {
   protected @Nullable List<ReplicationConfig> replicationConfig;
 
   public boolean isMaintenanceJobDisabled(JobConf.JobTypeEnum jobType) {
-    return Boolean.parseBoolean(jobExecutionProperties.get("maintenance.disabled"))
-        || Boolean.parseBoolean(
-            jobExecutionProperties.get(String.format("maintenance.%s.disabled", jobType)));
+    return Boolean.parseBoolean(jobExecutionProperties.get("disabled"))
+        || Boolean.parseBoolean(jobExecutionProperties.get(String.format("%s.disabled", jobType)));
   }
 
   public String fqtn() {
--- a/apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/TableMetadata.java
+++ b/apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/TableMetadata.java
@@ -1,5 +1,6 @@
 package com.linkedin.openhouse.jobs.util;
 
+import com.linkedin.openhouse.jobs.client.model.JobConf;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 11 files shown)


apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/TableDataCompactionTask.java
```diff
@@ -44,6 +44,8 @@ protected List<String> getArgs() {
 
   @Override
   protected boolean shouldRun() {
-    return metadata.isPrimary() && (metadata.isTimePartitioned() || metadata.isClustered());
+    return metadata.isPrimary()
```
cbb330: since every subclassed task needs it

and tomorrow someone may add a subclass task without it (by mistake)

shall we abstract this into the parent class?

e.g. 
```java
// parent
protected final boolean shouldRun() {
    return shouldRunImpl() && !metadata.isMaintenanceJobDisabled(OPERATION_TYPE);
}

protected abstract boolean shouldRunImpl();
```

```java
// Subclass:
@Override
protected boolean shouldRunImpl() {
    return metadata.isPrimary()
        && (metadata.isTimePartitioned() || metadata.isClustered());
}
```

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/TableDataCompactionTask.java
```diff
@@ -44,6 +44,8 @@ protected List<String> getArgs() {
...
+        && (metadata.isTimePartitioned() || metadata.isClustered())
+        && !metadata.isMaintenanceJobDisabled(OPERATION_TYPE);
```
maluchari: How do we control how this property is set? If this is open to all through a table property would that not cause other issues? For eg, I as a user, can just set a property to skip SE and have multiple snapshots or skip retention altogether? Consequently, the platform now is not in a position to control the jobs executions.

abhisheknath2011 (changes requested):
It seems like disable flag is enabled at cron job level. I believe use case is not to disable maintenance for all the tables right? If so the maintenance should be disabled only for a specific db or set of tables. In that case should we consider table level property or policy?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/TableMetadata.java
```diff
@@ -28,6 +29,11 @@ public class TableMetadata extends Metadata {
   protected @Nullable HistoryConfig historyConfig;
   protected @Nullable List<ReplicationConfig> replicationConfig;
 
+  public boolean isMaintenanceJobDisabled(JobConf.JobTypeEnum jobType) {
+    return Boolean.parseBoolean(jobExecutionProperties.get("disabled"))
```
abhisheknath2011, line 33: For replication we are checking if replication config exists or not and based on that job run decision is taken. So we can do something similar based on policy?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/scheduler/tasks/TableDataCompactionTask.java
teamurko: I agree, but this is not for user control, we can enable this only for specific databases being migrated

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/TableMetadata.java
teamurko: I don't think we have use case where we want users in general disable maintenance. It will be used by a platform team for a very specific use case.

teamurko:
> It seems like disable flag is enabled at job type (cron job) level. I believe use case is not to disable maintenance for all the tables right? If so the maintenance should be disabled only for a specific db or set of tables. In that case should we consider table level property or policy?

Yes, exactly, the flag is set on a table. maintenance.disabled=true stops all ops, maintenance.<type>.disabled=true stops one <type>

sumedhsakdeo:
> > > > why can't this behavior be controlled by policies like HISTORY and RETENTION?
> > > 
> > > 
> > > @sumedhsakdeo It's not something users are supposed to set, it's for internal usage only.
> > 
> > 
> > If we allow users to set an infinite history and retention, wouldn't the jobs be noop?
> 
> Yes, we also need to disable OFD, timestamps of renamed files we don't control (by choice). This is safer option for the use case we're handling.

OFD will take action only if snapshots are expired. Which will never happen if snaphot expiration is infinity. I am not following the timestamps argument.

I am not sure if this is a safe option. Exposing this config means anyone can set it in the middle of a job execution, when a job could succeed, fail, or becomes an orphan. There is no way to safely stop job from executing. At which point additional diable of a jobs and actually stopping the job becomes impossible to handle without race conditions.

I would suggest to keep things simple, if existing configs can be reused to control the behavior of the jobs then I would err on side of not adding another config. Only interesting case here is autocomp, which is not controlled by any behavior, but then the whole database where we want to disable is denylisted from autocomp.

sumedhsakdeo (approved):
Prefer disabling happen through configs but ok with this as stop gap as can see some benefits of this mechanism of completely turning off the valve immediately

abhisheknath2011:
> > It seems like disable flag is enabled at job type (cron job) level. I believe use case is not to disable maintenance for all the tables right? If so the maintenance should be disabled only for a specific db or set of tables. In that case should we consider table level property or policy?
> 
> @abhisheknath2011 Yes, exactly, the flag is set on a table. maintenance.disabled=true stops all ops, maintenance..disabled=true stops one

Sounds good if we can set the flag at table level.

PR #374, opened 2025-09-12
percentile histograms on by default -- which enables for catalog metadata latency

cbb330:
## Summary
the catalog metadata latency metric has high variance and is affected by size of request. so measuring by max, which is the only available metric type, does not give enough information for load testing. 

also, percentiles has a hidden key word "all" which enables turning on histogram by default for all @Timed annotated functions. so this is a bit of a refactor which will also help other metrics such as authentication timers. 

## Changes

- [ ] Internal API Changes
- [X] Bug Fixes
- [ ] New Features
- [X] Performance Improvements
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [X] Tests

## Testing Done

- [X] Manually Tested on local docker setup. Please include commands ran, and their output.

example:
image

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


services/housetables/src/main/resources/application.properties
```diff
@@ -12,11 +12,5 @@ management.endpoint.health.enabled=true
...
 management.endpoint.prometheus.enabled=true
 management.endpoint.beans.enabled=true
-management.metrics.web.server.request.autotime.percentiles=0.5,0.75,0.9,0.95,0.99
-management.metrics.distribution.percentiles-histogram.http.server.requests=true
-management.metrics.distribution.percentiles-histogram.jvm.gc.pause=true
-management.metrics.distribution.percentiles-histogram.hikaricp.connections.acquire=true
-management.metrics.distribution.percentiles-histogram.hikaricp.connections.usage=true
-management.metrics.distribution.percentiles-histogram.hikaricp.connections.creation=true
 spring.datasource.hikari.maximum-pool-size=20
-management.metrics.distribution.percentiles.http.server.requests=
```
abhisheknath2011, line 22: This is added to fix the malformed summary for histogram type metric for http server request. Here is the PR - https://github.com/linkedin/openhouse/pull/373

services/housetables/src/main/resources/application.properties
```diff
@@ -12,11 +12,5 @@ management.endpoint.health.enabled=true
 management.endpoint.shutdown.enabled=true
 management.endpoint.prometheus.enabled=true
 management.endpoint.beans.enabled=true
-management.metrics.web.server.request.autotime.percentiles=0.5,0.75,0.9,0.95,0.99
-management.metrics.distribution.percentiles-histogram.http.server.requests=true
-management.metrics.distribution.percentiles-histogram.jvm.gc.pause=true
-management.metrics.distribution.percentiles-histogram.hikaricp.connections.acquire=true
-management.metrics.distribution.percentiles-histogram.hikaricp.connections.usage=true
-management.metrics.distribution.percentiles-histogram.hikaricp.connections.creation=true
```
abhisheknath2011, line 20: Is the newly added config covers all these metrics here?

services/housetables/src/main/resources/application.properties
cbb330: Yes I saw the PR. the same behavior is held here after this change since the behavior when key is set to null

`management.metrics.distribution.percentiles.http.server.requests=`

is the same as the key missing from the properties

services/housetables/src/main/resources/application.properties
cbb330: yes, "all=true" sets every emitted timer and built in metric to a percentile. I checked in docker that the jvm metric is still being emitted

services/housetables/src/main/resources/application.properties
abhisheknath2011: Nice.

PR #509, opened 2026-03-21
Cache iceberg metadata to reduce redundant requests to storage

cbb330:
## Design Doc Summary

Full design doc: https://docs.google.com/document/d/1PjuXS4xER-oA0Q06e-qUvkrh7tSfhj5h873f7nq1rEg

### Problem

Kafka ETL reported that OpenHouse `PUT /snapshots` (commit) p99 latency is missing SLOs, sometimes exceeding 1 minute. Root cause: `refreshMetadata` is called **6 times** per commit request, but **3 of those are redundant real HDFS reads** of the same pre-commit metadata file. This happens because each `loadTable()` creates a fresh `TableOperations` instance with no shared cache.

The 6 refreshes break down as:

| # | Source | Real HDFS read? |
|---|--------|----------------|
| 1 | `TablesService.putTable` initial `findById` | Yes |
| 2 | `save()` redundant `existsById` | Yes |
| 3 | `save()` real `catalog.loadTable` | Yes |
| 4-5 | Transaction internal `ops.refresh()` | No (cheap no-ops) |
| 6 | Post-commit response building | Yes (new metadata) |

### Solution: TableMetadata Cache

- **7-day fleet analysis:** 33-38M metadata refreshes/day; ~28% are duplicates within a 10-minute window
- A `TableMetadata` cache keyed by metadata location eliminates redundant HDFS reads across `loadTable()` calls within the same pod

### Cache Configuration

| Property | Env var | Type | Default |
|---|---|---|---|
| `cluster.iceberg.tables.metadata-cache.enabled` | `CLUSTER_ICEBERG_TABLES_METADATACACHE_ENABLED` | bool | `false` |
| `cluster.iceberg.tables.metadata-cache.max-weight` | `CLUSTER_ICEBERG_TABLES_METADATACACHE_MAXWEIGHT` | DataSize (`2GB`, `512MB`) | `512MB` |
| `cluster.iceberg.tables.metadata-cache.ttl` | `CLUSTER_ICEBERG_TABLES_METADATACACHE_TTL` | Duration (`10m`, `30s`) | `10m` |

Default-off rollout: deploy fleet-wide, flip `CLUSTER_ICEBERG_TABLES_METADATACACHE_ENABLED=true` on a canary pod via its env, watch the cache metrics, then promote the value into `cluster.yaml` for the cluster. When `enabled=false`, the bean returns Spring's `NoOpCacheManager` so `@Cacheable` methods always pass through.

`max-weight` defaults to `512MB` to fit the 1GB JVM heap used in lower envs. Prod (4GB heap) can opt up to `2GB` via its `cluster.yaml`.

### Cache Metrics

| Metric | Source | What it tells you |
|---|---|---|
| `cache_gets_total{result="hit"\|"miss"}` | auto | Hit ratio = hit / (hit+miss) |
| `cache_evictions_total` | auto | Eviction rate (cause unknown) |
| `cache_eviction_weight_total` | auto | Bytes evicted — confirms weigher is live |
| `cache_size` | auto | Current weight in bytes |
| `cache_load_duration_seconds` | auto | Refresh latency on miss |
| `cache_load_total{result="success"\|"failure"}` | auto | Refresh failure rate |
| `metadata_cache_removal_total{cause}` | custom | Splits evictions by `SIZE` / `EXPIRED` / `REPLACED` / `EXPLICIT` / `COLLECTED` — tells you *why* to resize |

### Before/After Validation

- Traces are structurally identical (42 spans) — the cache only reduces latency within spans by avoiding redundant HDFS I/O
- AFTER eliminates 8 HDFS connection DEBUG lines (24 → 16)
- Single-sample latency dropped from 1643ms → 560ms; production-scale validation expected to show larger gains due to HDFS p90 variance

## Changes

- [x] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [x] Performance Improvements
- [ ] Code Style
- [x] Refactoring
- [ ] Documentation
- [x] Tests

Internal API Changes:
- Introduce `TableMetadataCache` interface and `SpringTableMetadataCache` implementation; wire a Spring `CacheManager` bean (`internalCatalogCacheManager`) backed by Caffeine.
- Bound the cache by byte weight (Caffeine `Weigher` over `TableMetadataParser.toJson` length) instead of entry count, exposed via `cluster.iceberg.tables.metadata-cache.max-weight`.

Performance Improvements:
- Eliminates redundant HDFS `TableMetadataParser.read` calls across `loadTable()` instances within a pod.

Refactoring:
- Add `RemovalListener` that emits a Micrometer counter tagged by `RemovalCause`, complementing Spring Boot's auto-instrumented `cache_*` metrics.
- Gate the whole cache on `metadata-cache.enabled`; when off, the bean is a `NoOpCacheManager` and no Caffeine resources are allocated.

Tests:
- New: `CacheConfigurationTest` covers default-disabled, enabled-with-overrides, and the `@Cacheable` round-trip; `InternalCatalogBeansTest` covers property propagation and the cross-module bean wiring for both modes.

## Testing Done

Manually tested on local docker setup (`infra/recipes/docker-compose/oh-only`):
- Brought up compose, hit `/actuator/prometheus` on the tables service: all seven cache metrics in the table above appeared with `cache="tableMetadata"` tags.
- Created/read tables under a tightened `max-weight=4KB`; observed `cache_evictions_total = metadata_cache_removal_total{cause="SIZE"} = 12`, confirming the listener fires with the correct cause tag for every eviction.

Validation run:
`JAVA_HOME=$(/usr/libexec/java_home -v 11) ./gradlew :iceberg:openhouse:internalcatalog:test :services:tables:test`

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
```diff
@@ -63,6 +66,10 @@ public class OpenHouseInternalCatalog extends BaseMetastoreCatalog {
 
   @Autowired MeterRegistry meterRegistry;
 
+  /** Cache FileIO by TableIdentifier to avoid redundant HTS lookups within the same request. */
+  private final Cache<TableIdentifier, FileIO> fileIOCache =
```
mkuchenbecker: did we do tracing that this is actually used from the 6 callsites to reduce to 1? presumably this is visible on the flame graph.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
```diff
@@ -63,6 +66,10 @@ public class OpenHouseInternalCatalog extends BaseMetastoreCatalog {
...
+      CacheBuilder.newBuilder().expireAfterWrite(30, TimeUnit.SECONDS).maximumSize(500).build();
```
mkuchenbecker: Is there a reason to expire it if its per-request? Its a maximum of one file, 50mb correct?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
```diff
@@ -269,6 +280,8 @@ protected FileIO resolveFileIO(TableIdentifier tableIdentifier) {
                 .selectStorage(tableIdentifier.namespace().toString(), tableIdentifier.name())
                 .getType();
 
-    return fileIOManager.getFileIO(type);
+    FileIO fileIO = fileIOManager.getFileIO(type);
```
mkuchenbecker: Why are we putting this here vs in fileIO manager so its transparent? (I don't konw the answer, trying to understand the trade offs.)

mkuchenbecker: I guess this makes sense, the repository is the abstraction.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/IcebergSnapshotsServiceImpl.java
```diff
@@ -79,7 +81,9 @@ public Pair<TableDto, Boolean> putIcebergSnapshots(
           databaseId, tableCreatorUpdater, Privileges.CREATE_TABLE);
     }
     try {
-      return Pair.of(openHouseInternalRepository.save(tableDtoToSave), !tableDto.isPresent());
+      return Pair.of(
+          openHouseInternalRepository.save(tableDtoToSave, rawTable.orElse(null)),
+          !tableDto.isPresent());
```
mkuchenbecker: Why are changes needed here? i don't grok.

mkuchenbecker: Why are changes needed here? i don't grok.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
```diff
@@ -63,6 +64,9 @@ public class OpenHouseInternalCatalog extends BaseMetastoreCatalog {
 
   @Autowired MeterRegistry meterRegistry;
 
+  @Autowired(required = false)
```
cbb330: will fix this, it requires test fixes.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
cbb330: I will add this (and hdfs logs) to the PR as per e2e testing in test cluster

iceberg/openhouse/internalcatalog/src/test/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperationsTest.java
```diff
@@ -1861,4 +1989,52 @@ void testDoCommitCreatesOtelSpans() {
       tracerProvider.close();
     }
   }
+
+  private static final class MapBackedRequestAttributes implements RequestAttributes {
```
cbb330: i don't like this

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/InternalCatalogProperties.java
```diff
@@ -0,0 +1,16 @@
...
+import java.time.Duration;
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/** Typed configuration for internal catalog metadata caching. */
+@Getter
+@Setter
+@ConfigurationProperties(prefix = "cluster.iceberg.metadata-cache")
+public class InternalCatalogProperties {
```
cbb330: we don't want a internalcatalogproperties at all. it should be a services/tables/TablesProperties

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/config/InternalCatalogSettings.java
```diff
@@ -0,0 +1,19 @@
...
+@Setter
+public class InternalCatalogSettings {
+
+  private MetadataCache metadataCache = new MetadataCache();
+
+  @Getter
+  @Setter
+  public static class MetadataCache {
+    private Duration ttl = Duration.ofMinutes(5);
+    private long maxSize = 1000;
```
mkuchenbecker: If this is a per request cache, why do we need more than one entity stored and a TTL?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -83,6 +84,8 @@ public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperatio
 
   FileIOManager fileIOManager;
 
+  TableMetadataCache tableMetadataCache;
```
mkuchenbecker, line 89: This can be a generic file cache no? I might call it a file cache but the variable name be metadata.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/cache/SpringTableMetadataCache.java
```diff
@@ -0,0 +1,29 @@
...
+      key = "#metadataLocation")
+  public TableMetadata load(String metadataLocation, Supplier<TableMetadata> metadataLoader) {
+    return metadataLoader.get();
+  }
+
+  @Override
+  @CachePut(
+      cacheManager = "internalCatalogCacheManager",
+      cacheNames = "tableMetadata",
+      key = "#metadataLocation")
```
mkuchenbecker, line 25: SO this abstracts the cache as a spring component?

services/tables/src/test/java/com/linkedin/openhouse/tables/config/InternalCatalogBeansTest.java
```diff
@@ -0,0 +1,76 @@
+package com.linkedin.openhouse.tables.config;
```
mkuchenbecker, line 1: I think we need a test where there are two racing commits so we can ensure stale metadata being returned from the cache does not impact.

mkuchenbecker (commented):
Is this ready for review?

services/tables/src/test/java/com/linkedin/openhouse/tables/config/InternalCatalogBeansTest.java
cbb330: I think the racing-commits case is safe here because the metadata is UUID-keyed, so concurrent commits are not reusing the same cached object.

For example, commit 1 writes uuid-1.json and atomically swaps that into HTS. If commit 2 races, it reads uuid-1.json from HTS, evolves that state, writes the updated metadata back to HTS, and then caches it as uuid-2.json. So even if an older cache entry exists, the next commit is still advancing from the HTS-backed metadata, not from stale cached metadata.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/cache/SpringTableMetadataCache.java
cbb330: Yea.`SpringTableMetadataCache` is the Spring-managed bean we inject as the implementation. This gives us runtime dependency injection so it allows us to override in the li-openhouse variant at runtime if needed.

And `@Cacheable` / `@CachePut` delegate the actual cache read/write through Spring’s cache abstraction rather than us calling Caffeine directly. This gives us auto-instrumentation among other things.

`internalCatalogCacheManager` represents a named reference to the logic actually used for implementation , and in our config that’s currently a CaffeineCacheManager.

so the end result is that callers stay behind the TableMetadataCache interface and don’t need to know about the concrete cache impl and the impl is mostly framework config.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
cbb330: there are two layers here: `internalCatalogCacheManager` can be generic for any file since it is just the Spring CacheManager, but the injected wrapper is still intentionally `TableMetadataCache` and specific to tablemetdata.

Right now the cache namespace is tableMetadata and the value type is still TableMetadata

We can easily reuse the generic internalCatalogCacheManager for other use cases later like this

```java
  @CachePut(
      cacheManager = "internalCatalogCacheManager",
      cacheNames = "newUseCase",
      key = "#newUseCaseUUID")
```

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/config/InternalCatalogSettings.java
cbb330: the per request cache is not able to emit prometheus metrics, so we can't monitor it. 

hence I'm changed strategy to use per-pod cache and key with UUID -- we get metrics, better cache-hit ratio, and don't affect our concurrency model.

mkuchenbecker: >don't affect our concurrency model.

I'm not sure.

T0 - Table Manifest State X
T1 - MAnifest is cached
T2 - Table Manifest updated to state Y
T3 - Cached value is read vs latest value and transaction fails.
T4 - Cached value is read vs latest value and transaction fails.
T5 - Cache TTLs.

A per-request cache limited lifecycle makes this a non-issue.

cbb330: "the manifest" refers to a manifest file (the Avro file that tracks a list of data files). this is distinct from "the metadata" which specifically refers to a json file that represents table state e.g.  `/data/openhouse/<db>/<table>/01226-03d77993-b04c-46a2-b3a8-dbb7b6b45783.metadata.json`

and the metadata is immutable, once the UUID is generated and file is written it can never change. 

so in the same format listed above it more accurately this:
```
T0 - Table Metadata State X
T1 - Metadata UUIDv1 is cached
T2 - Table Metadata updated to state Y
T3 - Cache miss because UUID is not found
T4 - Metadata UUIDv2 is cached
T5 - Metadata UUIDv1 TTLs
```

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/config/InternalCatalogSettings.java
```diff
@@ -0,0 +1,19 @@
...
+@Getter
+@Setter
+public class InternalCatalogSettings {
+
+  private MetadataCache metadataCache = new MetadataCache();
+
+  @Getter
+  @Setter
+  public static class MetadataCache {
+    private Duration ttl = Duration.ofMinutes(5);
```
maluchari, line 16: Was any profiling was done on per-entry TableMetadata memory size to validate the 1000-entry max ?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/cache/SpringTableMetadataCache.java
```diff
@@ -0,0 +1,29 @@
...
+import org.apache.iceberg.TableMetadata;
+import org.springframework.cache.annotation.CachePut;
+import org.springframework.cache.annotation.Cacheable;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SpringTableMetadataCache implements TableMetadataCache {
+
+  @Override
+  @Cacheable(
```
maluchari, line 13: what would happen if there are concurrent requests for same metadata location? Probably both could miss and read from HDFS.  Consider https://docs.spring.io/spring-framework/reference/integration/cache/annotations.html#cache-annotations-cacheable-synchronized ?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/cache/SpringTableMetadataCache.java
```diff
@@ -0,0 +1,29 @@
...
+      cacheManager = "internalCatalogCacheManager",
+      cacheNames = "tableMetadata",
+      key = "#metadataLocation")
+  public TableMetadata load(String metadataLocation, Supplier<TableMetadata> metadataLoader) {
+    return metadataLoader.get();
```
abhisheknath2011, line 18: Is the caching here per request basis? What happens for the scenario immediate get after put or multiople concurrent get after put?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/config/InternalCatalogSettings.java
abhisheknath2011: Trying to understand the approach here and why per pod cache? As there are 56 tables service pod, the issue with per pod cache is cache miss scenario could be more as most of the requests could land on different pods. What are we missing with request level cache? I believe idea was to optimize multiple metadata calls for a given requests.

prat0318:
The 2nd refresh linked with the source `save() redundant existsById` - can this be removed altogether as findById() should already suffice and existsById() should ideally not needed?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/cache/SpringTableMetadataCache.java
mkuchenbecker: That's fine isn't it? This strictly reduces but does not eliminate HDFS load.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/cache/SpringTableMetadataCache.java
mkuchenbecker: Per Pod is my understanding.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/config/InternalCatalogSettings.java
cbb330: the primary reason is because per request cache is too ephemeral to capture any metrics. and, our logs are only retained for 14days. so with 14d lookback we can't see a trend of before/after impact of caching. 

second reason is because per pod cache has a benefit for the getTable path. 

overall our solutions looks like this in increasing order of complexity and value.
1. no reuse
2. reuse a variable
3. reuse per request cache
4. per pod memory cache <--- we are here
5. dedicated cache service

and this PR sets the right abstractions to get to step 5if we need to in future

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/config/InternalCatalogSettings.java
cbb330: yes, but it is in progress, I will post the results on the PR as a heatmap (x axis ttl, y axis size, z is hit rate)

cbb330: I also am working on a way to visualize our efficiency intra-request. the goal is to have the right size / ttl for cache to last for whole request where request may be up to 3 min

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/cache/SpringTableMetadataCache.java
cbb330: PUT seeds cache with newest UUID after inserting to HTS.

Caching table metadata by file location is safe because the locations are immutable, UUID-based paths. Each commit writes a new file like 00003-<random-uuid>.metadata.json.

The content at a given path never changes, so the cache key always maps to the same value. There's no invalidation problem because a new version is a new key, not an update to an old one.

Stale reads are harmless because the commit path uses compare-and-swap, not the cache. if you try to commit based on an old version, the CAS check rejects it at HTS layer. The cache only affects read performance and not correctness.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
cbb330: This was added to the design doc

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/config/InternalCatalogSettings.java
```diff
@@ -13,7 +14,8 @@ public class InternalCatalogSettings {
   @Getter
   @Setter
   public static class MetadataCache {
-    private Duration ttl = Duration.ofMinutes(5);
-    private long maxSize = 1000;
+    private boolean enabled = false;
+    private Duration ttl = Duration.ofMinutes(10);
+    private DataSize maxWeight = DataSize.ofMegabytes(512);
```
mkuchenbecker, line 19: Config for these?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/cache/CacheConfiguration.java
```diff
@@ -21,15 +29,43 @@ public InternalCatalogSettings internalCatalogSettings() {
...
+    }
     CaffeineCacheManager cacheManager = new CaffeineCacheManager();
     cacheManager.setAllowNullValues(false);
     cacheManager.setCacheNames(List.of("tableMetadata"));
     cacheManager.setCaffeine(
         Caffeine.newBuilder()
             .expireAfterWrite(settings.getMetadataCache().getTtl())
-            .maximumSize(settings.getMetadataCache().getMaxSize())
+            .maximumWeight(settings.getMetadataCache().getMaxWeight().toBytes())
+            .weigher(tableMetadataWeigher())
```
mkuchenbecker, line 44: How often is this weighed? What happens at capacity? LRU eviction?

abhisheknath2011: It seems like this represent size of tablemetadata. How the weight is used?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/config/InternalCatalogSettings.java
abhisheknath2011: +1. Is the plan to inject the config from li repo as well?

services/tables/src/main/java/com/linkedin/openhouse/tables/config/InternalCatalogProperties.java
```diff
@@ -0,0 +1,23 @@
...
+import java.time.Duration;
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.util.unit.DataSize;
+
+@Getter
+@Setter
+@ConfigurationProperties(prefix = "cluster.iceberg.tables")
+public class InternalCatalogProperties {
```
abhisheknath2011, line 12: Seems like configs are defined with prefix `cluster.iceberg.tables`. Is there a corresponding property file?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/config/InternalCatalogSettings.java
cbb330: This file (`InternalCatalogSettings`) is just the **defaults**.

At runtime, three pieces work together. Think of them as **defaults → overrides → glue**:

| Class | Role |
|---|---|
| `InternalCatalogSettings` (this file) | the **defaults** and is the Pojo the code actually reads |
| `InternalCatalogProperties` | the **overrides** which is a schema object to bind from YAML/stringly typed keys under a environment prefix of `cluster.iceberg.tables` |
| `InternalCatalogBeans` | the **glue** at startup, copies any non-null override onto the defaults |

The override side is bound by Spring Boot:

```java
@ConfigurationProperties(prefix = "cluster.iceberg.tables")
public class InternalCatalogProperties { ... }
```

So setting this in li's cluster.yaml:

```yaml
cluster:
  iceberg:
    tables:
      metadata-cache:
        ttl: 10m
        max-weight: 2GB
```

…flows through automatically. the config prefix also resolves against env vars and JVM system properties, not just YAML.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/cache/CacheConfiguration.java
cbb330: Weight is how big an entry counts as. Here it is the size of the metadata in bytes.

`maxWeight` is the cap on total size of everything in the cache. So `max-`weight: 512MB means the cache holds up to ~512MB worth of metadata. 

At capacity, Caffeine kicks entries out until we're back under the cap. The eviction order isn't strict LRU, Caffeine prefers to keep entries that get hit often, with some preference for recent ones too. Close enough to LRU in practice [0].

Entries also get kicked out when they hit the TTL (10m). Either way, we emit a counter (`METADATA_CACHE_REMOVAL_CTR`) tagged with the reason (size vs. expired vs. replaced), so we can tell which is happening in metrics.



[0] - More details, the eviction policy is really `W-TinyLFU`. When the cache is full and a new entry comes in, it compares the newcomer's recent frequency to the victim LRU would pick; if the newcomer hasn't been seen often, it's rejected and the old hot entry stays. There's a small "window" segment at the front of the cache that behaves with LRU, so brand-new entries are present for some time before being judged on frequency.

services/tables/src/main/java/com/linkedin/openhouse/tables/config/InternalCatalogProperties.java
cbb330: we can use the existing cluster.yaml file  with this prefix

mkuchenbecker (approved):

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/cache/SpringTableMetadataCache.java
cbb330: Good catch, I spent some time looking deeper.

Write-side is safe since metadataLocation is unique and seed() is the only writer, so concurrent puts on the same key can't happen. 

For read-side concurrent requests (e.g. stampede on cold/evicted keys) we'll watch these signals:
1. `cache_gets_total{result=miss}` - the miss rate, especially around deploys. 
2.  the log `refreshMetadata from location {} succeeded, took {} ms`

If I see that deploys result in high #1 per pod, I will look at #2 on the same pod and seek for same metadataLocation within a small window, indicating concurrent requests. 

Because the worst case behavior is still parity with what happens today, I'll merge this PR and follow up on this post deploy.

let me know if that makes sense, I'm happy to adjust in follow up PRs.

PR #531, opened 2026-04-06
feat(optimizer): [2/N] Optimizer REST Service and Controller

mkuchenbecker:
## Optimizer Stack

| PR | Content |
|---|---|
| https://github.com/linkedin/openhouse/pull/527 | Data Model |
| https://github.com/linkedin/openhouse/pull/530 | Database Repos |
| https://github.com/linkedin/openhouse/pull/531 **(this)** | REST service |
| https://github.com/linkedin/openhouse/pull/533 | Analyzer app |
| https://github.com/linkedin/openhouse/pull/534 | Scheduler app |
| #tbd | Spark BatchedOFD app |
| #tbd | Infra, docker-compose, smoke test |

## Summary

PR 2 of N in the optimizer stack. 

Service layer and REST controllers for the optimizer service, plus the `apps/optimizer` shared module providing lightweight entity/repo copies for the analyzer and scheduler apps.

## Changes

- [ ] Client-facing API Changes
- [x] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Performance Improvements
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

**Service layer**: `OptimizerDataService` interface and `OptimizerDataServiceImpl` — CRUD operations, complete-operation lifecycle, stats upsert with history double-write, filtered queries.

**Controllers**: `TableOperationsController`, `TableOperationsHistoryController`, `TableStatsController` — REST endpoints per the design doc API spec.

**Shared module** (`apps/optimizer`): Lightweight entity and repository copies used by the analyzer and scheduler apps to read optimizer state directly from MySQL.

## Testing Done

- [ ] Manually Tested on local docker setup. Please include commands ran, and their output.
- [x] Added new tests for the changes made.
- [ ] Updated existing tests to reflect the changes made.
- [ ] No tests added or updated. Please explain why. If unsure, please feel free to ask for help.
- [ ] Some other form of testing like staging or soak time in production. Please explain.

H2 integration tests in `OptimizerDataServiceImplTest` (5 tests):
- `completeOperation_writesHistoryFromOperationRow` — saves SCHEDULED row, completes it, asserts history DTO fields
- `completeOperation_notFound_returnsEmpty` — completes nonexistent ID, asserts empty
- `upsertTableStats_createsNewRow` — upserts new table, asserts DTO and repo row
- `upsertTableStats_updatesExistingRow` — upserts twice, asserts overwrite with single row
[...]

PR #530, opened 2026-04-06
[BDP-102028] feat(optimizer): [1/N] Optimizer Database

mkuchenbecker:
## Optimizer Stack

| PR | Content |
|---|---|
| https://github.com/linkedin/openhouse/pull/527 | Data Model |
| https://github.com/linkedin/openhouse/pull/530 **(this)** | Database Repos |
| https://github.com/linkedin/openhouse/pull/531 | REST service |
| https://github.com/linkedin/openhouse/pull/533 | Analyzer app |
| https://github.com/linkedin/openhouse/pull/534 | Scheduler app |
| #tbd | Spark BatchedOFD app |
| #tbd | Infra, docker-compose, smoke test |

## Summary

PR 1 of N in the optimizer stack. 
[Overall Project](https://docs.google.com/document/d/1oGQWkmlVw0HG-D4Nx37q0oUEQd53Ni4q5Q7h1voaZIQ/edit?tab=t.0#heading=h.vtl818e4m9f7)
[Service Design doc](https://docs.google.com/document/d/1xYOD7iuPjZO05UWfT1Dkmf4lYNw4iOjqY10UT2X1FAg/edit?tab=t.0).

Spring Data JPA repositories for all four optimizer tables with filtered query support, plus tests exercising save/find, filtered queries, upsert semantics, and append-only history.

## Changes

- [ ] Client-facing API Changes
- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Performance Improvements
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

**Repositories**: `TableOperationsRepository`, `TableOperationsHistoryRepository`, `TableStatsRepository`, `TableStatsHistoryRepository` — each with JPQL filtered query methods.

**Tests**: Repository tests for all four tables plus `OptimizerServiceContextTest` verifying the Spring context loads.

## Testing Done

- [ ] Manually Tested on local docker setup. Please include commands ran, and their output.
- [x] Added new tests for the changes made.
- [ ] Updated existing tests to reflect the changes made.
- [ ] No tests added or updated. Please explain why. If unsure, please feel free to ask for help.
- [ ] Some other form of testing like staging or soak time in production. Please explain.

`./gradlew :services:optimizer:test` — all tests pass (H2 in MySQL mode).

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations
- [x] Large PR broken into smaller PRs, and PR plan linked in the description.

PR #527, opened 2026-04-03
[BDP-102028] feat(optimizer): [0/N] Optimizer API and internal model

mkuchenbecker:
## Optimizer Stack

| PR | Content |
|---|---|
| https://github.com/linkedin/openhouse/pull/527 **(this)** | API and internal models |
| https://github.com/linkedin/openhouse/pull/530 | Database Repos |
| https://github.com/linkedin/openhouse/pull/531 | REST service |
| https://github.com/linkedin/openhouse/pull/533 | Analyzer app |
| https://github.com/linkedin/openhouse/pull/534 | Scheduler app |
| #tbd | Spark BatchedOFD app |
| #tbd | Infra, docker-compose, smoke test |

## Summary

PR 0 of N in the optimizer stack. 
[Overall Project](https://docs.google.com/document/d/1oGQWkmlVw0HG-D4Nx37q0oUEQd53Ni4q5Q7h1voaZIQ/edit?tab=t.0#heading=h.vtl818e4m9f7)
[Service Design doc](https://docs.google.com/document/d/1xYOD7iuPjZO05UWfT1Dkmf4lYNw4iOjqY10UT2X1FAg/edit?tab=t.0).

Introduces the optimizer service API and internal model

image

## Changes

- [ ] Client-facing API Changes
- [x] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Performance Improvements
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

- [ ] Manually Tested on local docker setup. Please include commands ran, and their output.
- [ ] Added new tests for the changes made.
- [ ] Updated existing tests to reflect the changes made.
- [x] No tests added or updated. Please explain why. If unsure, please feel free to ask for help.
- [ ] Some other form of testing like staging or soak time in production. Please explain.

This PR contains only the data model (entities, DTOs, converters). Repository tests follow in PR 1. Verified:
- `./gradlew :services:optimizer:compileJava` passes
- `./gradlew compileJava` (full project) passes with no regressions
- Spotless formatting passes

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations
- [x] Large PR broken into smaller PRs, and PR plan linked in the description.

PR #533, opened 2026-04-07
feat(optimizer): [3/N] Analyzer

mkuchenbecker:
## Optimizer Stack

| PR | Content |
|---|---|
| https://github.com/linkedin/openhouse/pull/527 | Data Model |
| https://github.com/linkedin/openhouse/pull/530 | Database Repos |
| https://github.com/linkedin/openhouse/pull/531 | REST service |
| https://github.com/linkedin/openhouse/pull/533 **(this)** | Analyzer app |
| https://github.com/linkedin/openhouse/pull/534 | Scheduler app |
| https://github.com/linkedin/openhouse/pull/599 | Spark BatchedOFD app |
| #tbd | Infra, docker-compose, smoke test |

## Summary

PR 3 of N in the optimizer stack.

Introduces `apps/optimizer-analyzer`, a Spring Boot CommandLineRunner that evaluates every table in `table_stats` against pluggable `OperationAnalyzer` strategies. The first strategy, `OrphanFilesDeletionAnalyzer`, schedules OFD operations with 24h success / 1h failure retry cadence, a 6h SCHEDULED timeout, and a 5-strike circuit breaker.

Key design choices:
- Bulk-loads operations and history into maps (one query per type), then iterates the stats list — O(types) queries, not O(tables).
- Uses the existing generic `find()` repository methods with null params.
- Pure unit tests with Mockito — no Spring context needed.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

**Core**: `AnalyzerRunner` — loads table_stats, pre-loads operations and history into maps, evaluates each table against all analyzers, circuit breaker logic.

**Strategy interface**: `OperationAnalyzer` — `isEnabled(table)`, `shouldSchedule(table, currentOp, latestHistory)`, `getCircuitBreakerThreshold()`.

**Cadence policy**: `CadencePolicy` — encapsulates time-based retry logic shared across operation types.

**OFD analyzer**: `OrphanFilesDeletionAnalyzer` — enabled via `maintenance.optimizer.ofd.enabled` table property.

## Testing Done

25 unit tests:
- `AnalyzerRunnerTest` (7 tests) — eligible table insertion, cadence skip, disabled table, shouldSchedule=false, null UUID, circuit breaker trip, below-threshold pass
- `OrphanFilesDeletionAnalyzerTest` (18 tests) — isEnabled variants, shouldSchedule for no-op/PENDING/SCHEDULING/SCHEDULED with history combinations

```
./gradlew :apps:optimizer-analyzer:test
# BUILD SUCCESSFUL — 25 tests pass
```

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/model/TableOperationRecord.java
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Lightweight representation of an active table operation record. Mirrors the fields in {@link
+ * com.linkedin.openhouse.optimizer.entity.TableOperationRow} that the Analyzer needs.
+ */
+@Data
+@NoArgsConstructor
+public class TableOperationRecord {
+
+  private String id;
```
mkuchenbecker: @claude comments please

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,185 @@
...
+            e.getTableProperties() != null ? e.getTableProperties() : Collections.emptyMap())
+        .stats(e.getStats())
+        .build();
+  }
+
+  /**
+   * Returns {@code true} if the circuit breaker has tripped. Uses the pre-loaded history list
+   * instead of querying the DB per table.
+   */
+  private boolean isCircuitBroken(
```
mkuchenbecker: Circuit breaker should be per-analyzer. They will have different reasons. And add a todo on communication path to communicate the failure to users.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/CadencePolicy.java
```diff
@@ -0,0 +1,74 @@
...
+ * back, the SCHEDULED row would otherwise block the table forever. When the row has been SCHEDULED
+ * (or SCHEDULING) longer than {@code scheduledTimeout}, the Analyzer treats it as stale and returns
+ * {@code true}, causing a new PENDING row to be inserted.
+ */
+@RequiredArgsConstructor
+public class CadencePolicy {
+
+  private final Duration successRetryInterval;
+  private final Duration failureRetryInterval;
+  private final Duration scheduledTimeout;
```
mkuchenbecker: Explain these better in plain english with examples.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,185 @@
...
+
+  private final List<OperationAnalyzer> analyzers;
+  private final TableStatsRepository statsRepo;
+  private final TableOperationsRepository operationsRepo;
+  private final TableOperationHistoryRepository historyRepo;
+
+  /** Run the full analysis loop once. */
+  public void analyze() {
+    // Pre-load the small sides of the joins — one query per analyzer type.
+    Map<String, Map<String, TableOperationRecord>> opsByType =
```
mkuchenbecker: Add a todo to iterate over databases and run per-db.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,185 @@
...
+                    OperationAnalyzer::getOperationType,
+                    a -> loadHistoryMap(a.getOperationType())));
+
+    List<TableStatsRow> tableList =
+        statsRepo.find(null, null, null).stream()
+            .filter(row -> row.getTableUuid() != null)
+            .collect(Collectors.toList());
+    log.info("Found {} tables in optimizer table_stats", tableList.size());
+
+    tableList.forEach(
```
mkuchenbecker: add a todo for benchmarking and scale tests. Memory should be measured. If we go DB-by-db we still might end up with 10k tables per iteration, but that should be fine.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,185 @@
...
+        row -> {
+          TableSummary table = toSummary(row);
+          analyzers.forEach(
+              analyzer -> {
+                String type = analyzer.getOperationType();
```
mkuchenbecker: no need to cache this type, use use analyzer.getOperationType

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,185 @@
...
+                    OperationAnalyzer::getOperationType, a -> loadOpsMap(a.getOperationType())));
+
+    Map<String, Map<String, List<TableOperationHistoryRow>>> historyByType =
+        analyzers.stream()
+            .collect(
+                Collectors.toMap(
+                    OperationAnalyzer::getOperationType,
+                    a -> loadHistoryMap(a.getOperationType())));
+
+    List<TableStatsRow> tableList =
```
mkuchenbecker: Tables need to be onboarded otherwise we should not analyze them. Make sure we don't analyze unless the table is enabled.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,185 @@
...
+@RequiredArgsConstructor
+public class AnalyzerRunner {
+
+  private final List<OperationAnalyzer> analyzers;
+  private final TableStatsRepository statsRepo;
+  private final TableOperationsRepository operationsRepo;
+  private final TableOperationHistoryRepository historyRepo;
+
+  /** Run the full analysis loop once. */
+  public void analyze() {
```
mkuchenbecker: parameterize this. It should filter by optype, database, tablename, tableuuid all of which are optional.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,185 @@
...
+        });
+
+    log.info("Analysis complete");
+  }
+
+  /**
+   * Loads the most recent operation record per table for the given type. Deduplicates by keeping
+   * the newer row when a table has more than one active record.
+   */
+  private Map<String, TableOperationRecord> loadOpsMap(String operationType) {
```
mkuchenbecker: this function doesn't need to exist.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,185 @@
...
+                    (a, b) -> mostRecent(a, b)));
+    log.info("Analyzer {} found {} tables with operation history", operationType, map.size());
+    return map;
+  }
+
+  /**
+   * Loads all history rows for the given type and groups them by {@code tableUuid}, newest first.
+   * Called once per analyzer type to eliminate per-table N+1 queries in the circuit breaker check.
+   */
+  private Map<String, List<TableOperationHistoryRow>> loadHistoryMap(String operationType) {
```
mkuchenbecker: This function does not need to exist.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,185 @@
...
+    return historyRepo.find(operationType, null, null, null, Pageable.unpaged()).stream()
```
mkuchenbecker: Add a todo to move to a querybuilder.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,185 @@
...
+        .collect(Collectors.groupingBy(TableOperationHistoryRow::getTableUuid));
+  }
+
+  private TableOperationRow buildOperation(TableStatsRow row, String operationType) {
```
mkuchenbecker: I see the problem. We have no TableOperation entity. We should greate that class. There should be an OFD TableOperation. Then from the operation you construct a row. We should go from TAbleSummary + Additional data = TableOperation. The row is then just a tableoperation.toRow().

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,185 @@
...
+        .databaseName(row.getDatabaseId())
+        .tableName(row.getTableName())
+        .operationType(operationType)
+        .status("PENDING")
+        .createdAt(Instant.now())
+        .version(0L)
+        .build();
+  }
+
+  private TableSummary toSummary(TableStatsRow e) {
```
mkuchenbecker: This should be a member of TableStatsRow

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: circuit breaker should also reset. I think of it as backoff.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,185 @@
...
+    return allFailed;
+  }
+
+  private static TableOperationRecord mostRecent(TableOperationRecord a, TableOperationRecord b) {
+    Comparator<TableOperationRecord> byCreatedAt =
+        Comparator.comparing(r -> r.getCreatedAt() != null ? r.getCreatedAt() : Instant.EPOCH);
+    return byCreatedAt.compare(a, b) >= 0 ? a : b;
+  }
+
+  private static TableOperationRecord toRecord(TableOperationRow e) {
```
mkuchenbecker: TableOperationRecord / row conversion should happen on the record. Why we ahve the record object at all is unclear to me.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/model/TableOperationRecord.java
mkuchenbecker: Claude: Will add field-level javadoc explaining the purpose of each field and the class itself.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Claude: The threshold is already per-analyzer via `getCircuitBreakerThreshold()`, but the evaluation logic itself lives in `AnalyzerRunner`. Will move the circuit breaker check into `OperationAnalyzer` as a default method so individual analyzers can override the entire strategy, not just the threshold.

For reset/backoff: will add a time-based reset — if the most recent failure is older than a configurable window, the breaker reopens. Adding a TODO for the communication path to surface tripped breakers to users.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/CadencePolicy.java
mkuchenbecker: Claude: Will expand the javadoc on all three fields with plain-english descriptions and concrete examples (e.g., "24 hours after a successful OFD run, the analyzer will re-evaluate the table").

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Claude: Will add TODO. Current approach loads all table_stats in one query; per-db iteration would scope memory and allow incremental progress.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Claude: Will add TODO for benchmarking and scale tests — measure memory footprint at 10k tables per iteration.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Claude: Will inline — `analyzer.getOperationType()` is cheap and reads better than the local variable.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Claude: Already handled — the inner loop checks `analyzer.isEnabled(table)` before `shouldSchedule()` or `isCircuitBroken()`. Tables that are not enabled are skipped entirely with no upsert. See line 76: `if (analyzer.isEnabled(table) && ...)`.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Claude: Will add optional filter parameters to `analyze()` — opType, databaseName, tableName, tableUuid — all nullable. When set, only matching tables/analyzers are evaluated. This also supports the per-db iteration path later.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Claude: Will inline into `analyze()`.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Claude: Will inline into `analyze()`.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Claude: Will add TODO. The current JPQL optional-filter approach works but gets unwieldy as filter count grows — a Criteria API or jOOQ query builder would be cleaner.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Claude: Will introduce a `TableOperation` domain class that the analyzer produces. The flow becomes: analyzer evaluates table → returns `Optional<TableOperation>` → `TableOperation.toRow()` produces the JPA entity. `buildOperation` moves out of the runner entirely. OFD gets its own subclass if it needs operation-specific fields later.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Claude: Will add `toSummary()` as a method on `TableStatsRow` in the `apps/optimizer` shared module. Additive change only — no existing methods modified.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Claude: Will add a static factory `TableOperationRecord.from(TableOperationRow)` on the record class and remove the standalone converter method from the runner.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,131 @@
...
+
+    List<OperationAnalyzer> activeAnalyzers =
+        operationType == null
+            ? analyzers
+            : analyzers.stream()
+                .filter(a -> a.getOperationType().equals(operationType))
+                .collect(Collectors.toList());
+
+    // Pre-load the small sides of the joins — one query per analyzer type.
+    // TODO: Move to a query builder (Criteria API or jOOQ) as filter count grows.
```
mkuchenbecker: remove this todo or move it to where we have a null,null,null when querying the history repo.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/model/TableOperation.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /** Return the more recently created of two operations. */
+  public static TableOperation mostRecent(TableOperation a, TableOperation b) {
+    Comparator<TableOperation> byCreatedAt =
+        Comparator.comparing(r -> r.getCreatedAt() != null ? r.getCreatedAt() : Instant.EPOCH);
+    return byCreatedAt.compare(a, b) >= 0 ? a : b;
+  }
+
+  private static TableOperation build(
```
mkuchenbecker: Shouldn't we just use lombok builder?

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,144 @@
...
+    List<OperationAnalyzer> activeAnalyzers =
+        operationType == null
+            ? analyzers
+            : analyzers.stream()
+                .filter(a -> a.getOperationType() == operationType)
+                .collect(Collectors.toList());
+
+    // Pre-load the small sides of the joins — one query per analyzer type.
+    // TODO: Move to a query builder (Criteria API or jOOQ) as filter count grows.
+    Map<OperationType, Map<String, TableOperation>> opsByType =
```
mkuchenbecker: Lets make the operation a required parameter. If we only do one operation at a time, this function becomes way simpler.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,144 @@
...
+                        historyRepo.findLatestPerTable(a.getOperationType().name()).stream()
+                            .filter(r -> r.getTableUuid() != null)
+                            .collect(
+                                Collectors.toMap(
+                                    TableOperationHistoryRow::getTableUuid,
+                                    r -> r,
+                                    AnalyzerRunner::moreRecentHistory))));
+
+    List<Table> tables =
+        statsRepo.find(databaseName, tableName, tableUuid).stream()
```
mkuchenbecker: group by database and process database by database.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,144 @@
...
+  private final TableOperationHistoryRepository historyRepo;
+
+  /** Run the full analysis loop once with no filters. */
+  public void analyze() {
+    analyze(null, null, null, null);
+  }
+
+  /**
+   * Run the analysis loop, optionally scoped to a specific operation type, database, table name, or
+   * table UUID. Pass {@code null} for any parameter to skip that filter.
```
mkuchenbecker: Use option vs null.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/CadencePolicy.java
```diff
@@ -0,0 +1,96 @@
...
+
+  private boolean decideFromHistory(Optional<TableOperationHistoryRow> latestHistory) {
+    if (latestHistory.isEmpty()) {
+      return true;
+    }
+    TableOperationHistoryRow entry = latestHistory.get();
+    switch (entry.getStatus()) {
+      case "SUCCESS":
+        return pastInterval(entry.getCompletedAt(), successRetryInterval);
+      case "FAILED":
```
mkuchenbecker: I absolutely hate this switch statement. Naked value comparison, seems to only check for malformed data. Completely unclear the purpose of this fuction.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/CadencePolicy.java
```diff
@@ -0,0 +1,96 @@
...
+        return pastInterval(entry.getCompletedAt(), failureRetryInterval);
+      default:
+        return true;
+    }
+  }
+
+  /** {@code true} if {@code timestamp} is null or {@code interval} has elapsed since then. */
+  private boolean pastInterval(Instant timestamp, Duration interval) {
+    return timestamp == null || Duration.between(timestamp, Instant.now()).compareTo(interval) > 0;
+  }
```
mkuchenbecker: Remove this funtion. Write the code as if timestamp is not null. Just do the duration comparison.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/model/TableOperation.java
mkuchenbecker: > Shouldn't we just use lombok builder?

Done — `TableOperation` now uses Lombok `@Builder` (`apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/model/TableOperation.java:18`). The private `build()` factory is removed; `from(row)` (line 48) and `pending(table, op)` (line 61) use `TableOperation.builder()…build()`.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: > Lets make the operation a required parameter. If we only do one operation at a time, this function becomes way simpler.

Done — `analyze(OperationType)` is required (`AnalyzerRunner.java:45` and `:54`). The per-analyzer inner loop is gone; the function processes one operation type per call, and the per-type maps collapse from `Map<OperationType, Map<uuid, T>>` to plain `Map<uuid, T>`. The caller (`AnalyzerApplication`'s `CommandLineRunner`) iterates registered analyzers and calls `analyze(op)` per type.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: > group by database and process database by database.

Done — `analyze` enumerates databases via `statsRepo.findDistinctDatabaseNames()` (or honors a supplied `databaseName` filter) and dispatches to `analyzeDatabase` per db (`AnalyzerRunner.java:68-71`, `:76`). Working set is bounded by tables-per-db.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: > Use option vs null.

Done across the stack:

- Analyzer: `analyze(OperationType, Optional<String> databaseName, Optional<String> tableName, Optional<String> tableUuid)` (`AnalyzerRunner.java:54`).
- Service layer (#531): the four filter-style methods on `OptimizerDataService` (`listTableOperations`, `listTableStats`, `getStatsHistory`, `listHistory`) take `Optional<T>` for every optional parameter. Controllers wrap their nullable `@RequestParam` values with `Optional.ofNullable()`.
- JPA `@Query` repo signatures stay nullable (the `:param IS NULL OR ...` pattern is idiomatic with nullable parameters). Unwrap with `.orElse(null)` at the boundary above.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/CadencePolicy.java
mkuchenbecker: > I absolutely hate this switch statement. Naked value comparison, seems to only check for malformed data. Completely unclear the purpose of this fuction.

Switch removed (`CadencePolicy.java:44`). `shouldSchedule` now:

- Returns `false` early when there is an active non-CANCELED current op — the scheduler owns those, the analyzer stays out.
- Otherwise maps the latest history entry to `readyAfterHistoryEntry` (line 52), which parses a `HistoryStatus` enum from the entity string and picks the retry interval via a ternary. No switch; unknown statuses throw via `HistoryStatus.valueOf` at the parse boundary.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/CadencePolicy.java
mkuchenbecker: > Remove this funtion. Write the code as if timestamp is not null. Just do the duration comparison.

`pastInterval` is removed. `scheduledAt` was nullable, but the analyzer no longer inspects it — a SCHEDULED row hits the "active op != CANCELED → false" early-exit in `shouldSchedule`. `scheduledTimeout` is removed from `CadencePolicy` and from the OFD analyzer config; stuck-SCHEDULED recovery is the scheduler's job, not cadence policy's. The remaining duration comparison is inlined in `readyAfterHistoryEntry` (`CadencePolicy.java:56`) and assumes `completedAt` is non-null per the schema's NOT NULL constraint.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,141 @@
...
+   * Run the analysis loop for the given operation type, optionally scoped to a single database,
+   * table name, or table UUID. Iterates databases one at a time so the working set is bounded by
+   * tables-per-db, not tables-total.
+   */
+  public void analyze(
+      OperationType operationType,
+      Optional<String> databaseName,
+      Optional<String> tableName,
+      Optional<String> tableUuid) {
+
```
mkuchenbecker: blank line.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,141 @@
...
+        analyzers.stream().filter(a -> a.getOperationType() == operationType).findFirst();
+    if (analyzerOpt.isEmpty()) {
+      log.warn("No analyzer registered for operation type {}; skipping", operationType);
+      return;
+    }
+    OperationAnalyzer analyzer = analyzerOpt.get();
+
+    List<String> dbs = databaseName.map(List::of).orElseGet(statsRepo::findDistinctDatabaseNames);
+    log.info("Analyzing {} across {} database(s)", operationType, dbs.size());
+
```
mkuchenbecker: blank line.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,141 @@
...
+    dbs.forEach(db -> analyzeDatabase(analyzer, db, tableName, tableUuid));
+
```
mkuchenbecker: blank line

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,141 @@
...
+            .stream()
+            .filter(e -> e.getTableUuid() != null)
+            .collect(
+                Collectors.toMap(
+                    TableOperationRow::getTableUuid,
+                    TableOperation::from,
+                    TableOperation::mostRecent));
+
+    // Latest history row per (table_uuid, op_type) for this analyzer. The repo query may return
+    // tied rows on identical completed_at; dedupe in memory.
```
mkuchenbecker: identical completed_at for a table? seems rare, but possible if there was a bug or race condition. It doesn't really matter though as we are checking for existance of records with properties.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: > blank line.

Removed. The blank line between the `analyze(...)` signature and the body of the method is gone (`AnalyzerRunner.java:58→59` in `91ba3624`).

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: > blank line.

Removed. The blank line between `OperationAnalyzer analyzer = analyzerOpt.get();` and `List<String> dbs = …` is gone (`91ba3624`).

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: > blank line

Removed. The blank lines between `log.info("Analyzing …")` / `dbs.forEach(…)` / `log.info("Analysis complete …")` are gone (`91ba3624`).

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: > identical completed_at for a table? seems rare, but possible if there was a bug or race condition. It doesn't really matter though as we are checking for existance of records with properties.

Agreed — the caveat overstated the issue. Removed the two-line comment entirely in `91ba3624`. The `mostRecent` merge function is still there to satisfy `Collectors.toMap`'s duplicate-key contract, but no commentary on it.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,135 @@
...
+                    r -> r,
+                    AnalyzerRunner::moreRecentHistory));
+
+    List<Table> tables =
+        statsRepo.find(databaseName, tableName.orElse(null), tableUuid.orElse(null)).stream()
+            .filter(row -> row.getTableUuid() != null)
+            .map(Table::from)
+            .collect(Collectors.toList());
+
+    tables.forEach(
```
mkuchenbecker: Add a plain text block comment explaining what this function does.

mkuchenbecker: > Add a plain text block comment explaining what this function does.

Added a `/* ... */` block above `tables.forEach(...)` in `AnalyzerRunner.analyzeDatabase` walking through the four steps: opt-in check, current-op + latest-history lookup, delegation to `shouldSchedule`, and PENDING persistence on true (`AnalyzerRunner.java:107`).

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,146 @@
...
+      OperationType operationType,
+      Optional<String> databaseName,
+      Optional<String> tableName,
+      Optional<String> tableUuid) {
+    Optional<OperationAnalyzer> analyzerOpt =
+        analyzers.stream().filter(a -> a.getOperationType() == operationType).findFirst();
+    if (analyzerOpt.isEmpty()) {
+      log.warn("No analyzer registered for operation type {}; skipping", operationType);
+      return;
+    }
```
mkuchenbecker: impossible this is false

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,146 @@
...
+    log.info("Analysis complete for {}", operationType);
+  }
+
+  private void analyzeDatabase(
+      OperationAnalyzer analyzer,
+      String databaseName,
+      Optional<String> tableName,
+      Optional<String> tableUuid) {
+
+    com.linkedin.openhouse.optimizer.db.OperationType dbOperationType =
```
mkuchenbecker: Import vs using fully qualified names. apply the rule across all prs in the stack. make a plan.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: > add a todo for benchmarking and scale tests. Memory should be measured. If we go DB-by-db we still might end up with 10k tables per iteration, but that should be fine.

Added on `AnalyzerRunner.java` class javadoc:

```java
/**
 * Core analysis loop. For one operation type per call, iterates databases and evaluates each table
 * in a database against the matching {@link OperationAnalyzer}.
 *
 * <p>Both sides of the join — current operations and latest history per (table, type) — are loaded
 * into maps once per database before the table loop. This is correct at small scale (≤~100k
 * tables); past that the per-db query shape and projection need further tuning. Scale-up work is
 * tracked in <a href="https://linkedin.atlassian.net/browse/BDP-102182">BDP-102182</a>.
 *
 * <p>// TODO(scale-test): benchmark the per-db working set at up to 10k tables and measure JVM heap
 * residency for the three intermediate maps; per-db iteration bounds memory by tables-per-db rather
 * than tables-total, but the upper bound still needs empirical validation.
 */
```

Commit 7f513601.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: > Add a todo to move to a querybuilder.

Added above the three optional-filter `find(...)` calls in `AnalyzerRunner.analyzeDatabase`:

```java
com.linkedin.openhouse.optimizer.db.OperationType dbOperationType =
    analyzer.getOperationType().toDb();

// Pre-load the small sides of the joins — bounded by tables in this database.
// TODO(query-builder): the JPQL optional-filter shape used by these find(...) calls gets
// unwieldy as the filter count grows. Migrate to Criteria API or jOOQ once the scaffolding
// stabilizes — applies to operationsRepo.find, historyRepo.findLatestPerTable, and
// statsRepo.find below.
Map<String, TableOperation> currentOps =
    operationsRepo
        .find(
            dbOperationType, null, tableUuid.orElse(null), databaseName, tableName.orElse(null))
        ...
```

Commit 7f513601.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: > Circuit breaker should be per-analyzer. They will have different reasons. And add a todo on communication path to communicate the failure to users.

The circuit breaker is not yet implemented; the design intent is captured as a TODO on the `OperationAnalyzer` interface so the per-analyzer threshold, backoff/reset, and operator-visible-signal requirements travel together with the strategy:

```java
// OperationAnalyzer.java
/**
 * Strategy interface for a single operation type. Each implementation decides whether a given
 * table needs an operation recommendation upserted in the Optimizer Service.
 *
 * <p>// TODO(circuit-breaker): a chronically-failing table currently produces a new PENDING row
 * on every Analyzer pass. Add a circuit breaker that suppresses scheduling for a (table, type)
 * after N consecutive FAILED history entries. Requirements: configurable threshold per operation
 * type, automatic reset via exponential backoff so tables can recover, and an operator-visible
 * signal (metric or query) so tripped breakers are diagnosable.
 */
public interface OperationAnalyzer { ... }
```

This covers the original comment ("per-analyzer threshold" + "communicate the failure to users") and the follow-up about backoff/reset in the same place. Open as a future PR.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: > This should be a member of TableStatsRow

The helper this comment referred to (the runner-local `toSummary(TableStatsRow)` projection) was removed during the layering refactor. Conversion now lives on `model.Table`'s own static factory; the runner does not project rows itself:

```java
// AnalyzerRunner.analyzeDatabase
List<Table> tables =
    statsRepo.find(databaseName, tableName.orElse(null), tableUuid.orElse(null)).stream()
        .filter(row -> row.getTableUuid() != null)
        .map(Table::fromRow)
        .collect(Collectors.toList());
```

```java
// model/Table.java
public static Table fromRow(TableStatsRow row) {
  if (row == null) { return null; }
  return Table.builder()
      .tableUuid(row.getTableUuid())
      .databaseName(row.getDatabaseName())
      .tableId(row.getTableName())
      .tableProperties(
          row.getTableProperties() != null ? row.getTableProperties() : Collections.emptyMap())
      .stats(TableStats.fromRows(row.getSnapshot(), null))
      .updatedAt(row.getUpdatedAt())
      .build();
}
```

The conversion-on-the-row spirit is preserved (no standalone helper in the runner); it just hangs off `model.Table` rather than `TableStatsRow` because model-layer construction needs to happen on the model side per the api/model/db boundary.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: > remove this todo or move it to where we have a null,null,null when querying the history repo.

The TODO this comment referenced was removed during the consolidation of `analyzeDatabase`. The remaining queries no longer carry a TODO at that location, and the history query is now single-arg (no `null,null,null` pattern):

```java
// AnalyzerRunner.analyzeDatabase
Map<String, TableOperationsHistory> latestHistory =
    historyRepo.findLatestPerTable(dbOperationType).stream()
        .filter(r -> r.getTableUuid() != null)
        .map(TableOperationsHistory::fromRow)
        .collect(
            Collectors.toMap(
                TableOperationsHistory::getTableUuid,
                h -> h,
                AnalyzerRunner::moreRecentHistory));
```

The only TODOs remaining in this file are the new `TODO(scale-test)` (class javadoc) and `TODO(query-builder)` (above the find calls) added this round; the original "null,null,null" TODO is gone.

apps/optimizer/analyzerapp/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerApplication.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.domain.EntityScan;
+import org.springframework.context.annotation.Bean;
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+
+/** Entry point for the Optimizer Analyzer application. */
+@SpringBootApplication
+@EntityScan(basePackages = "com.linkedin.openhouse.optimizer.entity")
+@EnableJpaRepositories(basePackages = "com.linkedin.openhouse.optimizer.repository")
+public class AnalyzerApplication {
```
abhisheknath2011, line 15: Is the idea to run Analyzer as as service to expose an API to analyze a table, but we have optimizer as a service to expose the required APIs? If needed optimizer can internally trigger Analyzer by exposing /optimizer/analyze API. That way we can expose only one service i.e. optimizer and it APIs. If we are aligned on this, can we make Analyzer as Java application (Similar to JobsScheduler) instead running as Spring based application? We should be able to deploy using k8s cron job and can run at regular cadence (say 30 min or so). If we run as scheduled job we can collect stats on every analyzer execution.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: java limitation

apps/optimizer/analyzerapp/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerApplication.java
mkuchenbecker: Yes I see it as an endpoint long term, and even by the time we release V0.  

I see it as Interface > Analyzer module > database. Interface today is a CLI kubectl cron, we can add a persistant API in the future though. So today its invoked directly as code via kubectl. An API can host that same functionality (or a CLI for that matter)

mkuchenbecker: This same comment is true of the scheduler.

apps/optimizer-analyzer/src/main/java/com/linkedin/openhouse/analyzer/AnalyzerRunner.java
mkuchenbecker: Replaced the `analyzerOpt.isEmpty()` log-and-return with `orElseThrow(IllegalStateException)`. If a caller passes an `OperationType` with no registered analyzer, that's a wiring bug and now fails loud. Matches how `SchedulerRunner.schedule` (opt-4) handles a missing `BinPacker`.

Commit f1c500b9 on opt-3; cascaded to opt-4 (d3152274).

abhisheknath2011:
Regarding the directory structure, can we make follow two options:
- Root project can be `apps/optimizer/analyzer`
- Root project can be existing one `apps/spark` and source dir can reflect optimizer.analyzer in the package name.

apps/optimizer/analyzerapp/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerApplication.java
abhisheknath2011: As analyzer is going to a service with APIs exposed, can we move the code to the `services` module instead of keeping under `apps` module?

abhisheknath2011:
> Regarding the directory structure, can we make follow two options:
> 
> * Root project can be `apps/optimizer/analyzer`
> * Root project can be existing one `apps/spark` and source dir can reflect optimizer.analyzer in the package name.

Somehow I don't see previous comment to move Analyzer from `apps` to `services` module as Analyzer is a service. So added the comment here.

apps/optimizer/analyzerapp/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerApplication.java
mkuchenbecker: done

mkuchenbecker:
PTAL @abhisheknath2011 

Comment is here: https://github.com/linkedin/openhouse/pull/533/changes#r3262568853

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,176 @@
...
+public class AnalyzerRunner {
+
+  private final List<OperationAnalyzer> analyzers;
+  private final TableStatsRepository statsRepo;
+  private final TableOperationsRepository operationsRepo;
+  private final TableOperationsHistoryRepository historyRepo;
+
+  // Inline default also set on the field so Mockito-constructed instances (no Spring context) get
+  // a usable value; with Spring, the @Value annotation overrides this.
+  @Value("${optimizer.repo.default-limit:10000}")
```
abhisheknath2011: Is this related to max default number of tables to be processed by on execution cycle of Analyzer run? Can we update the doc here and reflect the same on the property name as well?

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,176 @@
...
+  private int defaultLimit = 10_000;
```
abhisheknath2011: If this is pagination limit the default limit looks higher.

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
mkuchenbecker: Yeah,  I didn't want unbounded data as databases continue to grow. It iterates over the DB and will process at-most 10k tables which will limit the memory usage to my guesstimate <4GB for large DBs.

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
mkuchenbecker: this is autowired so it will use the config value automatically.

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,176 @@
...
+        operationsRepo
+            .find(
+                Optional.of(analyzer.getOperationType().toDb()),
+                Optional.empty(),
+                tableUuid,
+                Optional.of(databaseName),
+                tableName,
+                Optional.empty(),
+                Optional.empty(),
+                page)
```
abhisheknath2011: Going by the paging default limit of 10k, the DBs with more that 10k tables would bot be processed. Can we ensure that all the tables within a given DB are process by adjusting the pagination list until we reach the end?

mkuchenbecker: The pagination means that we will process N batches per page. For now its just a limit but I agree we need to process pages to exhaustion. Given testing will not happen at this scale I pushed out processing to the end, but adding it now.

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,180 @@
...
+          Optional.ofNullable(currentOps.get(table.getTableUuid()));
+      Optional<TableOperationsHistoryDto> entry =
+          Optional.ofNullable(latestHistory.get(table.getTableUuid()));
+      if (!analyzer.shouldSchedule(table, currentOp, entry)) {
+        continue;
+      }
+      try {
+        TableOperationDto op = TableOperationDto.pending(table, analyzer.getOperationType());
+        operationsRepo.save(op.toRow());
+        log.debug(
```
abhisheknath2011, line 146: Shall we log this with info for observability?

mkuchenbecker: Do we need it generally? Success is the database record. Failure we need the error. I'm fine upgrading but saw it more as debug than steady-state.

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,180 @@
...
+            analyzer.getOperationType(),
+            table.getDatabaseName(),
+            table.getTableId(),
+            e.toString(),
+            e);
+        failed++;
+      }
+    }
+    log.info(
+        "Database {}: created {} PENDING {} operation(s) ({} failed)",
```
abhisheknath2011:
```suggestion
        "Finished analyzing Database {}: created {} PENDING {} operation(s) ({} failed)",
```

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,199 @@
...
+      String databaseName,
+      Optional<String> tableName,
+      Optional<String> tableUuid) {
+
+    int created = 0;
+    int failed = 0;
+    int pageNumber = 0;
+
+    while (true) {
+      PageRequest page = PageRequest.of(pageNumber, tablesPageSize);
```
mkuchenbecker: This doesn't make sense, let me fix it.

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,180 @@
...
+  public void analyze(OperationTypeDto operationType) {
+    analyze(operationType, Optional.empty(), Optional.empty(), Optional.empty());
+  }
+
+  /**
+   * Run the analysis loop for the given operation type, optionally scoped to a single database,
+   * table name, or table UUID. Iterates databases one at a time so the working set is bounded by
+   * tables-per-db, not tables-total.
+   */
+  public void analyze(
```
abhisheknath2011, line 54: We need an orchestrator to trigger analyze() at the right cadence

Since the Analyzer is a long-running service that starts up post-deployment, it won't analyze anything on its own — we need an orchestrator layer that decides when to invoke analyze() and for which databases.

Proposed trigger strategies

The orchestrator could combine one or more of the following signals:

1. Commit-driven — Trigger analysis for DBs/tables that are seeing frequent commits, since those are the ones whose statistics drift fastest and benefit most from re-analysis.
2. Time-spread coverage — Trigger analysis across all DBs on a rolling schedule, spread evenly throughout the day to avoid load spikes and ensure no DB goes stale.
3. History-driven — Use historical analysis metadata (last-analyzed timestamp, prior runtime, change rate) to prioritize which DBs to pick up next and when.

Open questions
- Where should this orchestrator live — inside the Analyzer service, or as a separate component?
- Do we want a single strategy initially, or a pluggable model so we can layer 1–3 over time?

abhisheknath2011: We could setup a recurring schedule analyzer within the same service to run analysis continuously.

abhisheknath2011: This can be done in a separate PR going forward once initial code is merged.

mkuchenbecker: > We need an orchestrator to trigger analyze() at the right cadence. Since the Analyzer is a long-running service that starts up post-deployment, it won't analyze anything on its own — we need an orchestrator layer that decides when to invoke analyze() and for which databases.

The structure I don't think we need to run this as a service V0, but rather run this as an app with a k8s cron. This can live in-parallel to the same functionality being made available as an endpoint. I don't see much value in having a cron invoke an endpoint vs calling the functionality directly, but making the functionality available via an external API will be useful later.

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
mkuchenbecker: I removed the limit for now as I think it works better when we do table joins. It is captured as part of the epic.

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
mkuchenbecker: removed pagination for now

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
mkuchenbecker: moved to info

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
mkuchenbecker: removed pagination

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
mkuchenbecker: We talked offline, this will be a cron v0 but we will expose an endpoint for adhoc callers. The availibility requirements are much lower with a cron, with the tradeoff that we need to manage parallelism ourselves. As we increase parallelism, functionality can be internalized to the service completely.

services/optimizer/analyzer/src/main/java/com/linkedin/openhouse/optimizer/analyzer/AnalyzerRunner.java
```diff
@@ -0,0 +1,174 @@
...
+      String databaseName,
+      Optional<String> tableName,
+      Optional<String> tableUuid) {
+
+    // Load the three join inputs unbounded for this database. Aligned page-by-page pagination on
+    // these maps would leave keys in one map's page mismatched with the others' — a table whose
+    // op/history happens to fall in a different page would be misread as "no current op / no
+    // history" and trigger duplicate scheduling. Correctness requires the maps to be complete
+    // relative to the tables being processed; the working set is bounded by tables-in-db, not by
+    // any per-cycle cap. Memory characterization is tracked in BDP-102738.
```
abhisheknath2011: Can we remove the internal ticket reference in the code?

abhisheknath2011 (approved):

PR #534, opened 2026-04-07
feat(optimizer): [4/N] Scheduler app

mkuchenbecker:
## Optimizer Stack

| PR | Content |
|---|---|
| https://github.com/linkedin/openhouse/pull/527 | Data Model |
| https://github.com/linkedin/openhouse/pull/530 | Database Repos |
| https://github.com/linkedin/openhouse/pull/531 | REST service |
| https://github.com/linkedin/openhouse/pull/533 | Analyzer app |
| https://github.com/linkedin/openhouse/pull/534 **(this)** | Scheduler app |
| #tbd | Spark BatchedOFD app |
| #tbd | Infra, docker-compose, smoke test |

## Summary

PR 4 of N in the optimizer stack.

Introduces `apps/optimizer-scheduler`, a Spring Boot CommandLineRunner that claims PENDING operations and submits batched Spark jobs via the Jobs Service.

image

State machine:
Analyzer creates all Operations as PENDING
1. Scheduler marks PENDING as SCHEDULING to reserve the operation for a Bin ahead of scheduling to reduce duplicate job submission.
2. After claiming, the bin is submitted as a single job. If the batch is successfully submitted, the existing operations are updated to a status of scheduled with the associated jobID persisted.
3. If the submission fails, the job will be reset to pending so another scheduler run can pick it up.
4. Duplicate pending jobs are cancelled.
5. SCHEDULING is not expected to be in that state for long, minutes at most, and any operations that are stuck in the SCHEDULING status for >T period of time should move to a state of CANCELLED so next analyzer iteration will create a new PENDING op. This might happen if the scheduler was to suddenly crash or pod rotation. It may or may not happened after the job is scheduled. Therefore its safest to transition to CANCELLED in case the job was successful but not transitioned to scheduled. This is as-compared to failing to submit a job, in which case we know the job will never complete.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

**Scheduler runner**: Loads PENDING ops, bin-packs by file count, claims via two-step CAS (PENDING → SCHEDULING → SCHEDULED), submits one Spark job per bin.

**Bin packer**: Greedy first-fit descending algorithm. Oversized tables get their own bin (never dropped). Tables with no stats default to cost 0.

**Jobs client**: WebClient-based REST client submitting `POST /jobs` to the Jobs Service with table names, operation IDs, and results endpoint.

**Repository additions**: Three `@Modifying` CAS methods on `TableOperationsRepository` — `cancelDuplicatePending`, `markScheduling`, `markScheduled` — required for safe concurrent scheduling.

## Testing Done

13 unit tests:
- `BinPackerTest` (7 tests) — empty input, single table, under/over limit, oversized table, no stats, descending sort
- `SchedulerRunnerTest` (6 tests) — no pending ops, two-step claim + schedule, launch failure, already-claimed skip, duplicate cancellation, multi-row bin claim

```
./gradlew :apps:optimizer-scheduler:test
# BUILD SUCCESSFUL — 13 tests pass
```

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/BinPacker.java
```diff
@@ -0,0 +1,74 @@
+package com.linkedin.openhouse.scheduler;
+
+import com.linkedin.openhouse.optimizer.entity.TableOperationRow;
```
mkuchenbecker: Every operation needs it own binpacker. Its different based on each operation as constraints are different. Rather than generic binpacking, we should have an OFD scheduler that is attempting to binpack pending ofd jobs. it should be a similar pattern to the analyzer where we run the scheduler requires an operation type but database and table name are optional.

So we should have a Binpacker interface and a mapping from operation to binpacker.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/BinPacker.java
```diff
@@ -0,0 +1,74 @@
...
+   * Pack {@code pending} rows into bins.
+   *
+   * @param pending operations to pack; must not be empty
+   * @param fileCountByUuid map from tableUuid to number of current data files; missing entries
+   *     default to 0
+   * @param maxFilesPerBin maximum total file count per bin
+   * @return list of bins, each bin being a non-empty list of rows
+   */
+  public static List<List<TableOperationRow>> pack(
+      List<TableOperationRow> pending, Map<String, Long> fileCountByUuid, long maxFilesPerBin) {
```
mkuchenbecker: This interface is bad and inflexible. TableOperationRow  should contain all the nessasary information. FileCount should be a field. maxFilesPerBin is specific to OFD and it should not be a aparameter but rather something done when constructing the bin packer instance (or a conf)

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,121 @@
...
+  private long maxFiles;
+
+  @Value("${scheduler.operation-type}")
+  private String operationType;
+
+  @Value("${scheduler.results-endpoint}")
+  private String resultsEndpoint;
+
+  @Transactional
+  public void schedule() {
```
mkuchenbecker: operation type required, database and table names optional.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,121 @@
...
+                    row -> row.getTableUuid(),
+                    row -> {
+                      if (row.getStats() == null || row.getStats().getSnapshot() == null) {
+                        return 0L;
+                      }
+                      Long count = row.getStats().getSnapshot().getNumCurrentFiles();
+                      return count != null ? count : 0L;
+                    }));
+
+    List<List<TableOperationRow>> bins = BinPacker.pack(pending, fileCountByUuid, maxFiles);
```
mkuchenbecker: We should not be handling rows, that is the database model. Use TableOperation from the analyzer pr.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,121 @@
...
+              TableOperationRow keep = rows.get(0);
+              operationsRepo.cancelDuplicatePending(uuid, operationType, keep.getId());
+            });
+
+    // 1. Claim all rows (PENDING → SCHEDULING) — prevents a second scheduler instance from
+    //    double-submitting. Any row already claimed (returns 0) is excluded from this batch.
+    List<TableOperationRow> claimed =
+        bin.stream()
+            .filter(
+                r -> operationsRepo.markScheduling(r.getId(), r.getVersion(), Instant.now()) == 1)
```
mkuchenbecker: Can we batch the writes to limit round trips?

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,121 @@
...
+            .map(r -> r.getDatabaseName() + "." + r.getTableName())
+            .collect(Collectors.toList());
+    List<String> opIds =
+        claimed.stream().map(TableOperationRow::getId).collect(Collectors.toList());
+
+    String jobName = "batched-" + operationType.toLowerCase() + "-" + Instant.now().toEpochMilli();
+    Optional<String> jobId =
+        jobsClient.launch(jobName, operationType, tableNames, opIds, resultsEndpoint);
+
+    if (jobId.isPresent()) {
```
mkuchenbecker: else retry?

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/BinPacker.java
mkuchenbecker: > Every operation needs it own binpacker. Its different based on each operation as constraints are different. Rather than generic binpacking, we should have an OFD scheduler that is attempting to binpack pending ofd jobs. it should be a similar pattern to the analyzer where we run the scheduler requires an operation type but database and table name are optional.
>
> So we should have a Binpacker interface and a mapping from operation to binpacker.

Done. `BinPacker` is now an interface with a single `pack(List<TableOperation>)` method. `SchedulerConfig` defines a `Map<OperationType, BinPacker>` bean — `ORPHAN_FILES_DELETION` is wired to a `FileCountBinPacker` instance. `SchedulerApplication`'s `CommandLineRunner` loops over the map's keys and calls `runner.schedule(opType)` per entry, mirroring `AnalyzerApplication`. `SchedulerRunner.schedule(OperationType, Optional<String> db, Optional<String> tableName)` is the new signature.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/BinPacker.java
mkuchenbecker: > This interface is bad and inflexible. TableOperationRow should contain all the nessasary information. FileCount should be a field. maxFilesPerBin is specific to OFD and it should not be a aparameter but rather something done when constructing the bin packer instance (or a conf)

Done. `fileCount` is now a nullable field on the shared `TableOperation` domain object — populated by the scheduler at read time from `table_stats`. `FileCountBinPacker.pack(...)` reads it off each `TableOperation` directly; no more `Map<uuid, fileCount>` parameter. `maxFilesPerBin` is constructor-injected on `FileCountBinPacker` via the `scheduler.ofd.max-files-per-bin` config; not on `pack()`.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
mkuchenbecker: > operation type required, database and table names optional.

Done. New signature: `schedule(OperationType op)` and `schedule(OperationType op, Optional<String> databaseName, Optional<String> tableName)` (`SchedulerRunner.java:43,50`). Operation type is required; database and table name are `Optional<String>`. Mirrors `AnalyzerRunner.analyze`.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
mkuchenbecker: > We should not be handling rows, that is the database model. Use TableOperation from the analyzer pr.

Done. `TableOperation` (and its companion enums) moved from the analyzer-internal package to `apps/optimizer/.../model/` so both the analyzer and the scheduler consume the same domain type. `SchedulerRunner` now loads rows, converts them via `TableOperation.from(row)`, enriches each with `fileCount` from `table_stats`, and hands `List<TableOperation>` to the `BinPacker`. Per-bin logic and the BinPacker interface work on `TableOperation`, not the JPA row.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
mkuchenbecker: > Can we batch the writes to limit round trips?

Done. Added `cancelDuplicatePendingBatch`, `markSchedulingBatch`, `markScheduledBatch`, and `markPendingBatch` `@Modifying @Query` methods on `TableOperationsRepository`. Each issues one UPDATE/DELETE for a list of IDs with the status-guarded WHERE clause preserved. `SchedulerRunner.submitBin` now uses these instead of the per-row variants.

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
mkuchenbecker: > else retry?

Done. On job-launch failure (`jobsClient.launch` returns `Optional.empty()`), `SchedulerRunner` now calls `markPendingBatch(claimed-ids)` to revert the claimed rows from SCHEDULING back to PENDING. The next scheduler pass picks them up. Replaces the previous log-and-leak that relied on the analyzer's now-removed scheduledTimeout.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,142 @@
...
+    if (packer == null) {
+      log.warn("No BinPacker registered for operation type {}; skipping", operationType);
+      return;
+    }
+
+    com.linkedin.openhouse.optimizer.db.OperationType dbOperationType = operationType.toDb();
+    List<TableOperationsRow> pendingRows =
+        operationsRepo.find(
+            dbOperationType,
+            com.linkedin.openhouse.optimizer.db.OperationStatus.PENDING,
```
mkuchenbecker: Import vs using fully qualified.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,142 @@
...
+            databaseName.orElse(null),
+            tableName.orElse(null));
+    if (pendingRows.isEmpty()) {
+      log.info("No PENDING operations of type {}; nothing to schedule", operationType);
+      return;
+    }
+
+    Set<String> uuids =
+        pendingRows.stream().map(TableOperationsRow::getTableUuid).collect(Collectors.toSet());
+    Map<String, Long> fileCountByUuid =
```
mkuchenbecker: becuase this is an OFD bin packing, we need to use the file count. We should convert TableStatsRow to model.TAblestats. Then we pass the TableStats to the binpacker. 

Extracting should be done in the bin backer, and TableStats is the API surface.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/BinPacker.java
```diff
@@ -0,0 +1,20 @@
...
+ * strategy class.
+ */
+public interface BinPacker {
+
+  /**
+   * Pack {@code pending} into one or more bins. Each returned bin is a non-empty list of
+   * operations. The relative order of operations within a bin is implementation-defined; the
+   * returned list of bins is what the scheduler dispatches one job per.
+   */
+  List<List<TableOperation>> pack(List<TableOperation> pending);
```
mkuchenbecker: Lets ahve this return a bin, and a bin has a schedule function. On failure to schedule, we clean up. On success we update teh jobID / status. The scheduler handles teh clean up / update but the bin schedules itself.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
mkuchenbecker: > Import vs using fully qualified.

Done. `SchedulerRunner` now imports `db.OperationStatus` and uses the simple name. For `db.OperationType` — which collides with the already-imported `model.OperationType` — I removed the local `dbOperationType` variable and let the value flow through at the call site so the db type name never appears as a type in this file:

```java
// SchedulerRunner.java
import com.linkedin.openhouse.optimizer.db.OperationStatus;
...
List<TableOperationsRow> pendingRows =
    operationsRepo.find(
        operationType.toDb(),
        OperationStatus.PENDING,
        null,
        databaseName.orElse(null),
        tableName.orElse(null));
```

Commit 9ad8861a.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
mkuchenbecker: > becuase this is an OFD bin packing, we need to use the file count. We should convert TableStatsRow to model.TAblestats. Then we pass the TableStats to the binpacker.
>
> Extracting should be done in the bin backer, and TableStats is the API surface.

Done. Per follow-up direction, `TableOperation` does not carry stats; the scheduler does the lookup at scheduling time and passes `(operations, statsByUuid)` to the packer. `TableStats` is the cost source at the `BinPacker` interface; `FileCountBinPacker` projects it down to a per-op `long` internally and does not retain the full payload.

`SchedulerRunner` — stats lookup at scheduling time, passed to packer:
```java
List<TableOperation> pending =
    pendingRows.stream().map(TableOperation::fromRow).collect(Collectors.toList());

Set<String> uuids =
    pending.stream().map(TableOperation::getTableUuid).collect(Collectors.toSet());
Map<String, TableStats> statsByUuid =
    statsRepo.findAllById(uuids).stream()
        .collect(Collectors.toMap(TableStatsRow::getTableUuid, TableStats::fromRow));

List<Bin> bins = packer.pack(pending, statsByUuid);
```

`BinPacker` interface:
```java
List<Bin> pack(List<TableOperation> pending, Map<String, TableStats> statsByUuid);
```

`FileCountBinPacker` — extraction in the packer; projects to a `long` once:
```java
Map<String, Long> costByOperationId =
    pending.stream()
        .collect(Collectors.toMap(TableOperation::getId, op -> cost(op, statsByUuid)));
...
private static long cost(TableOperation op, Map<String, TableStats> statsByUuid) {
  TableStats stats = statsByUuid.get(op.getTableUuid());
  if (stats == null || stats.getSnapshot() == null) {
    return 0L;
  }
  Long n = stats.getSnapshot().getNumCurrentFiles();
  return n != null ? n : 0L;
}
```

Also dropped the `Long fileCount` enrichment field from `model.TableOperation` (commit b6c7f427 on opt-0, cascaded). Scheduler commit 9ad8861a.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/BinPacker.java
mkuchenbecker: > Lets ahve this return a bin, and a bin has a schedule function. On failure to schedule, we clean up. On success we update teh jobID / status. The scheduler handles teh clean up / update but the bin schedules itself.

Done. `BinPacker.pack(...)` now returns `List<Bin>`. `Bin` owns the operations it groups together and has a `schedule()` method that submits a single Spark job and returns the job id. `SchedulerRunner` orchestrates dedup + claim before the call and mark-scheduled / revert-to-pending after.

`BinPacker`:
```java
List<Bin> pack(List<TableOperation> pending, Map<String, TableStats> statsByUuid);
```

`Bin`:
```java
public class Bin {
  @Getter private final OperationType operationType;
  @Getter private final List<TableOperation> operations;

  public List<String> getOperationIds() { ... }
  public List<String> getTableNames() { ... }  // "db.table" per op

  public Optional<String> schedule(JobsServiceClient client, String resultsEndpoint) {
    String jobName =
        "batched-" + operationType.name().toLowerCase() + "-" + Instant.now().toEpochMilli();
    return client.launch(
        jobName, operationType.name(), getTableNames(), getOperationIds(), resultsEndpoint);
  }
}
```

`SchedulerRunner.submitBin` — cleanup/update around `bin.schedule()`:
```java
private void submitBin(Bin bin) {
  List<String> ids = bin.getOperationIds();
  operationsRepo.cancelDuplicatePendingBatch(bin.getOperationType().toDb(), ids);
  int claimedCount = operationsRepo.markSchedulingBatch(ids, Instant.now());
  if (claimedCount == 0) { /* skip */ return; }

  Optional<String> jobId = bin.schedule(jobsClient, resultsEndpoint);
  if (jobId.isPresent()) {
    operationsRepo.markScheduledBatch(ids, jobId.get());
  } else {
    operationsRepo.markPendingBatch(ids);
  }
}
```

Commit 9ad8861a.

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/client/JobsServiceClient.java
```diff
@@ -0,0 +1,80 @@
...
+      ObjectNode jobConf = body.putObject("jobConf");
+      jobConf.put("jobType", jobType);
+
+      ArrayNode args = jobConf.putArray("args");
+      args.add("--tableNames");
+      args.add(String.join(",", tableNames));
+      args.add("--operationIds");
+      args.add(String.join(",", operationIds));
+      args.add("--resultsEndpoint");
+      args.add(resultsEndpoint);
```
mkuchenbecker, line 61: This is the key change, rather than single table name it needs to take in a list of table names, their associated IDs, and a callback endpoint.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/FileCountBinPacker.java
```diff
@@ -0,0 +1,84 @@
...
+            .map(SchedulingCandidate::getOperation)
+            .sorted(
+                Comparator.comparingLong((TableOperation op) -> costByOperationId.get(op.getId()))
+                    .reversed())
+            .collect(Collectors.toList());
+
+    List<List<TableOperation>> binContents = new ArrayList<>();
+    List<Long> binTotals = new ArrayList<>();
+
+    for (TableOperation op : sorted) {
```
mkuchenbecker: don't use for loops to iterate over a collection vs functional pattern.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,124 @@
...
+ * For one operation type per call, reads PENDING rows, looks up per-table stats, dispatches to the
+ * registered {@link BinPacker}, and submits one Spark job per returned {@link Bin}. The {@link
+ * com.linkedin.openhouse.scheduler.SchedulerApplication}'s CommandLineRunner loops over the
+ * registered packers and invokes {@code schedule(opType)} for each.
+ */
+@Slf4j
+@Component
+@RequiredArgsConstructor
+public class SchedulerRunner {
+
```
mkuchenbecker: blank line

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,124 @@
...
+  private final TableOperationsRepository operationsRepo;
+  private final TableStatsRepository statsRepo;
+  private final JobsServiceClient jobsClient;
+  private final Map<OperationType, BinPacker> binPackers;
+
+  @Value("${scheduler.results-endpoint}")
+  private String resultsEndpoint;
```
mkuchenbecker: final

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,124 @@
...
+   * Schedule PENDING operations for {@code operationType}, optionally scoped to a single database
+   * or table name.
+   */
+  @Transactional
+  public void schedule(
+      OperationType operationType, Optional<String> databaseName, Optional<String> tableName) {
+
+    BinPacker packer = binPackers.get(operationType);
+    if (packer == null) {
+      log.warn("No BinPacker registered for operation type {}; skipping", operationType);
```
mkuchenbecker: throw or default to a SingletonBinPacker. Singleton / single can be the default. If you can't find one specified for OFD, we should throw.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/FileCountBinPacker.java
mkuchenbecker: > don't use for loops to iterate over a collection vs functional pattern.

Converted both for-loops in `FileCountBinPacker.pack`. Cost-map building is now a stream collect; the bin-assignment search is `IntStream.range().filter().findFirst()`; the outer iteration is `sorted.forEach(...)`.

```java
// FileCountBinPacker.pack
Map<String, Long> costByOperationId =
    pending.stream()
        .collect(Collectors.toMap(c -> c.getOperation().getId(), c -> cost(c.getStats())));

List<TableOperation> sorted =
    pending.stream()
        .map(SchedulingCandidate::getOperation)
        .sorted(
            Comparator.comparingLong((TableOperation op) -> costByOperationId.get(op.getId()))
                .reversed())
        .collect(Collectors.toList());

List<List<TableOperation>> binContents = new ArrayList<>();
List<Long> binTotals = new ArrayList<>();
sorted.forEach(op -> placeInBin(op, costByOperationId.get(op.getId()), binContents, binTotals));

return binContents.stream()
    .map(ops -> new Bin(operationType, ops))
    .collect(Collectors.toList());
```

```java
// placeInBin — bin-assignment search via IntStream
private void placeInBin(
    TableOperation op, long cost, List<List<TableOperation>> bins, List<Long> totals) {
  OptionalInt placed =
      IntStream.range(0, bins.size())
          .filter(i -> totals.get(i) + cost <= maxFilesPerBin || totals.get(i) == 0)
          .findFirst();
  if (placed.isPresent()) {
    int idx = placed.getAsInt();
    bins.get(idx).add(op);
    totals.set(idx, totals.get(idx) + cost);
  } else {
    List<TableOperation> newBin = new ArrayList<>();
    newBin.add(op);
    bins.add(newBin);
    totals.add(cost);
  }
}
```

The `bins`/`totals` lists themselves still mutate across placements because first-fit-descending is inherently stateful (each placement depends on prior running totals), but no `for` keyword remains. Commit 9e5fbae8.

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,124 @@
...
+      OperationType operationType, Optional<String> databaseName, Optional<String> tableName) {
+
+    BinPacker packer = binPackers.get(operationType);
+    if (packer == null) {
+      log.warn("No BinPacker registered for operation type {}; skipping", operationType);
+      return;
+    }
+
+    List<TableOperationsRow> pendingRows =
+        operationsRepo.find(
```
mkuchenbecker: So for every op we will schedule, we will have one operation to get current stats. This data could be denormalized into table operation to avoid the lookup at the cost of larger Operations object with staler data. I would add trade off in the comment here.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/FileCountBinPacker.java
```diff
@@ -0,0 +1,86 @@
...
+    List<List<TableOperation>> binContents = new ArrayList<>();
+    List<Long> binTotals = new ArrayList<>();
+    sorted.forEach(op -> placeInBin(op, costByOperationId.get(op.getId()), binContents, binTotals));
+
+    return binContents.stream()
+        .map(ops -> new Bin(operationType, ops))
+        .collect(Collectors.toList());
+  }
+
+  private void placeInBin(
```
mkuchenbecker: This function can be defined ineline in sorted.forEach rather than a one-off private helper.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/FileCountBinPacker.java
```diff
@@ -0,0 +1,86 @@
...
+    } else {
+      List<TableOperation> newBin = new ArrayList<>();
+      newBin.add(op);
+      bins.add(newBin);
+      totals.add(cost);
+    }
+  }
+
+  private static long cost(TableStats stats) {
+    if (stats == null || stats.getSnapshot() == null) {
```
mkuchenbecker: These should never be null, no?

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
mkuchenbecker: > blank line

Dropped the stray blank line between the class brace and the field block:

```java
public class SchedulerRunner {
  private final TableOperationsRepository operationsRepo;
  private final TableStatsRepository statsRepo;
  ...
```

Commit 7dd1e972.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
mkuchenbecker: > final

Done. `resultsEndpoint` is now a final field, populated via constructor injection. Dropped Lombok's `@RequiredArgsConstructor` so the constructor can carry the `@Value` parameter:

```java
public class SchedulerRunner {
  private final TableOperationsRepository operationsRepo;
  private final TableStatsRepository statsRepo;
  private final JobsServiceClient jobsClient;
  private final Map<OperationType, BinPacker> binPackers;
  private final String resultsEndpoint;

  public SchedulerRunner(
      TableOperationsRepository operationsRepo,
      TableStatsRepository statsRepo,
      JobsServiceClient jobsClient,
      Map<OperationType, BinPacker> binPackers,
      @Value("${scheduler.results-endpoint}") String resultsEndpoint) {
    this.operationsRepo = operationsRepo;
    this.statsRepo = statsRepo;
    this.jobsClient = jobsClient;
    this.binPackers = binPackers;
    this.resultsEndpoint = resultsEndpoint;
  }
  ...
}
```

Tests updated to pass the endpoint through the constructor instead of `ReflectionTestUtils.setField`. Commit 7dd1e972.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
mkuchenbecker: > throw or default to a SingletonBinPacker. Singleton / single can be the default. If you can't find one specified for OFD, we should throw.

Went with **throw** for the missing-packer case — silent skip was masking a configuration error. `SingletonBinPacker` as a default-fallback is a sensible follow-up but I'd rather wire it explicitly in `SchedulerConfig` (per op type that wants single-op-per-bin) than make it the runner's implicit default.

```java
// SchedulerRunner.schedule
BinPacker packer = binPackers.get(operationType);
if (packer == null) {
  throw new IllegalStateException(
      "No BinPacker registered for operation type " + operationType);
}
```

For OFD this already throws — `SchedulerConfig` registers `FileCountBinPacker` so no fallback is needed. When a future op type lands without a configured packer, this will fail loud at the first scheduling pass. Commit 7dd1e972.

Happy to add `SingletonBinPacker` (one-op-per-bin no-op packer) and wire it for op types that don't need batching, but separately — let me know.

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
mkuchenbecker: > So for every op we will schedule, we will have one operation to get current stats. This data could be denormalized into table operation to avoid the lookup at the cost of larger Operations object with staler data. I would add trade off in the comment here.

Added the trade-off as a comment at the stats lookup site:

```java
// Tradeoff: we fetch fresh table_stats per scheduling cycle (one batched query) rather than
// denormalizing the relevant fields onto TableOperation. The denormalized alternative would
// remove the per-cycle lookup but widen the TableOperation row and serve staler data; the
// current shape favors smaller operations + freshness over fewer queries.
Set<String> uuids =
    pending.stream().map(TableOperation::getTableUuid).collect(Collectors.toSet());
Map<String, TableStats> statsByUuid =
    statsRepo.findAllById(uuids).stream()
        .collect(Collectors.toMap(TableStatsRow::getTableUuid, TableStats::fromRow));
```

Worth noting it's one batched `findAllById` per scheduling cycle, not per-op — but the broader trade-off (lookup vs. wider operation row) is now documented at the call site. Commit 7dd1e972.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/FileCountBinPacker.java
mkuchenbecker: > This function can be defined inline in sorted.forEach rather than a one-off private helper.

Inlined. `placeInBin` is gone; the body lives in the `sorted.forEach` lambda:

```java
// FileCountBinPacker.pack
List<List<TableOperation>> binContents = new ArrayList<>();
List<Long> binTotals = new ArrayList<>();
sorted.forEach(
    op -> {
      long c = costByOperationId.get(op.getId());
      OptionalInt placed =
          IntStream.range(0, binContents.size())
              .filter(i -> binTotals.get(i) + c <= maxFilesPerBin || binTotals.get(i) == 0)
              .findFirst();
      if (placed.isPresent()) {
        int idx = placed.getAsInt();
        binContents.get(idx).add(op);
        binTotals.set(idx, binTotals.get(idx) + c);
      } else {
        List<TableOperation> newBin = new ArrayList<>();
        newBin.add(op);
        binContents.add(newBin);
        binTotals.add(c);
      }
    });
```

Commit 7dd1e972.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/FileCountBinPacker.java
mkuchenbecker: > These should never be null, no?

Agreed. Made it a real invariant: `SchedulingCandidate.stats` is `@NonNull` and the scheduler filters at the boundary so candidates with no stats row never reach the packer.

```java
// SchedulingCandidate.java
@Value
public class SchedulingCandidate {
  @NonNull TableOperation operation;
  @NonNull TableStats stats;
}
```

```java
// SchedulerRunner.schedule — filter at the boundary
List<TableOperation> withStats =
    pending.stream()
        .filter(op -> statsByUuid.containsKey(op.getTableUuid()))
        .collect(Collectors.toList());
if (withStats.size() < pending.size()) {
  log.warn(
      "Skipped {} {} operations with no table_stats row",
      pending.size() - withStats.size(),
      operationType);
}
```

```java
// FileCountBinPacker — null checks gone
private static long cost(TableStats stats) {
  Long n = stats.getSnapshot().getNumCurrentFiles();
  return n != null ? n : 0L;
}
```

Kept the `n != null` guard for `getNumCurrentFiles()` since the underlying Long can be null for a freshly-written snapshot that hasn't been measured yet — treating that as zero is the same conservative default. If you want that to NPE too, easy to drop.

New test `schedule_opsWithoutStats_skipped` exercises the filter; `schedule_allOpsWithoutStats_noJobSubmitted` covers the empty-after-filter short-circuit. Commit 7dd1e972.

apps/optimizer-scheduler/src/main/java/com/linkedin/openhouse/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,155 @@
...
+    // Claim the rows in one batched UPDATE: PENDING → SCHEDULING.
+    int claimedCount = operationsRepo.markSchedulingBatch(ids, Instant.now());
+    if (claimedCount == 0) {
+      log.info("All rows in bin already claimed by another scheduler instance; skipping");
+      return;
+    }
+
+    Optional<String> jobId = bin.schedule(jobsClient, resultsEndpoint);
+    if (jobId.isPresent()) {
+      int updated = operationsRepo.markScheduledBatch(ids, jobId.get());
```
mkuchenbecker: BUG: If some ids are claimed, this will incorrectly mark them as scheduled. You need bin.markSchedulingBatch to either fail if they are not transactionally claimed OR return the ids that were actually claimed.

mkuchenbecker: > BUG: If some ids are claimed, this will incorrectly mark them as scheduled. You need bin.markSchedulingBatch to either fail if they are not transactionally claimed OR return the ids that were actually claimed.

Good catch. Went with **return the ids actually claimed**.

New repo method `findClaimedIds(ids, scheduledAt)` on opt-1 — pairs with `markSchedulingBatch(ids, scheduledAt)` and uses the same watermark to identify which rows this caller transitioned PENDING→SCHEDULING in this call:

```java
// TableOperationsRepository
@Query(
    "SELECT r.id FROM TableOperationsRow r "
        + "WHERE r.id IN :ids "
        + "AND r.status = com.linkedin.openhouse.optimizer.db.OperationStatus.SCHEDULING "
        + "AND r.scheduledAt = :scheduledAt")
List<String> findClaimedIds(
    @Param("ids") List<String> ids, @Param("scheduledAt") Instant scheduledAt);
```

`SchedulerRunner.submitBin` now subsets the bin to actually-claimed ops before launching:

```java
Instant claimedAt = Instant.now();
operationsRepo.markSchedulingBatch(ids, claimedAt);
List<String> claimedIds = operationsRepo.findClaimedIds(ids, claimedAt);
if (claimedIds.isEmpty()) {
  log.info("All rows in bin already claimed by another scheduler instance; skipping");
  return;
}
if (claimedIds.size() < ids.size()) {
  log.info(
      "Partial claim: {} of {} ops in bin claimed; launching job for claimed subset only",
      claimedIds.size(),
      ids.size());
}

Bin claimedBin = bin.subset(claimedIds);
Optional<String> jobId = claimedBin.schedule(jobsClient, resultsEndpoint);
if (jobId.isPresent()) {
  operationsRepo.markScheduledBatch(claimedIds, jobId.get());
} else {
  operationsRepo.markPendingBatch(claimedIds);
}
```

New `Bin.subset(Collection<String> keepIds)` returns a new Bin filtered to the kept ids. The job is launched only for the claimed subset, and `markScheduledBatch`/`markPendingBatch` use `claimedIds` only — so rows owned by another instance are left alone.

New integration test on the repository covers `findClaimedIds` (`TableOperationsRepositoryTest.findClaimedIds_returnsOnlyClaimedSubset` — verifies a pre-existing SCHEDULING row with a different `scheduledAt` is not returned). New unit test on the scheduler covers the partial-claim path (`SchedulerRunnerTest.schedule_partialClaim_launchesAndMarksOnlyClaimedSubset`).

Commits: 2b06c92e (opt-1 repo method, cascaded up) and c73d8cb0 (opt-4 fix).

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,167 @@
...
+
+    // Claim the rows in one batched UPDATE: PENDING → SCHEDULING. The UPDATE's row count is just
+    // an aggregate — to know *which* rows we own, re-query for SCHEDULING rows tagged with our
+    // scheduledAt watermark. Anything not in that subset belongs to another instance or was
+    // canceled, and must not be submitted or marked SCHEDULED.
+    Instant claimedAt = Instant.now();
+    operationsRepo.markSchedulingBatch(ids, claimedAt);
+    List<String> claimedIds = operationsRepo.findClaimedIds(ids, claimedAt);
+    if (claimedIds.isEmpty()) {
+      log.info("All rows in bin already claimed by another scheduler instance; skipping");
```
mkuchenbecker: In-practice this shouldn't happen with a single scheduler but any parallel schedulers could race.  On some rows still needing work, we schedule. We could error, or repack but it seems like we should just continue as its slightly less efficient but doesn't impact the correctness in that rare occurrence.

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,167 @@
...
+      return;
+    }
+    if (claimedIds.size() < ids.size()) {
+      log.info(
```
mkuchenbecker: Warning.

apps/optimizer/schedulerapp/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerApplication.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.domain.EntityScan;
+import org.springframework.context.annotation.Bean;
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+
+/** Entry point for the Optimizer Scheduler application. */
+@SpringBootApplication
+@EntityScan(basePackages = "com.linkedin.openhouse.optimizer.db")
+@EnableJpaRepositories(basePackages = "com.linkedin.openhouse.optimizer.repository")
+public class SchedulerApplication {
```
abhisheknath2011: Question: Should SchedulerApplication be a standalone Java application or a Spring Boot service?

I'd like to discuss the trade-offs before we commit to a service. My concern is that we may end up managing too many long-running services over time.

Service approach
- Pro: Can be invoked via API
- Con: Adds another always-on service to operate and monitor
- Con: In practice, the only trigger point is the Analyzer, so the API surface is largely unused

Java application approach (preferred)
- Analyzer creates bulk rows for analyzed records in a PENDING state and dynamically triggers the Scheduler
- Scheduler runs as a Kubernetes Pod or CronJob — scheduled, on-demand, or ad-hoc
- Execution is naturally trackable through pod/job lifecycle (status, logs, exit codes)
- Aligns with the current implementation direction and the existing code structure already fits this model cleanly

The Analyzer remains the single trigger point, we avoid the operational overhead of another service, and we still retain on-demand execution via manual Pod/Job runs. If we ever need an API trigger, we can wrap it later. Also the current implementation does not align with service specific implementation. The analyzer codebase has the bin packer implementation and we are  triggering scheduler more of a lib by passing required parameters. Let me know your thoughts.

mkuchenbecker: I want this as a k8s cron java app in the short term. The endpoint is only needed for adhoc calls from customer or other services. I think its good to expose an endpoint, but its not strictly needed and something that occationally recieves calls from a user has a way different SLA than one that needs to be up 24/7. 

PR has been structured so that the interface (API / Application ) is decoupled from the logic and its a two-way door. I would not implement the API endpoint until its actually needed, but e.g. we might expose an analyze/scheduel command to users in the future.

mkuchenbecker: discussed on the other PR, this will be a cron.

abhisheknath2011: Yes we can run this as k8s cron java app. As this would be more of Spring batch style application using we can we implement something like below (using `CommandLineRunner `) so that we can have clean shutdown when completed. Another option is to keep as plain java application. This is applicable for AnalyzerApplication as well as the plan is to run as cron job to begin with.


```suggestion
@SpringBootApplication
public class SchedulerApplication implements CommandLineRunner, ExitCodeGenerator {

  private int exitCode = 0;

  public static void main(String[] args) {
    System.exit(SpringApplication.exit(SpringApplication.run(SchedulerApplication.class, args)));
  }

  @Override
  public void run(String... args) {
    try {
      log.info("Scheduler starting");
      // code to execute
      log.info("Scheduler completed successfully");
    } catch (Exception e) {
      log.error("Scheduler failed", e);
      exitCode = 1;
    }
  }

  @Override
  public int getExitCode() {
    return exitCode;
  }
}

Config: application.properties:
spring.main.web-application-type=none
spring.main.banner-mode=off

What this provides:
- SpringApplication.exit(...) triggers graceful context shutdown — @PreDestroy hooks run, datasources close, etc.
- ExitCodeGenerator propagates failure to the K8s pod exit code, so the CronJob status reflects reality
- No embedded web server, no Tomcat port binding
- try/catch ensures the JVM exits even if the service throws
```

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/config/SchedulerConfig.java
```diff
@@ -0,0 +1,46 @@
...
+@Configuration
+public class SchedulerConfig {
+
+  @Value("${jobs.base-uri}")
+  private String jobsBaseUri;
+
+  @Value("${scheduler.cluster-id}")
+  private String clusterId;
+
+  @Value("${scheduler.ofd.max-files-per-bin}")
```
abhisheknath2011: Can we provide common prefix like `optimizer.scheduler.ofd.max-files-per-bin`? Same comment for other properties.

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,248 @@
...
+  private final JobsServiceClient jobsClient;
+  private final Map<OperationTypeDto, BinPacker> binPackers;
+  private final String resultsEndpoint;
+
+  public SchedulerRunner(
+      TableOperationsRepository operationsRepo,
+      TableStatsRepository statsRepo,
+      JobsServiceClient jobsClient,
+      Map<OperationTypeDto, BinPacker> binPackers,
+      @Value("${scheduler.results-endpoint}") String resultsEndpoint) {
```
abhisheknath2011:
```suggestion
      @Value("${optimizer.scheduler.results-endpoint}") String resultsEndpoint) {
```

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,248 @@
...
+    Instant claimedAt = Instant.now();
+    operationsRepo.updateBatch(
+        ids,
+        OperationStatus.PENDING,
+        OperationStatus.SCHEDULING,
+        Optional.of(claimedAt),
+        Optional.empty());
+    // Unpaged: the result set is already bounded by ids.size() (the bin we just claimed); no
+    // need to cap it further.
+    List<String> claimedIds =
```
abhisheknath2011: As there will be multiple scheduler instances going forward, we need to ensure that one operation for a table is picked up by only one scheduler instance. Another thing is can we run multiple operations on a single table and those are picked up by different scheduler? Also distribution DBs across different scheduler instances. I guess we can think of these scenarios and adjust the code accordingly going forward. For the initial version this should be fine, just need to ensure if there are no race conditions.

mkuchenbecker: Race conditions are handled by claiming today; even if we shard I don't want overlapping runs or retries to have an issue.

1) Scheduler gets N Operations
2) Scheduler attempts to claim each.
3) Scheduler actually schedules those it was able to claim. 

(2) will fail if another scheduler races. 

Sharding will enable parallelism and we would want them working on disjoints sets of operations, but I think we still need state-machine protection against duplicate job submission, in-particular the operations that create new snapshots and can cause txn conflicts

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/config/SchedulerConfig.java
mkuchenbecker: done

apps/optimizer/schedulerapp/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerApplication.java
mkuchenbecker: updated.

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/Bin.java
```diff
@@ -0,0 +1,61 @@
...
+import lombok.RequiredArgsConstructor;
+
+/**
+ * A set of operations the scheduler will submit together as a single Spark job. A bin owns its own
+ * launch — callers ask it to schedule itself and react to the returned job id. The surrounding
+ * status-update machinery (claim, mark-scheduled, revert-to-pending) lives in the scheduler because
+ * it is shared across all bins regardless of operation type.
+ */
+@RequiredArgsConstructor
+public class Bin {
```
abhisheknath2011: I am introducing BinItem as part of this PR - https://github.com/linkedin/openhouse/pull/599/changes#diff-5e026d8449953ed5f2853964c9fc6427827dac24fa3b9dba10318fb6618fb703 to represent data a granular level. I will rebase my PR once this PR is merged.

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/Bin.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @Getter private final OperationTypeDto operationType;
+  @Getter private final List<TableOperationDto> operations;
```
abhisheknath2011: Shall we keep bin packing generic as common utility instead of referencing internal models and operations?

abhisheknath2011: That should give more flexibility and we should be able to integrate well with the optimizer flow as well existing scheduler flow. I am planning to leverage as a common lib as used in this PR - https://github.com/linkedin/openhouse/pull/604/changes#diff-bd8bddafa29e6a0d0dcc04642cf89b969c4890f53efa9828826e51c25f970a7d.

mkuchenbecker:
I thought though this deeply:

Scheduler operation is as follows:
Scheduler s = new Scheduler().regiserOperation(Orphaned_file_deletion, ...).registerOperation(SE, ...)
s.schedule(Orphaned_File_deletion)
    
 We register a FirstFitBinPacker<TotalFilesBinItem> bin packer for OFD or any operation that scales based on files size. There might be a singleton bin packer for operations we can't safely pack. Another could be based on table size, or take database into account. 
 
 The Bin Packer interface just cares that you can convert a TableStat to BinItem. By making a FirstFitBinPacker<TotalFilesBinItem> we are delegating the conversion to the generic type. The pattern is basically "A bin item is valid if if can be constructed from stats" and we can evolve that definition over time. 
 
 cc @abhisheknath2011

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/Bin.java
mkuchenbecker: https://github.com/linkedin/openhouse/pull/534#issuecomment-4607018553

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/Bin.java
mkuchenbecker: https://github.com/linkedin/openhouse/pull/534#issuecomment-4607018553

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,250 @@
...
+            .collect(Collectors.toList());
+    List<String> toCancel =
+        byTableUuid.values().stream()
+            .filter(rows -> rows.size() > 1)
+            .flatMap(rows -> rows.stream().sorted(oldestFirst).skip(1))
+            .map(TableOperationsRow::getId)
+            .collect(Collectors.toList());
+
+    if (!toCancel.isEmpty()) {
+      int cancelled = operationsRepo.cancel(toCancel);
```
mkuchenbecker: make the repo handle an empty list. 

Then we can jsut do:

operationsRepo.cancel(byTableUuid.values().stream()
            .filter(rows -> rows.size() > 1)
            .flatMap(rows -> rows.stream().sorted(oldestFirst).skip(1))
            .map(TableOperationsRow::getId)
            .collect(Collectors.toList()))

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,250 @@
...
+      log.warn("Cancelled {} duplicate PENDING rows", cancelled);
+    }
+
+    Set<String> uuidsToSchedule =
```
mkuchenbecker: why are we making this in two passes rather than making it directly?
byTableUuid.values().stream()
            .map(rows -> rows.stream().min(oldestFirst).orElseThrow())
            .map(TableOperationDto::fromRow)
            .map(TableOperationDto::getTableUuid)
            .collect(Collectors.toSet())

please update if no reason.

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
mkuchenbecker: Done in 26c525f3 — added a `default int cancel(List<String>)` wrapper that short-circuits on empty, then inlined the duplicate-id derivation directly into the `operationsRepo.cancel(...)` call, dropping the local + the `if (!toCancel.isEmpty())` guard.

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
mkuchenbecker: Done in 26c525f3 — collapsed all the way to `statsRepo.findAllById(byTableUuid.keySet())`. `byTableUuid.keySet()` is exactly the set of tableUuids we will schedule (one survivor per group), so the prior `uuidsToSchedule` derivation from `toSchedule` is gone — no second pass, no local.

services/optimizer/scheduler/src/main/java/com/linkedin/openhouse/optimizer/scheduler/SchedulerRunner.java
```diff
@@ -0,0 +1,251 @@
...
+        .collect(
+            Collectors.collectingAndThen(
+                Collectors.toList(),
+                ids -> {
+                  if (!ids.isEmpty()) {
+                    operationsRepo.cancel(ids);
+                  }
+                  return null;
+                }));
+
```
mkuchenbecker:
```suggestion
                }));
    // Read stats, bin pack, and schedule the operations per bin.
```

libs/optimizer/binpack/src/main/java/com/linkedin/openhouse/optimizer/binpack/Bin.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * One scheduling unit: the operation type the bin will run as, and the items the scheduler will
+ * claim, narrow to claimed, and launch a single Spark job for. Pure data — the scheduler reads from
+ * a bin to do the work; the bin does no IO itself.
+ */
+@AllArgsConstructor
+@Getter
+@ToString
+public class Bin {
```
abhisheknath2011, line 17: Thanks for making bin packing generic. I think we can't have services dependency on apps module. As the bin packing utilities are common it would helpful if we could  extract them to the common module (such as optimizer/common) outside of `apps` and `services` module so that they can be reused across different modules.

libs/optimizer/binpack/src/main/java/com/linkedin/openhouse/optimizer/binpack/FirstFitBinPacker.java
```diff
@@ -0,0 +1,92 @@
...
+  private final int maxItemsPerBin;
+
+  public FirstFitBinPacker(Supplier<T> binItemSupplier, long maxWeightPerBin, int maxItemsPerBin) {
+    this.binItemSupplier = binItemSupplier;
+    this.maxWeightPerBin = maxWeightPerBin;
+    this.maxItemsPerBin = maxItemsPerBin;
+  }
+
+  @Override
+  public List<List<BinItem>> pack(
```
abhisheknath2011, line 42: I think we can have the common bin packing classes (like Bin, BinItem, base classes etc.) in common module and can have specific implementation for pack in the services module under optimizer. That as we can plugin specific codebase.

libs/optimizer/binpack/src/main/java/com/linkedin/openhouse/optimizer/binpack/FirstFitBinPacker.java
```diff
@@ -0,0 +1,92 @@
...
+      List<TableOperationDto> operations, Map<String, TableStatsDto> statsByTableUuid) {
```
abhisheknath2011, line 43: All DTOs can have baseDTOs and that way it would be easier to extract in common module.

libs/optimizer/binpack/src/main/java/com/linkedin/openhouse/optimizer/binpack/FirstFitBinPacker.java
abhisheknath2011: I think I can put up a PR on this. We can merge this PR first as the diff will be more.

apps/optimizer/schedulerapp/build.gradle
```diff
@@ -0,0 +1,14 @@
+plugins {
+  id 'openhouse.springboot-ext-conventions'
+  id 'org.springframework.boot' version '2.7.8'
```
abhisheknath2011, line 4: We need to add id 'openhouse.maven-publish' here.

abhisheknath2011 (approved):
Thanks @mkuchenbecker for addressing the comments!

PR #541, opened 2026-04-12
Surface corrupt Iceberg metadata as 422 instead of masking as table-not-found

dushyantk1509:
## Summary
A table had corrupt metadata — current-schema-id=6 but schema 6 was missing from the schemas array. Both Spark and Trino reported "Table does not exist", making the issue impossible to diagnose without digging through tables-service logs.

The reason behind Spark and Trino reporting "Table does not exist" is - 
* Iceberg's `TableMetadataParser.fromJson()` throws `IllegalArgumentException` when parsing corrupt metadata.
* `OpenHouseExceptionHandler` maps `IllegalArgumentException` → HTTP 400 (Bad Request).
* Catalog client (`OpenHouseTableOperations.doRefresh()`) silently swallows 400 responses via `.onErrorResume(BadRequest.class, e -> Mono.empty())`, converting them to `NoSuchTableException` → "Table does not exist".

There are couple of issues in this code path
1. Mapping of `IllegalArgumentException` → HTTP 400 (Bad Request) is done assuming that user side validation will be mapped to `IllegalArgumentException`. It's done correctly in certain places like like `OpenHouseInternalTableOperations.rebuildSortOrder` but seems messed up in certain places like `StorageManager` where it indicates server side issue. This is something can be corrected in follow up PR(s).
2. Catalog client silently swallows 400 responses doesn't seem correct as bad request doesn't mean table doesn't exist. Such issues should be surfaced to users. This is something can be corrected in follow up PR(s).

Keeping the scope only to surfacing iceberg metadata corruption issue, this PR adds `InvalidTableMetadataException` (extends `UnprocessableEntityException` → HTTP 422) and catch metadata corruption exceptions in `OpenHouseInternalTableOperations.refreshMetadata()`. The catch covers the full range of Iceberg metadata failures from `refreshFromMetadataLocation()`:
* `IllegalArgumentException`: corrupt content — malformed JSON fields, missing schema IDs, invalid partition specs, sort orders, snapshots (~60+ throw sites across TableMetadataParser, SchemaParser, PartitionSpecParser, SortOrderParser, SnapshotParser, JsonUtil, and TableMetadata constructor)               
* `IllegalStateException`: UUID mismatch — refreshed metadata has a different table UUID than expected.   
* `NotFoundException`: metadata file missing from storage — HouseTable record exists but the metadata file it points to has been deleted (dangling pointer)                                                                
* `ValidationException`: snapshot/partition spec inconsistencies — e.g. snapshot sequence number exceeds last sequence number.

`RuntimeIOException` and `UncheckedIOException` are intentionally not caught as they also represent transient infrastructure failures (HDFS down, network timeout), not necessarily metadata corruption. So, 500 server error is fine.

## Changes

- [ ] Internal API Changes
- [X] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
@@ -12,6 +12,7 @@ import com.linkedin.openhouse.cluster.storage.Storage;
 import com.linkedin.openhouse.cluster.storage.StorageClient;
 import com.linkedin.openhouse.cluster.storage.hdfs.HdfsStorageClient;
 import com.linkedin.openhouse.cluster.storage.local.LocalStorageClient;
+import com.linkedin.openhouse.common.exception.InvalidTableMetadataException;
 import com.linkedin.openhouse.internal.catalog.exception.InvalidIcebergSnapshotException;
 import com.linkedin.openhouse.internal.catalog.fileio.FileIOManager;
 import com.linkedin.openhouse.internal.catalog.mapper.HouseTableMapper;
@@ -61,6 +62,7 @@ import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.exceptions.BadRequestException;
 import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.exceptions.CommitStateUnknownException;
+import org.apache.iceberg.exceptions.NotFoundException;
 import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.expressions.Expressions;
 import org.apache.iceberg.expressions.Term;
@@ -143,6 +145,17 @@ public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperatio
           "refreshMetadata from location {} succeeded, took {} ms",
           metadataLoc,
           System.currentTimeMillis() - startTime);
+    } catch (IllegalArgumentException
+        | IllegalStateException
+        | NotFoundException
+        | ValidationException e) {
+      log.error(
+          "refreshMetadata from location {} failed after {} ms",
+          metadataLoc,
+          System.currentTimeMillis() - startTime,
+          e);
+      throw new InvalidTableMetadataException(
+          tableIdentifier.namespace().toString(), tableIdentifier.name(), e.getMessage(), e);
     } catch (Exception e) {
       log.error(
           "refreshMetadata from location {} failed after {} ms",
--- a/iceberg/openhouse/internalcatalog/src/test/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperationsTest.java
+++ b/iceberg/openhouse/internalcatalog/src/test/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperationsTest.java
@@ -8,6 +8,7 @@ import com.linkedin.openhouse.cluster.storage.StorageClient;
 import com.linkedin.openhouse.cluster.storage.StorageType;
 import com.linkedin.openhouse.cluster.storage.local.LocalStorage;
 import com.linkedin.openhouse.cluster.storage.local.LocalStorageClient;
+import com.linkedin.openhouse.common.exception.InvalidTableMetadataException;
 import com.linkedin.openhouse.internal.catalog.fileio.FileIOManager;
 import com.linkedin.openhouse.internal.catalog.mapper.HouseTableMapper;
 import com.linkedin.openhouse.internal.catalog.model.HouseTable;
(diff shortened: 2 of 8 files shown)


iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -143,6 +145,17 @@ protected void refreshMetadata(final String metadataLoc) {
...
+    } catch (IllegalArgumentException
+        | IllegalStateException
+        | NotFoundException
+        | ValidationException e) {
+      log.error(
+          "refreshMetadata from location {} failed after {} ms",
+          metadataLoc,
+          System.currentTimeMillis() - startTime,
+          e);
+      throw new InvalidTableMetadataException(
```
sumedhsakdeo, line 157: It seems not intuitive to have user specified invalid or illegal argument throw an invalid table metadata exception

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -143,6 +145,17 @@ protected void refreshMetadata(final String metadataLoc) {
           "refreshMetadata from location {} succeeded, took {} ms",
           metadataLoc,
           System.currentTimeMillis() - startTime);
+    } catch (IllegalArgumentException
+        | IllegalStateException
```
sumedhsakdeo, line 149: What is the iceberg client side behavior you expect? Cleanup the written commit or leave it there like how it is handled in CUSE?

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
dushyantk1509: This is not an user specified invalid or illegal argument. Here, iceberg is throwing `IllegalArgumentException` because of corrupted metadata. More specifically, the example I've see is - json file is referring to current-schema-id=6 which doesn't exist.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
dushyantk1509: I just looked at the read path while making this change but in the write path - commit step, refresh is done before commit. So, if refreshMetadata() fails, it fails before any new metadata is written — nothing to clean up.

maluchari (commented):
422 implies request was not processible. I am inclined towards 500 for this usecase. 500 does not promise transience - IMO, it means server failed to fulfill a valid request. It could cause retries but I would also say that implies retry for 500 errors should be revisited. This is a server side issue and hence we should not categorize it as a client class error.

dushyantk1509:
> 422 implies request was not processible. I am inclined towards 500 for this usecase. 500 does not promise transience - IMO, it means server failed to fulfill a valid request. It could cause retries but I would also say that implies retry for 500 errors should be revisited. This is a server side issue and hence we should not categorize it as a client class error.

yes, I agree it's not a client issue. We can go with 500 with clear error message as it falls under a generic server side issue.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -143,6 +145,17 @@ protected void refreshMetadata(final String metadataLoc) {
...
+        | NotFoundException
```
abhisheknath2011, line 150: We could separate out NotFoundException and ValidationException to clearly categorize them as 4xx error.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
abhisheknath2011, line 157: This may not be right exception as we are considering handling 4 types of exception within the catch block.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
dushyantk1509: In this context, both `NotFoundException` and `ValidationException` are server-side data integrity issues, not client errors:
1. `NotFoundException`: the HouseTable record exists but the metadata file it points to is missing from HDFS — a dangling pointer.
2. `ValidationException`: snapshot sequence number inconsistency in stored metadata.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
dushyantk1509: Seems related to above comment.

dushyantk1509:
@sumedhsakdeo / @maluchari / @abhisheknath2011 gentle ping on this.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -143,6 +145,17 @@ protected void refreshMetadata(final String metadataLoc) {
...
+        | ValidationException e) {
+      log.error(
+          "refreshMetadata from location {} failed after {} ms",
```
abhisheknath2011, line 153: Should we specifically mention reason as invalid/corrupted table metadata in the error log? Would be helpful from tracking/querying future occurances.

dushyantk1509: `InvalidTableMetadataException` has this message.

sumedhsakdeo (commented):
Thanks @dushyantk1509 , can you add a e2e table or test that maps a server side exception to client side exception? 
For example, if we raise IllegalArgumentException from the server side, what it translates into on client side (i.e. OH Catalog Client and Iceberg).

The reason for this e2e test is that this behavior has been a bit sensitive in the past, where changes made independently on server side have led to Iceberg deleting committed files on the client side.

I think your PR here is a good checkpoint for us to build stronger guardrails. It does increase the scope but I think I would like to contain it. If you want to do e2e test exception map between client and server in a separate PR, then I am ok with that too.

dushyantk1509:
> Thanks @dushyantk1509 , can you add a e2e table or test that maps a server side exception to client side exception? For example, if we raise IllegalArgumentException from the server side, what it translates into on client side (i.e. OH Catalog Client and Iceberg).
> 
> The reason for this e2e test is that this behavior has been a bit sensitive in the past, where changes made independently on server side have led to Iceberg deleting committed files on the client side.
> 
> I think your PR here is a good checkpoint for us to build stronger guardrails. It does increase the scope but I think I would like to contain it. If you want to do e2e test exception map between client and server in a separate PR, then I am ok with that too.

I've added an end-2-end unit test covering the server-to-client exception mapping for `InvalidTableMetadataException` + validating behaviour of not removing committed files for both read and write paths.

Added unit tests in `DoRefreshTest` to documents client-side behavior for HTTP status codes on the refresh path. The existing TablesControllerTest already tests the server-side mapping (InvalidTableMetadataException → 500), and DoCommitTest already tests that 500 → CommitStateUnknownException (no cleanup) on the commit path. 

I believe this covers the scope of this PR extensively. Thank you for the feedback.

Created another PR for broader exception mappings - https://github.com/linkedin/openhouse/pull/548.

sumedhsakdeo (approved):
lgtm

abhisheknath2011 (approved):
Thanks for the fix and useful tests!

PR #557, opened 2026-04-23
Consolidate single-level namespace validation in OpenHouseInternalCatalog

cbb330:
## Summary

Follow-up to #538 ([review thread](https://github.com/linkedin/openhouse/pull/538)). Consolidates OpenHouse's namespace-shape rule — currently asserted in four places inside `OpenHouseInternalCatalog` — into a single helper `NamespaceUtil` in `:services:common`. Behavior and error messages are unchanged.

The previous PR added an `isValidIdentifier(...)` override that, like the existing `listTables` / paginated `listTables` / `searchSoftDeletedTables` checks, encoded OpenHouse's namespace-shape constraint inline. Reviewer asked to route those checks through one place. This PR does that for the server-side catalog. Client-side `OpenHouseCatalog` has the same pattern in 7 sites; that can be a separate follow-up if desired (kept out of scope here to avoid coupling the server module's `:services:common` helper across the wire to the thin Java client runtime).

### Helper API

Two predicates on `Namespace`. They are intentionally separate concepts, not stricter/looser flavors of the same rule — the names reflect the namespace's *role*, not its *shape*, so a future change to the contract (e.g. supporting `catalog.database.table`) requires editing only the helper bodies, not the call-site names.

- `NamespaceUtil.isTableNamespace(Namespace)` — predicate, true when the namespace can host an OpenHouse base table. Used by `isValidIdentifier(...)`. Empty namespace is **not** a table namespace because there is no database under which to place a table.
- `NamespaceUtil.validateOperationNamespace(Namespace)` — throws `ValidationException` when the namespace cannot be used as the argument to a database-scoped catalog operation. Used by `listTables` / paginated `listTables` / `searchSoftDeletedTables`. Empty namespace **is** accepted because callers use it as a sentinel for "across all databases" (e.g. `listTables(Namespace.empty())`).

### Sites refactored in `OpenHouseInternalCatalog`

- `isValidIdentifier(TableIdentifier)` → `NamespaceUtil.isTableNamespace(...)`
- `listTables(Namespace)` → `NamespaceUtil.validateOperationNamespace(...)`
- `listTables(Namespace, Pageable)` → `NamespaceUtil.validateOperationNamespace(...)`
- `searchSoftDeletedTables(Namespace, String, Pageable)` → `NamespaceUtil.validateOperationNamespace(...)`

The previously inlined exception type (`ValidationException`) and message string (`"Input namespace has more than one levels " + String.join(".", namespace.levels())`) are preserved exactly inside the helper.

### Why `:services:common`

`OpenHouseInternalCatalog` already depends on `:services:common`, and `:services:common` already pulls in the Iceberg conventions, so `Namespace` and `ValidationException` are available there. The helper is `final` with a private constructor, mirroring the existing `PageableUtil` style.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [x] Refactoring
- [ ] Documentation
- [x] Tests

Refactoring:

- New `com.linkedin.openhouse.common.utils.NamespaceUtil` in `:services:common` exposing `isTableNamespace(Namespace)` and `validateOperationNamespace(Namespace)`. Identical exception type and identical error message string as the previous inline checks.
- Four call sites in `OpenHouseInternalCatalog` now delegate to the helper. Unused `org.apache.iceberg.exceptions.ValidationException` import removed.

Tests:

- New `NamespaceUtilTest` covering: `isTableNamespace` for null / empty / single-level / multi-level inputs, and `validateOperationNamespace` for empty (allowed), single-level (allowed), and multi-level (rejected with the exact pre-existing error message text — so any accidental wording change in a future edit fails loudly).
- Existing `OpenHouseInternalCatalogTest` continues to pass — it exercises `isValidIdentifier` end-to-end and is the regression guard for the table-namespace predicate.

## Testing Done

Added new tests for the changes made:

- `JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew :services:common:build :iceberg:openhouse:internalcatalog:build` — BUILD SUCCESSFUL. All existing tests pass, including `OpenHouseInternalCatalogTest` from #538 (which exercises `isValidIdentifier`) and `HouseTableRepositoryImplTest` (which exercises the list/search paths whose validation block was extracted).
- `JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew :services:common:spotlessCheck :iceberg:openhouse:internalcatalog:spotlessCheck` — clean.
- New `NamespaceUtilTest` runs as part of `:services:common:build`.

Testing gaps / risks:

- I did not run the full repo build (only the two modules touched). Downstream modules consume `OpenHouseInternalCatalog` indirectly via Spring wiring; behavior is byte-for-byte identical (same exception type, same message), so the blast radius should be limited to compile-only impact, but a CI run is the authoritative signal.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
@@ -79,7 +79,7 @@ public class OpenHouseInternalCatalog extends BaseMetastoreCatalog {
 
   @Override
   protected boolean isValidIdentifier(TableIdentifier tableIdentifier) {
-    return tableIdentifier != null && tableIdentifier.namespace().levels().length == 1;
+    return tableIdentifier != null && NamespaceUtil.isSingleLevel(tableIdentifier.namespace());
   }
 
   @Override
@@ -94,7 +94,7 @@ public class OpenHouseInternalCatalog extends BaseMetastoreCatalog {
 
   @Override
   public List<TableIdentifier> listTables(Namespace namespace) {
-    NamespaceUtil.checkSingleLevelNamespace(namespace);
+    NamespaceUtil.checkAtMostSingleLevelNamespace(namespace);
     // TODO: Implement SupportsNamespace interface and listNamespaces() method to remove this
     //  branch. This is anti-pattern and only a temporary solution.
     if (namespace.isEmpty()) {
@@ -108,7 +108,7 @@ public class OpenHouseInternalCatalog extends BaseMetastoreCatalog {
   }
 
   public Page<TableIdentifier> listTables(Namespace namespace, Pageable pageable) {
-    NamespaceUtil.checkSingleLevelNamespace(namespace);
+    NamespaceUtil.checkAtMostSingleLevelNamespace(namespace);
     if (namespace.isEmpty()) {
       return houseTableRepository
           .findAll(pageable)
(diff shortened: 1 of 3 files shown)


abhisheknath2011 (approved):
Thanks for the quick PR.

PR #558, opened 2026-04-24
Normalize write schema casing to table casing at catalog layer (Make OH writes case-insensitive)

pandaamit91:
## Summary
Writers (DaliSpark, Spark SQL, Trino DML, Java Iceberg API) may submit column names with different casing than what the table stores (e.g. "id" vs "ID"). Because validateWriteSchema is case-sensitive, these writes were rejected with InvalidSchemaEvolutionException even when the field IDs matched, making case-insensitive writes impossible without changing the table's existing column names.

## Changes

- [X] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

Fix: in doUpdateSchemaIfNeeded, normalize the write schema's top-level column names to match the table schema's casing (matched by Iceberg field ID, not by name) before any comparison or storage. The table's existing casing is never mutated, and writers do not need to know or match the exact casing stored in the catalog.

Tables where two or more top-level columns share the same case-folded name (e.g. "id" and "ID") are excluded from normalization because the target column would be ambiguous, so writes to such tables must still use exact casing.

## Testing Done

- BaseIcebergSchemaValidatorTest: 10 unit tests covering hasCaseDuplicateFields and normalizeSchemaCasingToTable in isolation, including new-column passthrough and field attribute preservation.
- RepositoryTest (H2 Spring integration): 2 new tests exercising the full save() path through the Spring context:
  - testCaseInsensitiveWrite_succeeds_andPreservesTableCasing: creates a table with "ID", saves an update with "id", asserts no exception and that the stored schema still shows "ID".
  - testCaseInsensitiveWrite_blockedForCaseDuplicateTable: creates a table with case-duplicate columns, asserts that a mismatched-casing save still throws InvalidSchemaEvolutionException.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BaseIcebergSchemaValidator.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BaseIcebergSchemaValidator.java
@@ -2,6 +2,10 @@ package com.linkedin.openhouse.tables.repository.impl;
 
 import com.linkedin.openhouse.common.exception.InvalidSchemaEvolutionException;
 import com.linkedin.openhouse.tables.repository.SchemaValidator;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.relocated.com.google.common.collect.MapDifference;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
@@ -13,9 +17,63 @@ import org.springframework.stereotype.Component;
  * A base implementation which delegates the validation to iceberg library that is configured to: -
  * not allow to write optional values to a required field. - allow input schema to have different
  * ordering than table.
+ *
+ * <p>Also provides static helpers used by {@link OpenHouseInternalRepositoryImpl} to normalize a
+ * write schema's top-level column names to the casing already present in the table schema, enabling
+ * case-insensitive writes without mutating the table's existing column casing.
  */
 @Component
 public class BaseIcebergSchemaValidator implements SchemaValidator {
+
+  /**
+   * Returns true if {@code schema} has two or more top-level columns whose names differ only in
+   * case (e.g. "id" and "ID"). Such tables are excluded from case-insensitive write normalization
+   * because the target column would be ambiguous.
+   */
+  static boolean hasCaseDuplicateFields(Schema schema) {
+    Map<String, Integer> seen = new HashMap<>();
+    for (Types.NestedField field : schema.columns()) {
+      String lower = field.name().toLowerCase();
+      if (seen.containsKey(lower) && !seen.get(lower).equals(field.fieldId())) {
+        return true;
+      }
+      seen.put(lower, field.fieldId());
+    }
+    return false;
+  }
+
+  /**
+   * Renames top-level fields in {@code writeSchema} to use the casing from {@code tableSchema},
+   * matched by Iceberg field ID (not by name). Fields in {@code writeSchema} whose ID does not
+   * appear in {@code tableSchema} (genuinely new columns) are left unchanged.
+   *
+   * <p>This preserves the table's existing column casing when a writer submits columns with
+   * different casing (e.g. writer sends "id", table has "ID" → normalized to "ID").
+   */
+  static Schema normalizeSchemaCasingToTable(Schema writeSchema, Schema tableSchema) {
+    // Build fieldId → table column name map for O(1) lookup
+    Map<Integer, String> tableNameById = new HashMap<>();
+    for (Types.NestedField field : tableSchema.columns()) {
+      tableNameById.put(field.fieldId(), field.name());
+    }
+
+    List<Types.NestedField> normalizedFields = new ArrayList<>();
+    for (Types.NestedField field : writeSchema.columns()) {
+      String tableName = tableNameById.get(field.fieldId());
+      if (tableName != null && !tableName.equals(field.name())) {
+        // Existing column with different casing: rename to table casing, keep all other attributes
+        normalizedFields.add(
+            field.isOptional()
+                ? Types.NestedField.optional(field.fieldId(), tableName, field.type(), field.doc())
+                : Types.NestedField.required(
+                    field.fieldId(), tableName, field.type(), field.doc()));
+      } else {
+        normalizedFields.add(field);
+      }
+    }
+    return new Schema(writeSchema.schemaId(), normalizedFields, writeSchema.identifierFieldIds());
+  }
+
   @Override
   public void validateWriteSchema(Schema oldSchema, Schema newSchema, String tableUri)
       throws InvalidSchemaEvolutionException, IllegalArgumentException {
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BaseIcebergSchemaValidator.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BaseIcebergSchemaValidator.java
@@ -3,11 +3,8 @@ package com.linkedin.openhouse.tables.repository.impl;
 import com.linkedin.openhouse.common.exception.InvalidSchemaEvolutionException;
 import com.linkedin.openhouse.tables.repository.SchemaValidator;
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
-import java.util.Set;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.relocated.com.google.common.collect.MapDifference;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
@@ -28,57 +25,6 @@ import org.springframework.stereotype.Component;
  */
 @Component
 public class BaseIcebergSchemaValidator implements SchemaValidator {
-
-  /**
-   * Returns true if {@code schema} has, at any nesting depth, two sibling fields whose names differ
-   * only in case (e.g. {@code "id"} and {@code "ID"} inside the same struct). Such schemas are
-   * excluded from case-insensitive write normalization because the target field would be ambiguous.
-   *
-   * <p>Note: fields with the same name in <em>different</em> structs (e.g. {@code user.id} and
-   * {@code session.id}) are not considered duplicates — Iceberg's field-ID semantics treat them as
-   * independent fields.
-   */
-  static boolean hasCaseDuplicateFields(Schema schema) {
-    return Boolean.TRUE.equals(
-        TypeUtil.visit(
-            schema,
-            new TypeUtil.SchemaVisitor<Boolean>() {
-              @Override
-              public Boolean schema(Schema s, Boolean structResult) {
-                return structResult;
-              }
-
-              @Override
-              public Boolean struct(Types.StructType struct, List<Boolean> childResults) {
-                if (childResults.stream().anyMatch(Boolean.TRUE::equals)) return true;
-                Set<String> seen = new HashSet<>();
-                for (Types.NestedField f : struct.fields()) {
-                  if (!seen.add(f.name().toLowerCase(Locale.ROOT))) return true;
-                }
-                return false;
-              }
-
-              @Override
-              public Boolean field(Types.NestedField f, Boolean fieldResult) {
-                return fieldResult;
-              }
-
-              @Override
-              public Boolean list(Types.ListType l, Boolean elementResult) {
-                return elementResult;
-              }
-
-              @Override
-              public Boolean map(Types.MapType m, Boolean kRes, Boolean vRes) {
-                return Boolean.TRUE.equals(kRes) || Boolean.TRUE.equals(vRes);
-              }
-
-              @Override
-              public Boolean primitive(Type.PrimitiveType p) {
-                return false;
-              }
-            }));
-  }
 
   /**
    * Rewrites every field name in {@code writeSchema} — at top level <em>and</em> inside nested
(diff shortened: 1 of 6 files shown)


cbb330 (changes requested):
Previously some tests were run which show that writes with different casing already succeeds.

So I want to validate whats the existing behavior before applying a fix. to do that, can you reproduce the issue in a separate PR under spark e2e tests? similar to this:

https://github.com/cbb330/openhouse/blob/22baa29fdc6fc043d1b2f6d024d3b2341b1f3bec/integrations/spark/spark-3.1/openhouse-spark-itest/src/test/java/com/linkedin/openhouse/spark/e2e/dml/IncrementalReadTest.java

pandaamit91:
@cbb330 here is the PR as requested, documenting existing casing behavior for all write paths - https://github.com/linkedin/openhouse/pull/562

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BaseIcebergSchemaValidator.java
```diff
@@ -13,9 +17,63 @@
...
  */
 @Component
 public class BaseIcebergSchemaValidator implements SchemaValidator {
+
+  /**
+   * Returns true if {@code schema} has two or more top-level columns whose names differ only in
+   * case (e.g. "id" and "ID"). Such tables are excluded from case-insensitive write normalization
+   * because the target column would be ambiguous.
+   */
+  static boolean hasCaseDuplicateFields(Schema schema) {
```
cbb330: this was also added elsewhere. how is this duplicating what already exists ?

cbb330: looked deeper. the function exists in similar way in li-openhouse.

1. The two implementations should converge. They case-fold top-level columns the same way. The li-openhouse one is strictly more capable (nested structs, returns conflict paths). This PR reimplements a less-capable subset.
 2. The right move is to promote  SchemaValidationUtil upstream from li-openhouse to here as part of #558 then use it from both sides:
    - BaseIcebergSchemaValidator.normalizeSchemaCasingToTable calls SchemaValidationUtil.hasDuplicateCaseInsensitiveColumnNames for its safety guard.
    - li-openhouse continues to use it for the duplicate-rejection check.
3. Note that the two callers want opposite outcomes from the same predicate:
    - li-openhouse: if "duplicates exist → reject the write."
    - This PR: "duplicates exist → skip normalization (write may still succeed but only for exact-case)."

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BaseIcebergSchemaValidator.java
```diff
@@ -13,9 +17,63 @@
...
+
+  /**
+   * Renames top-level fields in {@code writeSchema} to use the casing from {@code tableSchema},
+   * matched by Iceberg field ID (not by name). Fields in {@code writeSchema} whose ID does not
+   * appear in {@code tableSchema} (genuinely new columns) are left unchanged.
+   *
+   * <p>This preserves the table's existing column casing when a writer submits columns with
+   * different casing (e.g. writer sends "id", table has "ID" → normalized to "ID").
+   */
+  static Schema normalizeSchemaCasingToTable(Schema writeSchema, Schema tableSchema) {
```
cbb330, line 41: this impl doesn't cover nested fields, which is in scope 

the impl can be quite complicated and impact is high since it returns a new schema. Please present/strategize it with AI agent in a way that gives confidence on the reliability. 

here is an example that looks prod ready at first glance

```Good direction. The right shape is functional + recursive + uses Iceberg's own visitor, so you stop hand-rolling type traversal and stop having a bug per nested-type variant. Three principles drive the redesign:

  1. No mutable state outside a single local builder per struct level. Inputs are immutable Iceberg types; the function returns a fresh Schema.
  2. Recurse using Iceberg's own machinery (TypeUtil.visit + TypeUtil.indexById) instead of instanceof ladders. Iceberg's visitor handles every Type variant exhaustively — if a new one is added (e.g. VariantType), the compiler tells you, not a silent passthrough.
  3. Match by fieldId at every depth. Iceberg's indexById(struct) already gives you a flat Map<Integer, NestedField> for the entire tree, so the lookup at any nesting level is O(1) and the matching key never depends on names.

  Recommended implementation

  import org.apache.iceberg.Schema;
  import org.apache.iceberg.types.Type;
  import org.apache.iceberg.types.TypeUtil;
  import org.apache.iceberg.types.Types;

  import java.util.ArrayList;
  import java.util.HashSet;
  import java.util.List;
  import java.util.Locale;
  import java.util.Map;
  import java.util.Set;

  /**
   * Returns true if {@code schema} has, at any nesting depth, two sibling fields whose names
   * differ only in case. Such schemas are excluded from case-insensitive write normalization
   * because the target field would be ambiguous.
   */
  static boolean hasCaseDuplicateFields(Schema schema) {
    return Boolean.TRUE.equals(
        TypeUtil.visit(schema, new TypeUtil.SchemaVisitor<Boolean>() {
          @Override public Boolean schema(Schema s, Boolean structResult) { return structResult; }

          @Override public Boolean struct(Types.StructType struct, List<Boolean> childResults) {
            if (childResults.stream().anyMatch(Boolean.TRUE::equals)) return true;
            Set<String> seen = new HashSet<>();
            for (Types.NestedField f : struct.fields()) {
              if (!seen.add(f.name().toLowerCase(Locale.ROOT))) return true;
            }
            return false;
          }

          @Override public Boolean field(Types.NestedField f, Boolean fieldResult)   { return fieldResult; }
          @Override public Boolean list(Types.ListType l, Boolean elementResult)     { return elementResult; }
          @Override public Boolean map(Types.MapType m, Boolean kRes, Boolean vRes)  { return Boolean.TRUE.equals(kRes) || Boolean.TRUE.equals(vRes); }
          @Override public Boolean primitive(Type.PrimitiveType p)                   { return false; }
        }));
  }

  /**
   * Rewrites every field name in {@code writeSchema} (top-level <em>and</em> nested) to use the
   * casing from {@code tableSchema}, matched by Iceberg field ID. Fields whose ID is not present
   * in {@code tableSchema} (genuinely new columns) are left unchanged. All other field attributes
   * (type, doc, optional/required, identifier-field-ids) are preserved.
   */
  static Schema normalizeSchemaCasingToTable(Schema writeSchema, Schema tableSchema) {
    final Map<Integer, Types.NestedField> tableById = TypeUtil.indexById(tableSchema.asStruct());

    Type rewritten = TypeUtil.visit(writeSchema, new TypeUtil.SchemaVisitor<Type>() {
      @Override public Type schema(Schema s, Type structResult) { return structResult; }

      @Override
      public Type struct(Types.StructType struct, List<Type> fieldTypes) {
        List<Types.NestedField> originals = struct.fields();
        List<Types.NestedField> rebuilt = new ArrayList<>(originals.size());
        for (int i = 0; i < originals.size(); i++) {
          Types.NestedField original = originals.get(i);
          Types.NestedField tableField = tableById.get(original.fieldId());
          String name = (tableField != null) ? tableField.name() : original.name();
          Type type  = fieldTypes.get(i);
          // Reuse the original if nothing changed (cheap reference-equality short-circuit).
          if (name.equals(original.name()) && type == original.type()) {
            rebuilt.add(original);
          } else {
            rebuilt.add(original.isOptional()
                ? Types.NestedField.optional(original.fieldId(), name, type, original.doc())
                : Types.NestedField.required(original.fieldId(), name, type, original.doc()));
          }
        }
        return Types.StructType.of(rebuilt);
      }

      @Override public Type field(Types.NestedField f, Type fieldResult) { return fieldResult; }

      @Override
      public Type list(Types.ListType list, Type elementResult) {
        Types.NestedField elem = list.fields().get(0);
        return elem.isOptional()
            ? Types.ListType.ofOptional(elem.fieldId(), elementResult)
            : Types.ListType.ofRequired(elem.fieldId(), elementResult);
      }

      @Override
      public Type map(Types.MapType map, Type keyResult, Type valueResult) {
        Types.NestedField k = map.fields().get(0);
        Types.NestedField v = map.fields().get(1);
        return v.isOptional()
            ? Types.MapType.ofOptional(k.fieldId(), v.fieldId(), keyResult, valueResult)
            : Types.MapType.ofRequired(k.fieldId(), v.fieldId(), keyResult, valueResult);
      }

      @Override public Type primitive(Type.PrimitiveType p) { return p; }
    });

    Types.StructType normalizedStruct = (Types.StructType) rewritten;
    return new Schema(
        writeSchema.schemaId(),
        normalizedStruct.fields(),
        writeSchema.identifierFieldIds());
  }

  Why this is more bug-resistant than the original

  ┌────────────────────────────────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │                                    What changed                                    │                                                                                             Why it matters                                                                                              │
  ├────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Recurses through TypeUtil.visit                                                    │ Covers struct, list element, map key, and map value at any depth. The PR's version only handled top level.                                                                                              │
  ├────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ TypeUtil.indexById(table.asStruct())                                               │ One Iceberg-supplied O(n) walk produces a flat Map<Integer, NestedField> for every nested field. Lookup at any depth is now tableById.get(id). No re-implementing the walk.                             │
  ├────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Locale.ROOT on toLowerCase                                                         │ Removes the Turkish-i locale bug.                                                                                                                                                                       │
  ├────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Visitor pattern, not instanceof                                                    │ If Iceberg adds a new compound type, the visitor's API shape forces you to deal with it; an instanceof ladder would silently fall through.                                                              │
  ├────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Reference-equality short-circuit on rebuild                                        │ When a struct's children all came back unchanged and no name differs, the original NestedField is reused — no allocation. The whole call is then O(n) reads with no writes for already-aligned schemas. │
  ├────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ field() and primitive() are pass-throughs                                          │ Pure functions, no special cases hidden inside helpers.                                                                                                                                                 │
  ├────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ All inputs treated as immutable; everything is built from constructors / factories │ The function is referentially transparent — given the same two Schema inputs, you always get the same Schema output.                                                                                    │
  └────────────────────────────────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

  What stays the same on purpose

  - Match by fieldId, not by name. This is still the load-bearing contract — fieldIds are stable across renames; names are not.
  - Genuinely new fields pass through verbatim. tableById.get(id) returning null is the signal for "writer added this column"; we don't invent a name for it.
  - Type, doc, and required/optional flag are taken from the writer, name is taken from the table. This preserves the writer's evolution intent (e.g. ADD doc, widen type, change required) while only normalizing the dimension that the writer can't be expected to get exactly right.
  - schemaId and identifierFieldIds preserved from the write schema. Identifier columns are part of the writer's evolution surface; they shouldn't be replaced.

  What hasCaseDuplicateFields now catches that it didn't before

  Before, only top-level. After, siblings at any depth. So this is now a duplicate (and correctly excluded from normalization):

  event: struct<
    ID: string,
    id: string     <-- ambiguous within the struct
  >

  But this is not a duplicate (different parents, no ambiguity):

  user: struct<id: string>,
  session: struct<id: string>

  — which matches Iceberg's own field-ID-keyed semantics.

  Optional further hardening (skip unless you want it)

  - Compare by reference at the schema level too. Wrap the whole function with if (writeSchema == tableSchema) return writeSchema;. Useless in practice (different objects always come in), but makes the contract more obviously a no-op for identity inputs.
  - Defensive guard inside normalizeSchemaCasingToTable for hasCaseDuplicateFields(tableSchema) == true. The caller already does this, but a precondition check would make the function safe to call from a second site without re-establishing it.
  - Pull the visitor out into a top-level package-private class (SchemaCasingVisitor) once it grows past ~30 lines — anonymous inner classes get hard to test in isolation. Right now it's still readable inline.

  Trade-off you might push back on

  The visitor adds a layer of indirection that a hand-rolled recursive Type rebuildType(Type t) function wouldn't. If you prefer transparency over future-proofing, the hand-rolled version is fine — it's the same shape, just an if/else ladder instead of polymorphic dispatch:

  private static Type rebuildType(Type t, Map<Integer, Types.NestedField> tableById) {
    if (t.isPrimitiveType()) return t;
    if (t.isListType())      { /* rebuild list */ }
    if (t.isMapType())       { /* rebuild map */ }
    /* struct */
  }

  I'd still take the visitor: the cost of one indirection is much less than the cost of a future Iceberg Type variant being silently mis-handled.
```

pandaamit91: Done

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/BaseIcebergSchemaValidator.java
pandaamit91: Done

cbb330 (approved):

PR #559, opened 2026-04-24
Case-insensitive OH table reads via targeted Spark analyzer rule

pandaamit91:
## Summary
- OpenHouse tables preserve column casing as stored in the catalog (e.g. ID, not id). When a Spark job runs with spark.sql.caseSensitive=true, queries that reference columns in a different casing than what the table stores fail with an unresolved attribute error.
  - A simple session-level fix (caseSensitive=false in OpenHouseCatalog.initialize()) was considered but rejected: it is session-wide and breaks pipelines that have case-duplicate column names in joins or intermediate DataFrames.
  - This PR introduces OHCaseInsensitiveResolveRule, a custom Spark analyzer rule injected via OpenhouseSparkSessionExtensions.injectResolutionRule. The rule renames UnresolvedAttribute nodes to match the stored column casing only for OH table relations, so Spark's own ResolveReferences rule finds an exact match on the next fixed-point iteration. The rule does not modify spark.sql.caseSensitive and has no effect on non-OH tables in the same session.
  - This is the read-path complement to the write-path normalization in doUpdateSchemaIfNeeded (separate PR). Together they make all OH table reads and writes case-insensitive end-to-end without mutating table schemas or session config.

## Changes

- [X] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Problem
A Spark job that sets spark.sql.caseSensitive=true causes reads from OpenHouse tables to fail when a query references a column in different casing than what the table stores (e.g. querying id on a table whose schema holds ID). This also breaks
   Spark views whose SQL was written against the stored casing.

  A previous approach — overriding spark.sql.caseSensitive=false in OpenHouseCatalog.initialize() — fixed the read failure but introduced a session-wide side effect: DataFrames or joins that intentionally had case-duplicate column names would
  become ambiguous and throw AnalysisException, and pipelines that legitimately depend on caseSensitive=true would break.

## Solution

  Replace the session-level override with OHCaseInsensitiveResolveRule, a custom Spark analyzer rule registered via OpenhouseSparkSessionExtensions.injectResolutionRule.

 How it works

  Spark's analyzer runs resolution rules in a fixed-point loop. Our rule is injected into the same loop and runs alongside ResolveReferences:

  Query:  SELECT id FROM openhouse.db.t   (table stores "ID")

  Iteration 1:
    ResolveRelations:             UnresolvedRelation → DataSourceV2Relation(output=["ID"])
    ResolveReferences:            UnresolvedAttribute("id") — no exact match → stays unresolved
    OHCaseInsensitiveResolveRule: sees "ID" in OH schema, renames UnresolvedAttribute("id") → ("ID")

  Iteration 2:
    ResolveReferences:            UnresolvedAttribute("ID") — exact match → AttributeReference("ID") ✓

  The rule:
  1. Scans the plan for DataSourceV2Relation nodes whose backing catalog has an OpenHouse catalog-impl (checked via spark.sql.catalog.<name>.catalog-impl — no hardcoded catalog names).
  2. Builds a lowercase → stored_name map from the relation's output columns. Tables where two or more columns share the same case-folded name are excluded — consistent with the server-side write-path guard.
  3. Renames any UnresolvedAttribute whose last name-part case-insensitively matches an OH column to use the stored casing.

  The rule does NOT modify spark.sql.caseSensitive. Non-OH tables and intermediate DataFrame operations in the same session are completely unaffected.
Since spark-3.5 bundles spark-3.1's runtime as a dependency, both Spark versions pick up this change automatically.

 Relationship to write-path fix

  This PR is the read-path complement to the write-path normalization in doUpdateSchemaIfNeeded (separate PR). Together they ensure:
  - Writes (server-side): write schema normalized to table casing before validation/storage — covers all write clients
  - Reads (client-side): Spark column references resolve against stored casing without touching session config

## Testing Done

Three tests added to CatalogOperationTest (run via catalogTest Gradle task, backed by a live embedded OH server):

  - testReadWithCaseMismatchSucceeds_andDoesNotChangeCaseSensitiveConfig: Creates a table with uppercase ID column, sets caseSensitive=true, queries with lowercase id. Asserts the row is returned correctly and that spark.sql.caseSensitive is still "true" after the query — confirming the rule does not mutate session config.
  - testViewWithCaseMismatchResolvesViaRule: Same table setup, but accessed through a Spark temp view whose SQL references id (lowercase). Asserts the view reads succeed with caseSensitive=true, confirming that view expansion is also covered.
  - testCaseDuplicateTableIsExcludedFromNormalization: Creates a table with both id (field 1) and ID (field 2). Asserts that an ambiguous reference throws rather than silently resolving to the wrong column — confirming the case-duplicate guard works correctly.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


cbb330 (commented):
can you check if this analyzer also helps solve the Write side tests that need client changes which i've called out here:

https://github.com/linkedin/openhouse/pull/562#issuecomment-4331756925

integrations/spark/spark-3.1/openhouse-spark-runtime/src/main/scala/com/linkedin/openhouse/spark/extensions/OHCaseInsensitiveResolveRule.scala
```diff
@@ -0,0 +1,88 @@
...
+ * <p>The rule does NOT modify {@code spark.sql.caseSensitive} and has no effect on non-OH tables
+ * or intermediate DataFrame operations in the same query.
+ */
+class OHCaseInsensitiveResolveRule(spark: SparkSession) extends Rule[LogicalPlan] {
+
+  override def apply(plan: LogicalPlan): LogicalPlan = {
+    val mappings = collectOHColumnMappings(plan)
+    if (mappings.isEmpty) return plan
+
+    plan.transformExpressions {
```
dxichen: this scope is global correct? we would need to prevent this rename for non openhouse catalogs (hive) tables as well

cbb330: Yes, the extensions are applied to hive catalog as well per our spark cluster config

pandaamit91: Thanks for the catch @cbb330 and @dxichen 

The bug was that collectOHColumnMappings correctly identified OH columns, but plan.transformExpressions was global and it renamed every UnresolvedAttribute in the plan, including those belonging to Hive or other non-OH catalog tables in the same query. Under caseSensitive=true, this would have broken resolution for non-OH tables sharing a case-folded column name with an OH column.

Fix: collectOHColumnMappings now also tracks column names from every non-OH resolved LeafNode (non-OH DataSourceV2Relation, HiveTableRelation, file scans, etc.). Names that appear in any non-OH relation are excluded from the OH mapping. The rename is only applied to names that are unambiguously OH-specific.

pandaamit91:
> can you check if this analyzer also helps solve the Write side tests that need client changes which i've called out here:
> 
> [#562 (comment)](https://github.com/linkedin/openhouse/pull/562#issuecomment-4331756925)



> can you check if this analyzer also helps solve the Write side tests that need client changes which i've called out here:
> 
> [#562 (comment)](https://github.com/linkedin/openhouse/pull/562#issuecomment-4331756925)

So I have written a different write resolver because of these reasons:

1. **The reads rule operates too early in the wrong node type**
OHCaseInsensitiveResolveRule matches UnresolvedAttribute nodes — column references that haven't been resolved yet (e.g. id typed in a SQL string before Spark knows which table it belongs to). By the time a write is being analyzed, the source DataFrame's columns are already fully resolved AttributeReference nodes. There are no UnresolvedAttribute nodes left to rewrite, so the reads rule simply never fires on write plans.

2. **The reads rule can't run before ResolveOutputRelation anyway**
Both rules are injected via injectResolutionRule, which places them at the end of Spark's Resolution fixed-point batch. ResolveOutputRelation is a built-in rule that runs inside that same batch — before injected rules get a chance to fire. So even if the reads rule could match write nodes, ResolveOutputRelation would have already thrown "Cannot find data for output column 'ID'" before the rule ever ran.

The write rule sidesteps this entirely with a different mechanism: ACCEPT_ANY_SCHEMA on the table makes outputResolved=true, which causes ResolveOutputRelation to skip OH tables completely. Then OHWriteSchemaNormalizationRule is registered as a post-hoc rule — a separate batch that runs after the main Resolution batch succeeds — so it can do the column normalization without racing against ResolveOutputRelation.

PR #562, opened 2026-04-27
test: add reproducer for case-insensitive write rejection (same field ID, different column casing)

pandaamit91:
## Summary

We have noticed that OH writes with different column casing already succeed in some cases and we want to validate the existing behavior before applying any fix. This PR does that — it adds characterization tests that document exactly which write paths work today and which ones don't, with an explanation of why.

Key Findings

  - df.writeTo().append() already works with the default caseSensitive=false setting. Spark's Iceberg integration maps DataFrame column names to table column names case-insensitively at analysis time. The data files are written using the stored
  column names, so the commit carries the unchanged existing schema — writeSchema.sameSchema(tableSchema) is true and the server's validateWriteSchema is never invoked. DaliSpark (a wrapper over df.writeTo()) gets this for free.
  - Explicit column-list SQL INSERT always fails in Spark 3.1 when casing differs — even with caseSensitive=false. Spark 3.1's ResolveInsertInto rule matches the INSERT column list case-sensitively regardless of the config setting (this only
  governs column references in SELECT expressions, not INSERT column lists). The unresolved column is silently dropped, causing AnalysisException: not enough data columns.
  - The server-side normalization fix is scoped to non-Spark clients — Trino DML, direct Iceberg Java API, and plain REST — that send a PATCH body with column names in a different casing than what's stored. Those clients don't go through Spark's
   case-insensitive resolution layer, so the server is the only place to normalize.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

  Test: Adds CaseInsensitiveWriteTest — a mock-based Spark e2e characterization test that establishes a baseline of which write paths already handle case-mismatched column names before any server-side fix is applied.
Screenshot 2026-04-27 at 2 33 44 PM

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/integrations/spark/spark-3.1/openhouse-spark-itest/src/test/java/com/linkedin/openhouse/spark/e2e/dml/CaseInsensitiveWriteTest.java
+++ b/integrations/spark/spark-3.1/openhouse-spark-itest/src/test/java/com/linkedin/openhouse/spark/e2e/dml/CaseInsensitiveWriteTest.java
@@ -297,4 +297,66 @@ public class CaseInsensitiveWriteTest {
 
     Assertions.assertDoesNotThrow(() -> df.writeTo("openhouse.dbCaseWrite.dfWriteUpper").append());
   }
+
+  /**
+   * {@code df.writeTo().append()} fails when {@code spark.sql.caseSensitive=true} and the DataFrame
+   * column names differ in casing from the stored schema.
+   *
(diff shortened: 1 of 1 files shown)


cbb330:
Thanks @pandaamit91 it is clear the current state of the world.

what is the AI for spark client since the mixed case will throw error before it lands on OH server?

PR #569, opened 2026-05-04
[DataLoader] Support more literal types in filters DSL to DataFusion conversion

robreeves:
## Summary

Filter expressions using `datetime`, `date`, `time`, `Decimal`, or `UUID` values fail with `TypeError` when a `TableTransformer` is active because `_literal_to_sql()` only handled `str`, `bool`, `int`, and `float`. This extends support for other types.

## Changes

- [ ] Internal API Changes
- [x] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

## Testing Done

All 234 tests pass. `make verify` clean.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/integrations/python/dataloader/src/openhouse/dataloader/filters.py
+++ b/integrations/python/dataloader/src/openhouse/dataloader/filters.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import typing
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
+from datetime import date, datetime
 from typing import Any
 
 from pyiceberg import expressions as ice
@@ -327,6 +328,12 @@ def _literal_to_sql(value: object) -> str:
         return exp.Literal.string(value).sql()
     if isinstance(value, bool):
         return exp.Boolean(this=True).sql() if value else exp.Boolean(this=False).sql()
+    if isinstance(value, datetime):
+        lit = exp.Literal.string(value.strftime("%Y-%m-%d %H:%M:%S"))
+        return exp.Cast(this=lit, to=exp.DataType.build("TIMESTAMP")).sql()
+    if isinstance(value, date):
+        lit = exp.Literal.string(value.isoformat())
+        return exp.Cast(this=lit, to=exp.DataType.build("DATE")).sql()
     if isinstance(value, (int, float)):
         return exp.Literal.number(value).sql()
     raise TypeError(f"Unsupported literal type: {type(value).__name__}")
--- a/integrations/python/dataloader/src/openhouse/dataloader/filters.py
+++ b/integrations/python/dataloader/src/openhouse/dataloader/filters.py
@@ -3,8 +3,10 @@ from __future__ import annotations
 import typing
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from datetime import date, datetime
+from datetime import date, datetime, time
+from decimal import Decimal
 from typing import Any
+from uuid import UUID
 
 from pyiceberg import expressions as ice
 from sqlglot import exp
@@ -334,8 +336,15 @@ def _literal_to_sql(value: object) -> str:
     if isinstance(value, date):
         lit = exp.Literal.string(value.isoformat())
         return exp.Cast(this=lit, to=exp.DataType.build("DATE")).sql()
+    if isinstance(value, time):
+        lit = exp.Literal.string(value.strftime("%H:%M:%S"))
+        return exp.Cast(this=lit, to=exp.DataType.build("TIME")).sql()
     if isinstance(value, (int, float)):
         return exp.Literal.number(value).sql()
+    if isinstance(value, Decimal):
+        return exp.Literal.number(value).sql()
+    if isinstance(value, UUID):
+        return exp.Literal.string(str(value)).sql()
     raise TypeError(f"Unsupported literal type: {type(value).__name__}")
 
 
--- a/integrations/python/dataloader/src/openhouse/dataloader/filters.py
+++ b/integrations/python/dataloader/src/openhouse/dataloader/filters.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import math
 import typing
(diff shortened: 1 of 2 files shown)


integrations/python/dataloader/src/openhouse/dataloader/filters.py
```diff
@@ -327,8 +330,21 @@ def _literal_to_sql(value: object) -> str:
...
         return exp.Boolean(this=True).sql() if value else exp.Boolean(this=False).sql()
+    if isinstance(value, datetime):
+        lit = exp.Literal.string(value.strftime("%Y-%m-%d %H:%M:%S"))
+        return exp.Cast(this=lit, to=exp.DataType.build("TIMESTAMP")).sql()
+    if isinstance(value, date):
+        lit = exp.Literal.string(value.isoformat())
+        return exp.Cast(this=lit, to=exp.DataType.build("DATE")).sql()
+    if isinstance(value, time):
+        lit = exp.Literal.string(value.strftime("%H:%M:%S"))
+        return exp.Cast(this=lit, to=exp.DataType.build("TIME")).sql()
```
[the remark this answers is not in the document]
robreeves: Fixed — datetime and time now format with microseconds (%f) and timezone offset (%z).

integrations/python/dataloader/src/openhouse/dataloader/filters.py
```diff
@@ -327,8 +330,21 @@ def _literal_to_sql(value: object) -> str:
...
+    if isinstance(value, date):
+        lit = exp.Literal.string(value.isoformat())
+        return exp.Cast(this=lit, to=exp.DataType.build("DATE")).sql()
+    if isinstance(value, time):
+        lit = exp.Literal.string(value.strftime("%H:%M:%S.%f"))
+        return exp.Cast(this=lit, to=exp.DataType.build("TIME")).sql()
     if isinstance(value, (int, float)):
         return exp.Literal.number(value).sql()
+    if isinstance(value, Decimal):
+        return exp.Literal.number(value).sql()
```
ShreyeshArangath, line 355: how is the behavior for NaN and high-precision decimal vals?

integrations/python/dataloader/src/openhouse/dataloader/filters.py
```diff
@@ -327,8 +330,21 @@ def _literal_to_sql(value: object) -> str:
...
     if isinstance(value, bool):
         return exp.Boolean(this=True).sql() if value else exp.Boolean(this=False).sql()
+    if isinstance(value, datetime):
+        lit = exp.Literal.string(value.strftime("%Y-%m-%d %H:%M:%S.%f%z"))
+        return exp.Cast(this=lit, to=exp.DataType.build("TIMESTAMP")).sql()
+    if isinstance(value, date):
+        lit = exp.Literal.string(value.isoformat())
+        return exp.Cast(this=lit, to=exp.DataType.build("DATE")).sql()
+    if isinstance(value, time):
+        lit = exp.Literal.string(value.strftime("%H:%M:%S.%f"))
```
ShreyeshArangath, line 346: does this also need to be tz-aware?

ShreyeshArangath (dismissed):
LGTM, just a couple of questions

integrations/python/dataloader/src/openhouse/dataloader/filters.py
robreeves: DataFusion doesn't support time timezone ([code](https://github.com/apache/datafusion/blob/main/datafusion/sql/src/planner.rs#L773)), but on second thought dropping the time zone silently is a bad way to handle it. For this case I decided to convert to UTC before creating the SQL.

integrations/python/dataloader/src/openhouse/dataloader/filters.py
```diff
@@ -327,8 +330,27 @@ def _literal_to_sql(value: object) -> str:
...
+    if isinstance(value, date):
+        lit = exp.Literal.string(value.isoformat())
+        return exp.Cast(this=lit, to=exp.DataType.build("DATE")).sql()
+    if isinstance(value, time):
+        # DataFusion does not support TIME WITH TIME ZONE (TIMETZ), so
+        # tz-aware values are converted to UTC before formatting.
+        if value.tzinfo is not None:
+            # Python can only convert timezones on a full datetime, so combine
+            # with an arbitrary date, convert to UTC, then extract the time back.
+            value = datetime.combine(date(2000, 1, 1), value).astimezone(UTC).time()
```
ShreyeshArangath: this seems silent, do we want to log it?

integrations/python/dataloader/src/openhouse/dataloader/filters.py
robreeves: High-precision decimals are preserved by sqlglot and parsed by DataFusion as Float64 (precision beyond ~15 digits is truncated — standard numeric literal behavior). Added an execution test to verify. For `Decimal('NaN')` and `Decimal('Inf')`, we now emit `CAST('...' AS DOUBLE)` since DataFusion doesn't recognize bare NaN/Infinity as numeric literals. Same fix applied to the float branch.

integrations/python/dataloader/src/openhouse/dataloader/filters.py
robreeves: This is actually the incorrect thing to do. DataFusion doesn't have a concept of timezone for time so the time filter needs to be in the same time zone as the data. Instead I throw an exception if time zone is used and say that.

integrations/python/dataloader/src/openhouse/dataloader/filters.py
robreeves: Removed due to https://github.com/linkedin/openhouse/pull/569#discussion_r3190875746

ShreyeshArangath (approved):

PR #579, opened 2026-05-12
Support returning HTS fields in table list api

jiang95-dev:
## Summary

Add support to V2 `searchTablesPaginated` endpoint to return the HTS fields. Currently only support returning tableLocation which is restriced by the validation layer, and we can relax it later. This is to unblock the ODD job so that it doesn't need to fetch metadata per table. If the fetching failed for reasons like timeout, we don't know what to do with the table, and will have to end up with halting the job. Fetching tableLocation directly from HTS is a much safer choice. Also we will restrict this functionality to only the system admin since it is violating the per-table permissions. 

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

`POST /v2/databases/{db}/tables/search` now accepts the query param `fields`. Passing `?fields=tableLocation` populates each result's tableLocation without doing per-table metadata loads. Requesting unpermitted fields will return 400.

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


services/tables/src/main/java/com/linkedin/openhouse/tables/controller/TablesController.java
```diff
@@ -118,10 +119,22 @@ public ResponseEntity<GetAllTablesResponseBody> searchTablesPaginated(
...
       @RequestParam(required = false, defaultValue = "50") int size,
-      @RequestParam(required = false) String sortBy) {
+      @RequestParam(required = false) String sortBy,
+      @Parameter(
+              description =
+                  "Optional body to request additional columns be populated on each table. "
+                      + "If omitted, only databaseId and tableId are returned.",
+              schema = @Schema(implementation = SearchTablesRequestBody.class))
+          @RequestBody(required = false)
+          SearchTablesRequestBody searchTablesRequestBody) {
```
abhisheknath2011: Introduction of SearchTablesRequestBody is not aligning with REST conventions as we are not using the columns params underneath as search params. As these are params are used to fetch can we just add as requests params some thing like `GET /v2/databases/{db}/tables?fields=tableLocation&page=1`

abhisheknath2011: Other option is to rename SearchTablesRequestBody to something like TableFieldsRequest.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/response/GetTableResponseBody.java
```diff
@@ -49,6 +50,7 @@ public class GetTableResponseBody {
       example =
           "<fs>://<hostname>/<openhouse_namespace>/<database_name>/<tableUUID>/metadata/<uuid>.metadata.json")
   @JsonProperty(access = JsonProperty.Access.READ_ONLY)
+  @JsonInclude(JsonInclude.Include.NON_NULL)
```
abhisheknath2011: As GetTableResponseBody is common for other endpoints, changing to non null might impact other endpoints?

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/TablesController.java
```diff
@@ -118,10 +119,22 @@ public ResponseEntity<GetAllTablesResponseBody> searchTablesPaginated(
       @Parameter(description = "Database ID", required = true) @PathVariable String databaseId,
       @RequestParam(required = false, defaultValue = "0") int page,
       @RequestParam(required = false, defaultValue = "50") int size,
-      @RequestParam(required = false) String sortBy) {
+      @RequestParam(required = false) String sortBy,
```
maluchari, line 122: This is a security leak. This exposes tableLocation to callers who cannot call getTable, which lets any engine with storage read access bypass the catalog's GET_TABLE_METADATA check. Also errors could be thrown differently misleading the users. Gate the projection behind GET_TABLE_METADATA privilege something like  authorizationUtils.checkTablePrivilege(tableDto, actingPrincipal, GET_TABLE_METADATA) that getTableMetadata does.

maluchari (changes requested):

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/TablesController.java
jiang95-dev: I think we can restrict this api call to only Privileges.SYSTEM_ADMIN. Is that ok?

services/tables/src/main/java/com/linkedin/openhouse/tables/api/validator/impl/OpenHouseTablesApiValidator.java
```diff
@@ -73,10 +80,23 @@ public void validateSearchTables(String databaseId) {
...
-  public void validateSearchTables(String databaseId, int page, int size, String sortBy) {
+  public void validateSearchTables(
+      String databaseId, int page, int size, String sortBy, List<String> columns) {
     List<String> validationFailures = new ArrayList<>();
     validateDatabaseId(databaseId, validationFailures);
     ApiValidatorUtil.validatePageable(page, size, sortBy, validationFailures);
+    if (columns != null && !columns.isEmpty()) {
+      List<String> unsupported =
+          columns.stream()
+              .filter(c -> !SUPPORTED_SEARCH_COLUMNS.contains(c))
```
mkuchenbecker: Exists would have the same impact and resolve to a bool vs list.

mkuchenbecker:
why does ODD need to know the columns?

jiang95-dev:
> why does ODD need to know the columns?

ODD needs to know tableLocation for each ative table. Previously we were calling getTable, but that leads to the halting of the whole job when one getTable fails. In that case, we can't continue because without the tableLocation, we will accidentally delete an active table. Therefore, I'm changing ODD to get tableLocation from the search API directly.

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/TablesController.java
maluchari: Are we not planning to propose this API for other usecases as well?

jiang95-dev: We will reject using this API with selected columns. But for the default `<databaseId, tableId>` usage is not restricted.

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/TablesController.java
jiang95-dev: Make sense. Changed it to query params instead.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/validator/impl/OpenHouseTablesApiValidator.java
jiang95-dev: Here the purpose is to collect and log all the unsupported columns that user requests.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/response/GetTableResponseBody.java
jiang95-dev: It won't but it is also not necessary. Removing it.

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/TablesController.java
maluchari: Sure, that would work.

abhisheknath2011 (approved):

maluchari (approved):
Thanks for addressing the authz comments

PR #584, opened 2026-05-14
internalcatalog: bump HDFS replication on metadata.json after write

mkuchenbecker:
## Summary

Bump the HDFS replication factor on each newly-written `metadata.json` file to 30 immediately after `TableMetadataParser.write(...)` returns inside `OpenHouseInternalTableOperations.doCommit(...)`. Paired with an internal `li-openhouse` infra PR that drops `dfs.replication` on Holdem from 30 -> 10, this moves the durability cost off the synchronous commit-write-ack path while preserving repl=30 durability on the persisted file.

**Ordering constraint (important):** this PR must merge **before or together with** the paired `li-openhouse` PR. If `li-openhouse` lands first, every newly-written `metadata.json` on Holdem will live at repl=10 indefinitely until this PR ships — a durability regression. Reviewers, please coordinate on merge order.

## Changes

- [x] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [x] Performance Improvements
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

**Internal API change / new behavior in `doCommit`:** Added a private `bumpMetadataReplicationIfHdfs(...)` helper invoked immediately after the existing synchronous metadata write. Behavior:

- HDFS-only via `storage.getClient() instanceof HdfsStorageClient`. Non-HDFS clients (local, blob) silently skip — their durability is managed elsewhere.
- Pulls `FileSystem` from `fileIOManager.getStorage(fileIO).getClient().getNativeClient()`, mirroring the existing `updateMetadataFieldForTable(...)` helper at the bottom of the same file.
- `fs.setReplication(new Path(newMetadataLocation), CatalogConstants.METADATA_FILE_HDFS_REPLICATION)`.
- Failures are caught, logged at WARN with the path and exception, and swallowed. They do NOT fail the commit — the file is already durably written at the cluster default replication, so a failed bump is a degraded-durability event, not a commit failure.
- Wrapped in `metricsReporter.executeWithStats(...)` under a new `metadata_replication_set_latency` metric, matching the existing class-level Micrometer convention. (No OpenTelemetry tracing was introduced — none exists in this class today.)

**Performance improvement (rationale):** After the paired `li-openhouse` change drops `dfs.replication` to 10, synchronous `TableMetadataParser.write(...)` will return once the file is durable at 10 datanodes instead of 30. The async bump in this PR restores on-disk durability to repl=30 without putting that pipeline-fill cost on the commit-write-ack path. The expected effect is a meaningful drop in commit p50/p99 metadata-write latency on Holdem; the new metric will quantify both the savings and the bump's own cost in production.

**Configuration choice:** `METADATA_FILE_HDFS_REPLICATION = (short) 30` is added as a constant on `CatalogConstants`. There is no established Spring-config-into-`OpenHouseInternalTableOperations` convention today, and the constructor is `@AllArgsConstructor`-generated, so adding a config-injected field would ripple through every construction site. The constant carries an explicit `TODO` to promote to a Spring config property (e.g. `openhouse.metadata.hdfs.replication`) in a follow-up PR.

**Interpretation note for reviewers:** the internal trigger described "the manifest json file." In Iceberg's commit layout the only JSON file written per commit is the table metadata file `<version>-<uuid>.metadata.json`. Manifest files proper (`*-m#.avro`) and the manifest list (`snap-*.avro`) are Avro. This PR treats `metadata.json` only. If the same durability-bump treatment is wanted for Avro manifest/manifest-list files, that requires a `FileIO` wrapper and is a separate follow-up.

**Files changed (4):**
- `iceberg/openhouse/internalcatalog/.../OpenHouseInternalTableOperations.java` — new helper + call site in `doCommit`, `org.apache.hadoop.fs.Path` import added.
- `iceberg/openhouse/internalcatalog/.../CatalogConstants.java` — new `METADATA_FILE_HDFS_REPLICATION` short constant + TODO.
- `iceberg/openhouse/internalcatalog/.../InternalCatalogMetricsConstant.java` — new `METADATA_REPLICATION_SET_LATENCY` metric name constant.
- `iceberg/openhouse/internalcatalog/src/test/.../OpenHouseInternalTableOperationsTest.java` — 3 new tests, 1 existing-test verify-count update.

## Testing Done

**Added tests** (all on `OpenHouseInternalTableOperationsTest`, mirroring the file's existing Mockito-based style):

1. `testDoCommitBumpsMetadataReplicationOnHdfs` — sets up the storage mock to return an `HdfsStorageClient` whose native client is a mock `FileSystem`, drives a commit, captures the args to `setReplication(Path, short)` and asserts the replication factor equals `CatalogConstants.METADATA_FILE_HDFS_REPLICATION` (30) and the path lives under the table location.
2. `testDoCommitDoesNotBumpReplicationOnLocalStorage` — drives a commit with a `LocalStorageClient` and verifies `setReplication` was never invoked (the `instanceof HdfsStorageClient` guard short-circuits).
3. `testDoCommitSucceedsWhenReplicationBumpFails` — stubs `setReplication` to throw `IOException`, drives a commit, asserts the commit completes successfully (`houseTableRepository.save(...)` is invoked once) and the bump was attempted. Validates the warn-and-swallow durability/availability trade-off.

**Updated existing test:** `testDoCommitUpdateMetadataForInitalVersionCommit` — `fileIOManager.getStorage(...)` is now called twice on the replicated-initial-version path (once by the new bump helper, once by the pre-existing `updateMetadataFieldForTable`). Updated the verify counts and added inline comments explaining each call's purpose.

**Module test run:** `./gradlew :iceberg:openhouse:internalcatalog:test` — all 69 tests green (including the 3 new ones).

**Spotless:** `./gradlew :iceberg:openhouse:internalcatalog:spotlessCheck` — clean (after one `spotlessApply` pass during authoring).

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

**PR plan:**
1. **This PR (`openhouse` repo)** — adds the synchronous `setReplication(..., 30)` bump after `metadata.json` is written. Must merge first or together with (2).
2. **Paired `li-openhouse` PR (separate repo, in flight)** — lowers `dfs.replication` from 30 -> 10 in `holdem-hdfs-site-extra.xml`. Must NOT merge before (1) or new `metadata.json` files on Holdem will live at repl=10 until this PR ships.
3. **Follow-up (separate PR, not blocking)** — promote `METADATA_FILE_HDFS_REPLICATION` from a constant on `CatalogConstants` to a Spring config property (`openhouse.metadata.hdfs.replication`) so deployments can tune the durable replication factor without a code change.
4. **Follow-up (separate PR, conditional on reviewer intent)** — if the same durability-bump treatment is desired for Avro manifest / manifest-list files (`*-m#.avro`, `snap-*.avro`), that needs a `FileIO` wrapper; out of scope here.

### DIFF
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
@@ -29,15 +29,13 @@ public final class CatalogConstants {
   public static final String EVOLVED_SCHEMA_KEY = "evolved.table.schema";
 
   /**
-   * Target HDFS replication factor applied to newly-written {@code .metadata.json} files
-   * immediately after the synchronous write returns. Paired with a li-openhouse PR that
-   * lowers the cluster default {@code dfs.replication}.
-   *
-   * <p>TODO(mkuchenb): promote this to a Spring config property
-   * ({@code openhouse.metadata.hdfs.replication}) so deployments can tune it without a
-   * code change. Kept as a constant for now to minimize PR surface area.
+   * Default HDFS replication factor used by {@link MetadataReplicationProperties} when no explicit
+   * value is configured. Held here so unit tests that construct {@link
+   * OpenHouseInternalTableOperations} without wiring the Spring properties bean still get a sane
+   * default. The bump itself is gated by {@link MetadataReplicationProperties#isEnabled()} and is
+   * off by default.
    */
-  public static final short METADATA_FILE_HDFS_REPLICATION = (short) 30;
+  public static final int DEFAULT_METADATA_FILE_HDFS_REPLICATION = 30;
 
   static final String FEATURE_TOGGLE_STOP_CREATE = "stop_create";
 
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
@@ -28,6 +28,17 @@ public final class CatalogConstants {
 
   public static final String EVOLVED_SCHEMA_KEY = "evolved.table.schema";
 
+  /**
+   * Target HDFS replication factor applied to newly-written {@code .metadata.json} files
+   * immediately after the synchronous write returns. Paired with a li-openhouse PR that
+   * lowers the cluster default {@code dfs.replication}.
+   *
+   * <p>TODO(mkuchenb): promote this to a Spring config property
+   * ({@code openhouse.metadata.hdfs.replication}) so deployments can tune it without a
+   * code change. Kept as a constant for now to minimize PR surface area.
+   */
+  public static final short METADATA_FILE_HDFS_REPLICATION = (short) 30;
+
   static final String FEATURE_TOGGLE_STOP_CREATE = "stop_create";
 
   static final String CLIENT_TABLE_SCHEMA = "client.table.schema";
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
@@ -37,6 +37,7 @@ import java.util.stream.Collectors;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
 import org.apache.iceberg.BaseMetastoreTableOperations;
 import org.apache.iceberg.PartitionField;
 import org.apache.iceberg.PartitionSpec;
@@ -330,6 +331,9 @@ public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperatio
             "updateMetadata to location {} succeeded, took {} ms",
             newMetadataLocation,
             System.currentTimeMillis() - metadataUpdateStartTime);
+        // Set replication after the synchronous write so the write path is not blocked
+        // on the higher replica count. Failures are logged and do not fail the commit.
+        bumpMetadataReplicationIfHdfs(newMetadataLocation);
       } catch (Exception e) {
         log.error(
             "updateMetadata to location {} failed after {} ms",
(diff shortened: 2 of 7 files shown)


iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
```diff
@@ -30,6 +28,17 @@ public final class CatalogConstants {
 
   public static final String EVOLVED_SCHEMA_KEY = "evolved.table.schema";
 
+  /**
+   * Target HDFS replication factor applied to newly-written {@code .metadata.json} files
+   * immediately after the synchronous write returns. Paired with a li-openhouse PR that
+   * lowers the cluster default {@code dfs.replication}.
+   *
+   * <p>TODO(mkuchenb): promote this to a Spring config property
```
mkuchenbecker: do this now

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -736,6 +698,62 @@ private String[] getCatalogMetricTags() {
...
+    if (storage == null || !(storage.getClient() instanceof HdfsStorageClient)) {
+      return;
+    }
+    final FileSystem fs = (FileSystem) storage.getClient().getNativeClient();
+    final short targetReplication = CatalogConstants.METADATA_FILE_HDFS_REPLICATION;
+    try {
+      metricsReporter.executeWithStats(
+          () -> {
+            try {
+              fs.setReplication(new Path(newMetadataLocation), targetReplication);
```
mkuchenbecker: We should have a strict timeout.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -359,17 +331,16 @@ updatedMtDataRef, io().newOutputFile(newMetadataLocation)),
             "updateMetadata to location {} succeeded, took {} ms",
             newMetadataLocation,
             System.currentTimeMillis() - metadataUpdateStartTime);
+        // Set replication after the synchronous write so the write path is not blocked
+        // on the higher replica count. Failures are logged and do not fail the commit.
+        bumpMetadataReplicationIfHdfs(newMetadataLocation);
```
mkuchenbecker: calling this function should be bated by a config

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -736,6 +698,62 @@ private String[] getCatalogMetricTags() {
...
+   * expected count without making the synchronous write path wait on the longer datanode
+   * pipeline ack.
+   *
+   * <p>HDFS-only. Non-HDFS clients (local, blob storage) silently skip.
+   *
+   * <p>Failures are logged at WARN and swallowed; they do not fail the commit.
+   *
+   * @param newMetadataLocation absolute HDFS path of the newly-written metadata.json
+   */
+  private void bumpMetadataReplicationIfHdfs(String newMetadataLocation) {
```
mkuchenbecker: rename to applyReplicationIfHDFS with the repl as a parameter.

cbb330:
i believe tables-service FS only touches metadata 

so we can set replication as per hdfs client config centrally.

would prefer that solution because we don't merge config into the business logic 

and config for hdfs lives in one place

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
mkuchenbecker: Done in 35d3d23c. Promoted to a Spring `@ConfigurationProperties` bean: `MetadataReplicationProperties` (prefix `openhouse.internal.catalog.metadata.replication`) with `enabled` (default `false`), `target` (default `30`), `timeoutMs` (default `10000`). `CatalogConstants` now only holds the default value the bean falls back to.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Done in 35d3d23c. The `setReplication` RPC is dispatched to a static daemon single-thread `ExecutorService` and bounded by `Future.get(timeoutMs, MILLISECONDS)`. On timeout the future is cancelled with interrupt and a new `metadata_replication_set_timeout` counter is incremented; commit succeeds regardless. Default timeout 10s, configurable via `MetadataReplicationProperties.timeoutMs`. Added `testDoCommitSucceedsWhenReplicationApplyTimesOut` which uses a 100ms configured timeout vs a 3s-sleeping `setReplication` stub to prove the timeout fires deterministically.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Done in 35d3d23c. The call site in `doCommit` is now gated on `metadataReplicationProperties.isEnabled()`, which defaults to **`false`** (opt-in per deployment, safe rollout). When the gate is off, a `metadata_replication_set_disabled` counter is incremented so the gate state is visible in metrics. New test `testDoCommitDoesNotApplyReplicationWhenDisabled` verifies `setReplication` is never invoked when disabled even with an HDFS storage client.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Done in 35d3d23c. Renamed to `applyReplicationIfHDFS`. Signature: `private void applyReplicationIfHDFS(String location, short replication, long timeoutMs)`. The replication factor and timeout are now both parameters, pulled from `MetadataReplicationProperties` by the caller in `doCommit`.

PR #586, opened 2026-05-14
OH case-insensitive writes via OHSparkCatalog and OHWriteSchemaNormalizationRule

pandaamit91:
## Summary
Writers (Spark SQL, DataFrame writeTo, Trino DML) may submit column names with different casing than what the OH table stores (e.g. "id" vs "ID"). With spark.sql.caseSensitive=true, Spark's ResolveOutputRelation rejects such writes with "Cannot find data for output column" before the OH server is reached.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [X] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [] Tests

Fix (two-part):

1. OHSparkCatalog extends SparkCatalog and annotates every loaded OH table with TableCapability.ACCEPT_ANY_SCHEMA. This causes DataSourceV2Relation.skipSchemaResolution to return true, making V2WriteCommand.outputResolved true and causing ResolveOutputRelation to skip schema validation for OH write commands.

2. OHWriteSchemaNormalizationRule (injectPostHocResolutionRule) runs after all standard resolution rules. For each resolved V2WriteCommand targeting an OH relation, it inserts a Project node that renames source columns to match the stored column casing (matched by field ID). This ensures Iceberg sees the correct stored casing without mutating spark.sql.caseSensitive.

Tables with case-duplicate columns (e.g. both "id" and "ID") are excluded from normalization — the target is ambiguous and writes must use exact casing.

## Testing Done

TestSparkSessionUtil and SparkTestBase are updated to use OHSparkCatalog instead of the bare SparkCatalog so all integration test sessions pick up the ACCEPT_ANY_SCHEMA capability.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations


integrations/spark/spark-3.1/openhouse-spark-runtime/src/main/java/com/linkedin/openhouse/spark/OHSparkCatalog.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * <pre>
+ *   spark.sql.catalog.openhouse=com.linkedin.openhouse.spark.OHSparkCatalog
+ *   spark.sql.catalog.openhouse.catalog-impl=com.linkedin.openhouse.spark.OpenHouseCatalog
+ * </pre>
+ */
+public class OHSparkCatalog extends SparkCatalog {
+
+  @Override
+  public SparkTable loadTable(Identifier ident) throws NoSuchTableException {
```
aastha25, line 40: We already have V2 sparkCatalog interface  implemented in the internal fork i think, or we did and needed to deramp it for an unrelated failure. 
We would need to co-ordinate the the two set of changes / make them compatible

integrations/spark/spark-3.5/openhouse-spark-runtime/src/main/scala/com/linkedin/openhouse/spark/extensions/OHWriteSchemaNormalizationRule.scala
```diff
@@ -0,0 +1,163 @@
...
+ *       if the types differ, cast) to match the target. This replicates the {@code Alias} +
+ *       {@code Cast} that {@code ResolveOutputRelation} would have inserted.</li>
+ * </ul>
+ *
+ * <p>In both modes, if source and target already match in name and type the rule returns the plan
+ * unchanged. If the column count differs the rule is a no-op (the mismatch is left for Iceberg or
+ * the OH server to report).
+ */
+class OHWriteSchemaNormalizationRule(spark: SparkSession) extends Rule[LogicalPlan] {
+
```
aastha25, line 38: we should have instrumentation here, to get observability into where are the casing differences.

aastha25:
if we set ACCEPT_ANY_SCHEMA = true, do we still need defensive approach in the OH server side code changes to normalize schema?

pandaamit91:
> if we set ACCEPT_ANY_SCHEMA = true, do we still need defensive approach in the OH server side code changes to normalize schema?

We would still need them for non-spark writes right? Like Iceberg Java API.

PR #599, opened 2026-05-22
Optimizer: Batched orphan file deletion using bin packing

abhisheknath2011:
## Summary
This is a follow up of Optimizer series PRs (Previous PR #534). Introduces `BatchedOrphanFilesDeletionSparkApp`, the multi-table counterpart of the existing single-table `OrphanFilesDeletionSparkApp`. One Spark job now processes a list of `(table, operationId)` pairs that the optimizer scheduler bin-packed into a single batch, reporting SUCCESS/FAILED per operation back to the Optimizer Service.

Stands up the **`optimizerclient` codegen module** so consumers (this app, and future ones) talk to the Optimizer Service via an auto-generated client rather than hand-rolled HTTP — mirrors the existing
  `:client:jobsclient`/`:client:tableclient` pattern.

Lands a **first-fit-decreasing bin packer** that is used in the followup PR (#604 ) to assemble tables in batches. Also consolidated existing bin packer under this.

  ### Key design choices

  - **Per-table failure isolation** — exceptions in one table are caught, FAILED is posted for that operationId, and remaining tables continue. The job exits 0 if at least one table succeeds.
  - **Recoverable result reporting** — if `POST /v1/optimizer/operations/{id}/update` fails after retries, the row stays `SCHEDULED` and the Analyzer's stale-timeout will re-queue it. No retry storms in the Spark driver.
  - **Optional optimizer-service callbacks** — `--resultsEndpoint`/`--operationIds`/`--tableUuids` are all optional, so the legacy `JobsScheduler` can launch the app without optimizer integration. When absent, the per-operation callback is
  skipped; HTS `StateManager` still tracks per-job lifecycle.
  - **Scheduler decides parallelism, not the app** — `--driverParallelism` is honoured verbatim; the app never picks its own thread count.
  - **`MAX_BATCH_SIZE = 200`** — hard ceiling enforced at `buildEntries` (Spark side) so a misconfigured scheduler can't blow past `ARG_MAX`. Operating point stays `--batchMaxItems 25`.

  ### Optimizer client (codegen)

  - **`services:optimizer`** wired with the same `service-specgen-convention` + springdoc plugins as `services:jobs`/`services:tables`/`services:housetables`. Port 8003 (slots into the existing 8000/8001/8002 allocation).
  - **`client:optimizerclient`** — three-line `build.gradle` mirroring `client:jobsclient`; generates `com.linkedin.openhouse.optimizer.client.{api,model,invoker}` from the spec.
  - **`client:secureclient`** picks up `OptimizerApiClientFactory` (parallels `JobsApiClientFactory`) for SSL/auth-aware `ApiClient` construction.
  - **`OptimizerServiceClient`** in `apps:spark` is a thin wrapper around the generated `TableOperationsControllerApi`, mirroring `JobsClient`'s shape — `RetryTemplate` (new `RetryUtil.getOptimizerApiRetryTemplate()`), surface is
  `Optional<TableOperationsHistory> updateOperation(operationId, request)`. No more OkHttp, no more hand-rolled DTOs.

 ### Bin packer
  - `FirstFitDecreasingBinPacker` — FFD by weight with secondary caps on bytes and item count; oversized items get a dedicated bin. Reuses the existing `Bin`/`BinItem` types from #534.
  - FirstFitBinPacker.java and FirstFitBinPackerTest.java — deleted (renamed / consolidated).

  ### Batched Spark app
  - `BatchedOrphanFilesDeletionSparkApp` — extends `BaseSparkApp`; iterates entries via a fixed thread pool, reuses `Operations.deleteOrphanFiles(...)` per table, posts per-operation status, runs the existing `TableStateValidator` per table

### Additional changes
  **Service-side spec generation (`services/optimizer`)**
  - `build.gradle` — added `service-specgen-convention` + springdoc + processes plugins. `openApi.customBootRun` overrides the production MySQL DataSource with in-memory H2 at build time (overrides live in `build.gradle`, not a profile file, so
   they don't ship inside the fat jar). Port 8003.
  - `developmentOnly 'com.h2database:h2:2.1.210'` — H2 reaches `bootRun`/specgen but is excluded from `bootJar` and from `api`/`runtimeElements` propagation. Pinned to 2.1.210 to match the version `:iceberg:openhouse:htscatalog` already pulls
  in repo-wide and avoid the slf4j-2.x bump that 2.2.x would have introduced.

  **Client codegen (`client/optimizerclient`)**
  - New module (3-line `build.gradle`) generating `TableOperationsControllerApi`, `TableStatsControllerApi`, `TableOperationsHistoryControllerApi` + their DTOs.
  - `client/secureclient/OptimizerApiClientFactory.java` — SSL/auth-aware factory mirroring `JobsApiClientFactory`.
  - `settings.gradle` — registers `:client:optimizerclient`.

  **Bin packer (`libs/optimizer/binpack`)**
  - `FirstFitDecreasingBinPacker.java` (moved from apps/spark). Pre-projected `BinItem` inputs, weight + item-count caps, oversized items get their own bin.
  - `FirstFitDecreasingBinPackerTest.java` (moved + adapted to the libs `BinItem` interface).

  **Spark app (`apps/spark`)**
  - `spark/BatchedOrphanFilesDeletionSparkApp.java` — multi-table OFD with worker-thread pool, per-table failure isolation, `Iterables.size()`-based orphan count (bounded driver heap), `MAX_BATCH_SIZE=200` guard.
  - `spark/optimizer/OptimizerServiceClient.java` — wraps generated `TableOperationsControllerApi`.
  - `util/RetryUtil.java` — added `getOptimizerApiRetryTemplate()`.
  - `build.gradle` — depends on `:libs:optimizer:binpack` with a targeted `exclude` on `log4j-slf4j2-impl` (transitively brought by services:optimizer; conflicts with apps:spark's bundled `log4j-slf4j-impl` 1.x bridge).

**Test using CLI:**
  --tableNames db.t1,db.t2,db.t3
  --operationIds op-uuid-1,op-uuid-2,op-uuid-3
  --tableUuids tab-uuid-1,tab-uuid-2,tab-uuid-3
  --resultsEndpoint http://optimizer.svc:8080
  --driverParallelism 4
  plus existing OFD knobs (`--ttl`, `--backupDir`, `--concurrentDeletes`, `--streamResults`, `--maxOrphanFileSampleSize`).

pull request in 2-3 lines.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

  **Open items for reviewers:** 
 - `OptimizerServiceClient` is add on top of the generated `TableOperationsControllerApi` — matches `JobsClient`'s pattern. Fine, or should it live inside `:client:optimizerclient` as a higher-level facade?
  - Apps:spark currently has to `exclude log4j-slf4j2-impl` when depending on `:libs:optimizer:binpack` because services:optimizer transitively brings the slf4j-2.x bridge while apps:spark ships the 1.x bridge. A repo-wide bridge-alignment
  cleanup would let us drop this exclude; happy to file a separate issue.
  - The bin packer ends up as two siblings in libs (`FirstFitBinPacker` — optimizer-coupled; `FirstFitDecreasingBinPacker` — agnostic). Consolidation is done in FirstFitDecreasingBinPacker.
  - Builder defaults (maxWeightPerBin = 1_000_000L, maxItemsPerBin = 50) are sized for OFD. If we expect other operation types with materially different cost shapes, we may want to move the defaults out of the packer class and into the per-operation config (Spring @Value already provides this in SchedulerConfig).
 

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/optimizer/OptimizerServiceClient.java
```diff
@@ -0,0 +1,81 @@
...
+  public void close() {
+    httpClient.dispatcher().executorService().shutdown();
+    httpClient.connectionPool().evictAll();
+  }
+
+  private static OkHttpClient defaultClient() {
+    return new OkHttpClient.Builder()
+        .connectTimeout(10, TimeUnit.SECONDS)
+        .readTimeout(30, TimeUnit.SECONDS)
+        .writeTimeout(30, TimeUnit.SECONDS)
```
mkuchenbecker: Ideally these would be configs.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/optimizer/OptimizerServiceClient.java
```diff
@@ -0,0 +1,81 @@
...
+        .build();
+  }
+
+  private static String stripTrailingSlash(String url) {
+    if (url == null || url.isEmpty()) {
+      throw new IllegalArgumentException("Optimizer Service base URL must be non-empty");
+    }
+    return url.endsWith("/") ? url.substring(0, url.length() - 1) : url;
+  }
```
mkuchenbecker: (1) Seems heavyweight
(2) Can we just assume not null or attempt to use optional instead if null might be present? Under what conditions is the url valid to be null?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,421 @@
...
+  private static final int DEFAULT_MIN_OFD_TTL_IN_DAYS = 3;
+
+  private final List<BatchEntry> entries;
+  private final String resultsEndpoint;
+  private final int driverParallelism;
+  private final long ttlSeconds;
+  private final String backupDir;
+  private final int concurrentDeletes;
+  private final boolean streamResults;
+  private final int maxOrphanFileSampleSize;
```
mkuchenbecker, line 70: Thoughts on a OFD parameters object that we use a builder to construct? The main thought was to encapsulate parameters rather than having them be manually maintained. I'm generally a fan of generating  with annotations to avoid this boilerplate as lines 62-92 are all just defining parameters and a public funciton to supply them.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,421 @@
...
+      OtelEmitter otelEmitter,
+      List<BatchEntry> entries,
+      String resultsEndpoint,
+      int driverParallelism,
+      long ttlSeconds,
+      String backupDir,
+      int concurrentDeletes,
+      boolean streamResults,
+      int maxOrphanFileSampleSize) {
+    super(jobId, stateManager, otelEmitter);
```
mkuchenbecker, line 84: There is a callback to complete job on HTS right? Do we need to adapt that or  is it fine to leave as-is?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,421 @@
...
+
+  private int runBatch(Operations ops, OptimizerServiceClient client) {
+    ExecutorService pool = Executors.newFixedThreadPool(driverParallelism);
+    try {
+      List<Future<Boolean>> futures = new ArrayList<>(entries.size());
+      for (BatchEntry entry : entries) {
+        futures.add(pool.submit(new TableWorker(ops, entry, client)));
+      }
+      int successes = 0;
+      for (int i = 0; i < futures.size(); i++) {
```
mkuchenbecker: nit, can we use collection funcitonal apis over for loops? Mainly just a style preference I have. 

futures.map(future -> {})

It avoids entries.get(i) from possibly having a problem.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,421 @@
...
+        try {
+          if (Boolean.TRUE.equals(futures.get(i).get())) {
+            successes++;
+          }
+        } catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
+          log.error("Interrupted while waiting on table {}", entries.get(i).getFqtn(), e);
```
mkuchenbecker: User error or system error?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,421 @@
...
+        } catch (ExecutionException e) {
+          // Per-table workers swallow their own failures and report FAILED upstream; an
+          // ExecutionException here means the worker itself threw, which we treat as a failed
+          // operation but otherwise let the batch continue.
+          log.error("Worker threw for table {}", entries.get(i).getFqtn(), e.getCause());
```
mkuchenbecker: Re-throw?
Should we be reporting the results here?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,421 @@
...
+        otelEmitter.count(
+            METRICS_SCOPE,
+            AppConstants.ORPHAN_FILE_COUNT,
+            orphanCount,
+            Attributes.of(AttributeKey.stringKey(AppConstants.TABLE_NAME), fqtn));
+        validate(fqtn);
+        success = true;
+        log.info("OFD success: fqtn={} orphansDetected={}", fqtn, orphanCount);
+      } catch (TableValidationException e) {
+        log.error("Post-job validation failed for fqtn={}", fqtn, e);
```
mkuchenbecker: Can we explain what post-run validation is? Should we have an otel counter / how we monitor for it?

It seems weird we would catch and log vs logging in the validate and short circuiting.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,421 @@
...
+      } catch (Throwable t) {
+        log.error("OFD failed: fqtn={} operationId={}", fqtn, entry.getOperationId(), t);
+      } finally {
+        reportResult(success);
```
mkuchenbecker: finally runs even if you return from the try.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,421 @@
...
+        otelEmitter.count(
+            METRICS_SCOPE,
+            "post_run_validation_error",
+            1,
+            Attributes.of(
+                AttributeKey.stringKey(AppConstants.TABLE_NAME),
+                fqtn,
+                AttributeKey.stringKey(AppConstants.JOB_NAME),
+                BatchedOrphanFilesDeletionSparkApp.class.getSimpleName()));
+        throw e;
```
mkuchenbecker, line 297: log too? We might want to standardize all errors get a log and a counter for later observability.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,421 @@
...
+              .status(success ? STATUS_SUCCESS : STATUS_FAILED)
+              .tableUuid(entry.getTableUuid())
+              .databaseName(entry.getDatabaseName())
+              .tableName(entry.getTableName())
+              .operationType(OPERATION_TYPE)
+              .build();
+      try {
+        client.updateOperation(body);
+      } catch (IOException e) {
+        log.error(
```
mkuchenbecker: counter? We can get signal on how many jobs are failing to emit metrics.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,421 @@
...
+            "Failed to report operation result; row will stay SCHEDULED until stale-timeout: operationId={} fqtn={}",
+            entry.getOperationId(),
+            entry.getFqtn(),
+            e);
+      }
+    }
+
+    private long resolveTtlSeconds(Table table) {
+      long resolved = ttlSeconds;
+      boolean oneDayTtlEnabled =
```
mkuchenbecker: Nit. I don't think we need to cache this bool. The appconst is clear.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    private int countOrphans(DeleteOrphanFiles.Result result) {
+      int count = 0;
+      for (String unused : result.orphanFileLocations()) {
+        count++;
+      }
+      return count;
+    }
+  }
```
mkuchenbecker: Can we do result.count()?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,421 @@
...
+        Integer.parseInt(
+            cmdLine.getOptionValue(
+                "maxOrphanFileSampleSize", String.valueOf(DEFAULT_MAX_ORPHAN_FILE_SAMPLE_SIZE))));
+  }
+
+  static List<BatchEntry> buildEntries(String tableNames, String operationIds, String tableUuids) {
+    if (tableNames == null
+        || operationIds == null
+        || tableUuids == null
+        || tableNames.isEmpty()
```
mkuchenbecker, line 387: Is there a practical limit to the number of tables in the job based on the input string length limits?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Inherited from BaseSparkApp.run(): stateManager.updateState(jobId, SUCCEEDED/FAILED) already fires via onStarted/onFinished plus heartbeats — HTS sees this job's lifecycle. The new optimizer-side callbacks are per-operation (per-table); HTS callback is per-job. Both are intentional and orthogonal.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: yes, that's the design — reportResult(success) must fire on both success and failure paths, and finally is the only branch that hits both.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Yes updated to use steams api to avoid using indexed based futures.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: This should be system error.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Yes results are reported and logged. We don't need to re-throw as this will break the flow.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Post run validation is existing code (Invoked for all the Spark App) in the base class to validate if the table exists and metadata is accessible. On validation failure corresponding error is logged and metrics are emitted. The execution will continue for the remaining tables as we don't want to fail the flow. Basically error is logged and reported.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Done

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Yes emitting counter metrics and logging error.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Done

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Use iterables count to reduce driver memory usage as we don't need full path materialization.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/optimizer/OptimizerServiceClient.java
```diff
@@ -0,0 +1,83 @@
...
+ * over the auto-generated client, wired with a {@link RetryTemplate} that retries 5xx with
+ * exponential backoff.
+ *
+ * <p>The batched Spark app calls {@link #updateOperation(String, UpdateOperationRequest)} once per
+ * finished operation to record SUCCESS or FAILED. Per the design, a missed update is recoverable —
+ * the operation row stays {@code SCHEDULED} and the Analyzer's stale-timeout will re-queue it — so
+ * this client surfaces but does not swallow failures.
+ */
+@Slf4j
+public class OptimizerServiceClient {
```
abhisheknath2011, line 30: This is based on existing clients pattern.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/optimizer/OptimizerServiceClient.java
abhisheknath2011: Updated the client based on existing pattern.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: This is based on existing code pattern used used for all the spark App. The diff would be large if we change here and we need to change existing app as well for consistency. Hence, would prefer refactoring this in a separate PR.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/optimizer/OptimizerServiceClient.java
```diff
@@ -0,0 +1,83 @@
...
+
+  private static final int REQUEST_TIMEOUT_SECONDS = 20;
```
mkuchenbecker: Why so high? I would make it shorter with backoff.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/RetryUtil.java
```diff
@@ -51,6 +51,15 @@ public static RetryTemplate getTablesApiRetryTemplate() {
         .build();
   }
 
+  public static RetryTemplate getOptimizerApiRetryTemplate() {
+    RetryTemplateBuilder builder = new RetryTemplateBuilder();
+    return builder
+        .maxAttempts(5)
```
mkuchenbecker, line 57: config for attempts? 

5*20 = 100.

Can we do exponential backoff with shorter timeouts? Exponentual can be a follow-up (unless its easy), but I can't think of a good reason to hold a connection 20s.

apps/spark/build.gradle
```diff
@@ -25,6 +25,13 @@ dependencies {
 
     implementation "org.apache.logging.log4j:log4j-slf4j-impl:${log4jVersion}"
     implementation project(':libs:datalayout')
+    // libs:optimizer:binpack transitively pulls services:optimizer which depends on
+    // log4j-slf4j2-impl (the slf4j-2.x bridge). Apps:spark already ships log4j-slf4j-impl (1.x);
+    // having both bridges loaded crashes at runtime with NoSuchMethodError on Log4jLoggerFactory
+    // because each bridge ships its own copy of that class with a different signature.
```
mkuchenbecker: nit: Is this comment useful? I would simplfy the comment to just talk why log4j exlcuded

libs/optimizer/binpack/src/main/java/com/linkedin/openhouse/optimizer/binpack/FirstFitDecreasingBinPacker.java
```diff
@@ -0,0 +1,133 @@
+package com.linkedin.openhouse.optimizer.binpack;
```
mkuchenbecker, line 1: Is this a rename or are there material changes? (diff is unclear)

services/optimizer/build.gradle
```diff
@@ -1,14 +1,56 @@
...
   implementation 'org.springframework.boot:spring-boot-starter-web:2.7.8'
   implementation 'mysql:mysql-connector-java:8.+'
-  testImplementation 'com.h2database:h2:2.2.224'
+  // {@code developmentOnly} (a Spring Boot Gradle plugin configuration) makes H2 available to
+  // {@code bootRun} — which springdoc uses to extract the OpenAPI spec under the {@code codegen}
+  // profile — without including it in {@code bootJar} or propagating it via {@code api}/{@code
+  // implementation} to downstream modules. Production deployments use MySQL via env vars.
+  //
+  // Version pinned to 2.1.210 to match the H2 :iceberg:openhouse:htscatalog already pulls in
+  // transitively repo-wide; H2 2.2.x switched to slf4j-api 2.0.x and would conflict with the
```
mkuchenbecker: I am not sure the value of these comments

mkuchenbecker (commented):
Added a couple comments. If the binpacker was just renamed and we align on reties im good to ship

libs/optimizer/binpack/src/main/java/com/linkedin/openhouse/optimizer/binpack/FirstFitDecreasingBinPacker.java
abhisheknath2011: Yes this rename exactly reflects working of the bin packing strategy used here + consolidation of bin packing. Added overloaded pack method which is used to bin pack tables for batched OFD (in the follow up PR - #604 ) which is wired in the existing scheduler flow.

abhisheknath2011: The algorithm was already FFD (sorts by weight descending before placing); the original name omitted the "Decreasing" qualifier even though the code includes it. The rename guards against confusion if a plain first-fit variant ever lands.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,443 @@
...
+          .mapToInt(submission -> awaitOne(submission.getKey(), submission.getValue(), client))
+          .sum();
+    } finally {
+      shutdownPool(pool);
+    }
+  }
+
+  private int awaitOne(BatchEntry entry, Future<Boolean> future, OptimizerServiceClient client) {
+    try {
+      return Boolean.TRUE.equals(future.get()) ? 1 : 0;
```
mkuchenbecker, line 146: doesn't jsut returning feature.get() cast it to a an int?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,443 @@
...
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      log.error("Worker interrupted (likely job cancellation): fqtn={}", entry.getFqtn(), e);
```
mkuchenbecker: I would not addd the (likely job cancellation) as those reasons can change over time.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,443 @@
...
+   * POST the per-operation outcome to the Optimizer Service via the generated client. The wrapper
+   * returns {@link java.util.Optional#empty()} when the call exhausts retries — we log + count and
+   * leave the operation row at SCHEDULED so the Analyzer's stale-timeout can re-queue it.
+   */
+  private void reportResult(BatchEntry entry, boolean success, OptimizerServiceClient client) {
+    UpdateOperationRequest body =
+        new UpdateOperationRequest()
+            .operationId(entry.getOperationId())
+            .status(
+                success
```
mkuchenbecker: is boolean expressive enough? Do we need a delineation of use vs system error? I might assume states grow over time and pass the status here vs passing a bool interpreted as a status.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,443 @@
...
+            AppConstants.ORPHAN_FILE_COUNT,
+            orphanCount,
+            Attributes.of(AttributeKey.stringKey(AppConstants.TABLE_NAME), fqtn));
+        validate(fqtn);
+        success = true;
+        log.info("OFD success: fqtn={} orphansDetected={}", fqtn, orphanCount);
+      } catch (Throwable t) {
+        log.error("OFD failed: fqtn={} operationId={}", fqtn, entry.getOperationId(), t);
+      } finally {
+        reportResult(entry, success, client);
```
mkuchenbecker: your finally can throw and there is no catch.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,443 @@
...
+                AttributeKey.stringKey(AppConstants.JOB_NAME),
+                BatchedOrphanFilesDeletionSparkApp.class.getSimpleName()));
+        throw e;
+      }
+    }
+
+    private long resolveTtlSeconds(Table table) {
+      long resolved = ttlSeconds;
+      if (Boolean.parseBoolean(
+          table.properties().getOrDefault(AppConstants.OFD_ONE_DAY_TTL_ENABLED_KEY, "false"))) {
```
mkuchenbecker, line 304: nit, optional would avoid the "false" -> bool conversion.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,443 @@
...
+        resolved = TimeUnit.DAYS.toSeconds(1);
+      }
+      String tableType =
+          table
+              .properties()
+              .getOrDefault(AppConstants.OPENHOUSE_TABLE_TYPE_KEY, AppConstants.TABLE_TYPE_PRIMARY);
+      if (AppConstants.TABLE_TYPE_REPLICA.equals(tableType)) {
+        long days = Duration.ofSeconds(resolved).toDays();
+        if (days < DEFAULT_MIN_OFD_TTL_IN_DAYS) {
+          resolved = TimeUnit.DAYS.toSeconds(DEFAULT_MIN_OFD_TTL_IN_DAYS);
```
mkuchenbecker: nit
```suggestion
      if (AppConstants.TABLE_TYPE_REPLICA.equals(tableType)
          && Duration.ofSeconds(resolved).toDays() < DEFAULT_MIN_OFD_TTL_IN_DAYS) {
          resolved = TimeUnit.DAYS.toSeconds(DEFAULT_MIN_OFD_TTL_IN_DAYS);
```

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
```diff
@@ -0,0 +1,443 @@
...
+        new AppsOtelEmitter(Collections.singletonList(DefaultOtelConfig.getOpenTelemetry()));
+    createApp(args, otelEmitter).run();
+  }
+
+  public static BatchedOrphanFilesDeletionSparkApp createApp(
+      String[] args, OtelEmitter otelEmitter) {
+    List<Option> extraOptions = new ArrayList<>();
+    extraOptions.add(
+        new Option(
+            null, "tableNames", true, "Comma-separated list of fully-qualified table names"));
```
mkuchenbecker: nit, can we either do one param per line or single line option definition?

Additionally named parameters would help.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/optimizer/OptimizerServiceClient.java
abhisheknath2011: This was added as 20 sec as per existing code. As backoff with retry is done using RetryUtil, update the timeout as 5 sec and we can increase that if needed.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: future.get() returns Boolean (the worker is Callable<Boolean>), not an int — and Java doesn't implicitly cast Boolean → int. The ? 1 : 0 converts to int because awaitOne is consumed by .mapToInt(...).sum() to count successes. The Boolean.TRUE.equals(...) instead of a direct cast is also defensive against a null Boolean return.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Done.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Yes updated to include status in addition to success or failed which can be included in future.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Added a catch block for failures.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: This is existing properties map which returns string so we can apply optional.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Done

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Used named parameters

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Yes capped at MAX_BATCH_SIZE = 200 (enforced in buildEntries, which throws if exceeded). The math: each entry contributes up to ~120 chars (36-char UUID × 3 parallel fields in the optimizer-path: --tableNames + --operationIds + --tableUuids), so 200 tables ≈ 24 KB of CLI args. That's well under typical Linux ARG_MAX = 128 KB with headroom for the spark-submit envelope and JVM flags. The operating point is --batchMaxItems 25 (default on the scheduler side); MAX_BATCH_SIZE is the footgun ceiling, not the recommendation. The scheduler also fail-fasts on --batchMaxItems > 200 so a misconfiguration surfaces before any job launches. Plan is to start with max batch of 50 or less and we can adjust based on observation.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/RetryUtil.java
abhisheknath2011: This is an existing code. For the optimizer client updated the request timeout to 5 sec. The backoff with retry is already part of this utility. We can improve that in future if needed.

apps/spark/build.gradle
abhisheknath2011: updated.

services/optimizer/build.gradle
abhisheknath2011: Updated the comment.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
mkuchenbecker: I am not grokking the ternary operator on a boolean to convert to 1 or 0 vs a cast. One has branches and the other doesn't.

mkuchenbecker (dismissed):
I am interested in the ternary operator on boolean, but that is non-blocking.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/spark/BatchedOrphanFilesDeletionSparkApp.java
abhisheknath2011: Will check if this can be refactored or changed in the future PRs.

mkuchenbecker (approved):

PR #601, opened 2026-05-24
Add tableProperties to TableAuditEvent

james5418:
## Summary

Surface a snapshot of Iceberg table properties on the per-commit TableAuditEvent.

Emission is gated by allowlist:

  | Config | Default | Purpose |
  |---|---|---|
  | `cluster.iceberg.tables.audit.table-properties-allowlist` | `[]` | Which keys are appropriate to audit. [] = nothing emitted. |
  | `cluster.iceberg.tables.audit.table-property-value-max-size` | `256KB` | Per-value byte cap; oversized values skipped. |
  | `cluster.iceberg.tables.audit.table-properties-total-max-size` | `512KB` | Combined byte cap; properties skipped once total would exceed. |

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

./gradlew :services:tables:test --tests IcebergSnapshotsApiHandlerAuditTest

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
@@ -18,6 +18,7 @@ import com.linkedin.openhouse.tables.audit.model.OperationStatus;
 import com.linkedin.openhouse.tables.audit.model.OperationType;
 import com.linkedin.openhouse.tables.audit.model.TableAuditEvent;
 import com.linkedin.openhouse.tables.config.InternalCatalogProperties;
+import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.HashMap;
 import java.util.List;
@@ -543,24 +544,52 @@ public class TableAuditAspect {
 
   /**
    * Narrows the committed table properties down to the configured allowlist ({@code
-   * cluster.iceberg.tables.audit.table-properties-allowlist}). Returns {@code null} when there is
-   * nothing to emit so downstream audit handlers can skip the field entirely. Iterates the
-   * allowlist rather than the source so cost is O(|allowlist|) regardless of source size.
+   * cluster.iceberg.tables.audit.table-properties-allowlist}) and enforces two byte-size caps to
+   * keep the audit event within the Kafka producer's max.request.size budget: a per-value cap
+   * ({@code table-property-value-max-size}) and a combined-total cap ({@code
+   * table-properties-total-max-size}). Values exceeding either cap are skipped with a warning log.
+   * Returns {@code null} when there is nothing to emit so downstream audit handlers can skip the
+   * field entirely. Iterates the allowlist rather than the source so cost is O(|allowlist|)
+   * regardless of source size.
    */
   private Map<String, String> filterTableProperties(Map<String, String> source) {
     if (source == null || source.isEmpty()) {
       return null;
     }
-    List<String> allowlist = internalCatalogProperties.getAudit().getTablePropertiesAllowlist();
+    InternalCatalogProperties.Audit auditConfig = internalCatalogProperties.getAudit();
+    List<String> allowlist = auditConfig.getTablePropertiesAllowlist();
     if (allowlist == null || allowlist.isEmpty()) {
       return null;
     }
+    long maxValueBytes = auditConfig.getTablePropertyValueMaxSize().toBytes();
+    long maxTotalBytes = auditConfig.getTablePropertiesTotalMaxSize().toBytes();
     Map<String, String> filtered = new HashMap<>();
+    long totalBytes = 0;
     for (String key : allowlist) {
       String value = source.get(key);
-      if (value != null) {
-        filtered.put(key, value);
+      if (value == null) {
+        continue;
       }
+      long valueBytes = value.getBytes(StandardCharsets.UTF_8).length;
+      if (valueBytes > maxValueBytes) {
+        log.warn(
+            "Dropping audited table-property '{}': value size {} bytes exceeds per-value cap {} bytes",
+            key,
+            valueBytes,
+            maxValueBytes);
+        continue;
+      }
+      if (totalBytes + valueBytes > maxTotalBytes) {
+        log.warn(
+            "Dropping audited table-property '{}': including it would exceed total cap {} bytes "
+                + "(already accumulated {} bytes)",
+            key,
+            maxTotalBytes,
+            totalBytes);
+        continue;
+      }
+      filtered.put(key, value);
+      totalBytes += valueBytes;
     }
     return filtered.isEmpty() ? null : filtered;
   }
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
@@ -371,12 +371,18 @@ public class TableAuditAspect {
     } else {
       operationType = OperationType.COMMIT;
     }
+    CreateUpdateTableRequestBody createUpdateTableRequestBody =
+        icebergSnapshotRequestBody.getCreateUpdateTableRequestBody();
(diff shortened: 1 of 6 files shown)


services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
```diff
@@ -371,12 +371,18 @@ protected ApiResponse<GetTableResponseBody> auditPutIcebergSnapshots(
...
+        icebergSnapshotRequestBody.getCreateUpdateTableRequestBody();
     TableAuditEvent.TableAuditEventBuilder eventBuilder =
         TableAuditEvent.builder()
             .eventTimestamp(Instant.now())
             .databaseName(databaseId)
             .tableName(tableId)
-            .operationType(operationType);
+            .operationType(operationType)
+            .tableProperties(
+                createUpdateTableRequestBody == null
```
cbb330: tableProperties is read from createUpdateTableRequestBody (the client's request payload), not from the actual committed Iceberg table state. If OpenHouse adds or mutates properties server-side during commit (e.g. openhouse.tableVersion, openhouse.lastModifiedTime), those won't appear in the audit event. 


The fix can be to get it from the response body. this response from claude looks like good next steps.

>The fix is straightforward given what the aspect already has. result (the ApiResponse<GetTableResponseBody>) is available after point.proceed(), and
  result.getResponseBody() already has getTableLocation() called on it — getTableProperties() is right there too.
  The change in TableAuditAspect.java: remove tableProperties from the pre-proceed() builder, delay .build() until after proceed() so you can read from the
  response:

```java
  // Remove the createUpdateTableRequestBody variable entirely.
  // Remove .tableProperties(...) from the builder chain.
  // Change the try block to:
  try {
    result = (ApiResponse<GetTableResponseBody>) point.proceed();
    TableAuditEvent event = eventBuilder
        .tableProperties(result.getResponseBody().getTableProperties())
        .build();
    buildAndSendEvent(event, OperationStatus.SUCCESS, result.getResponseBody().getTableLocation());
  } catch (Throwable t) {
    buildAndSendEvent(eventBuilder.build(), OperationStatus.FAILED, null);
    throw t;
  }
```


>  On the failure path eventBuilder.build() has no tableProperties — which is correct, there's no committed state to read from.
  One thing to verify before doing this: confirm GetTableResponseBody has getTableProperties(). It almost certainly does since the response describes full
  table state, but worth a quick grep in the codebase.

cbb330 (changes requested):

jiang95-dev (changes requested):
What is the use case for adding table properties? 

We should cap on the size of table properties as some can be quite large. See this PR: https://github.com/linkedin-multiproduct/li-openhouse/pull/688. Large events can cause OOM and losing events.

james5418:
> What is the use case for adding table properties?
- Table properties are current-value-only. Once a property changes the prior value is gone.
- No durability. There is no audit trail linking a property value to the snapshot that was active when it was set.
- ACL-gated. Cannot read properties without per-table ACL grants.

> We should cap on the size of table properties as some can be quite large. See this PR: https://github.com/linkedin-multiproduct/li-openhouse/pull/688. Large events can cause OOM and losing events.

Added a new configuration property `cluster.iceberg.tables.audit.table-property-value-max-length`, which defaults to 0 (unlimited). Values longer than N characters are dropped from the event with a warning log.

services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
```diff
@@ -529,6 +540,43 @@ protected ApiResponse<Void> auditUpdateDatabaseAclPolicies(
...
 
+  /**
+   * Narrows the committed table properties down to the configured allowlist ({@code
+   * cluster.iceberg.tables.audit.table-properties-allowlist}) and drops any value longer than
+   * {@code cluster.iceberg.tables.audit.table-property-value-max-length} characters. Returns {@code
+   * null} when there is nothing to emit so downstream audit handlers can skip the field entirely.
+   * Iterates the allowlist rather than the source so cost is O(|allowlist|) regardless of source
+   * size.
+   */
+  private Map<String, String> filterTableProperties(Map<String, String> source) {
```
abhisheknath2011, line 555: Can we exclude some system level operation properties such as dlo specific which can be very large? (cc: @jiang95-dev )?

abhisheknath2011: We also ran into OOM issues earlier while logging large events.

cbb330:
@jiang95-dev the table property audit is required for a partner team's observability use case which must read freshness watermarks from table properties every time a commit is made

@jiang95-dev , @abhisheknath2011 the existing table property config `cluster.iceberg.tables.audit.table-properties-allowlist` prevents it from being too large, by only allowing specific keys. it is default {} and only allowed keys enter 

the keys we plan to add:
- any with "watermark" in the name e.g. completionWatermark, inferredHighWatermark, etc.
- ASL (which can be large)

cbb330:
@james5418 `cluster.iceberg.tables.audit.table-property-value-max-length`

is redundant with the other config. we can limit size already, by only allowing keys with small size. lets only have `cluster.iceberg.tables.audit.table-properties-allowlist` for now

cbb330 (dismissed):

abhisheknath2011:
> @jiang95-dev the table property audit is required for a partner team's observability use case which must read freshness watermarks from table properties every time a commit is made
> 
> @jiang95-dev , @abhisheknath2011 the existing table property config `cluster.iceberg.tables.audit.table-properties-allowlist` prevents it from being too large, by only allowing specific keys. it is default {} and only allowed keys enter
> 
> the keys we plan to add:
> 
> * any with "watermark" in the name e.g. completionWatermark, inferredHighWatermark, etc.
> * ASL (which can be large)

How do we make the solution generic to capture observability around table property update? In this case only known allowlisted properties would be emitted as part of audit event, but rest all would be ignored. We can exclude long system level properties. But for building  observability around table property update we need to capture the changes in audit event.

cbb330:
since this only carries the allowlisted subset, can we rename `tableProperties` to `auditedTableProperties`? avoids implying it's the full set

cbb330:
> How do we make the solution generic to capture observability around table property update? In this case only known allowlisted properties would be emitted as part of audit event, but rest all would be ignored. We can exclude long system level properties. But for building observability around table property update we need to capture the changes in audit event.

@abhisheknath2011 its a tradeoff between safety and usability. here we are choosing safety, and mitigating with easy cluster.yaml to add new keys.

maluchari:
> @jiang95-dev the table property audit is required for a partner team's observability use case which must read freshness watermarks from table properties every time a commit is made
> 
> @jiang95-dev , @abhisheknath2011 the existing table property config `cluster.iceberg.tables.audit.table-properties-allowlist` prevents it from being too large, by only allowing specific keys. it is default {} and only allowed keys enter
> 
> the keys we plan to add:
> 
> * any with "watermark" in the name e.g. completionWatermark, inferredHighWatermark, etc.
> * ASL (which can be large)

If a TableAuditEvent for update event is detected, can freshness watermark be determined from that? If we decide to go with the allowlist can we also add a value-size cap?

cbb330:
> If a TableAuditEvent for update event is detected, can freshness watermark be determined from that? If we decide to go with the allowlist can we also add a value-size cap?

I also asked the same question to the customer wanting this. they said that completionWatermark carries specific semantic: data of this snapshot has been updated with new stuff. 

while other signals, like commit time, can give a false positive signal. for example, if the data is backfilled. it looks identical from tableauditevent perspective.

cbb330 (approved):

jiang95-dev (approved):
LGTM. Thanks for addressing the comments.

abhisheknath2011 (approved):

PR #603, opened 2026-05-27
Fix multi-schema replication intermediates dropped on snapshots-endpoint commits

dushyantk1509:
## Summary
Replication commits arriving through PUT /iceberg/v2/snapshots with multi-schema-jump intermediate schemas were silently losing them due to two issues that combine to leave the dest replica with a snapshot whose schema-id points at a schema not present in schemas[]. Trino strictly resolves snapshot schema-id against schemas[] and fails with "Cannot find schema with schema id N"; Spark falls back to current-schema-id and reads succeed.

Two fixes:

1. TablesMapper#toTableDto(TableDto, IcebergSnapshotsRequestBody) was missing the @Mapping for newIntermediateSchemas. The CreateUpdateTableRequestBody overload had it (added in #407), but the IcebergSnapshotsRequestBody overload used for the snapshots endpoint did not — so intermediates packed correctly by the client got dropped at the server-side DTO mapping step.

2. OpenHouseTableOperations#doCommit (after #524 introduced putSnapshotsForReplace) was routing cross-cluster REPLICA_TABLE replication commits through the RTAS replace path because they update both metadata and snapshots in one commit. The server's REPLACE branch treats the commit as a fresh table creation and uses CLIENT_TABLE_SCHEMA (bootstrap-era schema) as the final schema, not the request's actual target. This produces a different but related corruption pattern. Detect REPLICA commits (base REPLICA_TABLE + metadata PRIMARY_TABLE + different cluster IDs) and route them through the regular putSnapshots path so the intermediate-schemas plumbing fires.

Both fixes are needed in tandem: without (1), even the regular update branch drops intermediates; without (2), commits hitting the post-#524 client get routed around (1) entirely.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

## Testing Done

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
@@ -111,7 +111,15 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
     boolean metadataUpdated = isMetadataUpdated(base, metadata);
     boolean snapshotsUpdated = areSnapshotsUpdated(base, metadata);
     try {
-      if (metadataUpdated && snapshotsUpdated && base != null) {
+      // REPLICA_TABLE commits from cross-cluster replication can update both
+      // metadata (schema, properties) and snapshots in one commit. They must go through the
+      // regular putSnapshots path so the server's update branch can apply newIntermediateSchemas;
+      // the replace path treats the commit as a fresh table creation and does not preserve
+      // multi-schema delta semantics. Reserve putSnapshotsForReplace for genuine CTAS/RTAS, where
+      // the metadata being committed is PRIMARY (not a REPLICA_TABLE update).
+      boolean isReplicaCommit =
+          getTableType(base, metadata) == CreateUpdateTableRequestBody.TableTypeEnum.REPLICA_TABLE;
+      if (metadataUpdated && snapshotsUpdated && base != null && !isReplicaCommit) {
         // Only CTAS and RTAS can update both metadata and snapshots at the same time.
         // When the table exists, it will be a replace commit operation.
         putSnapshotsForReplace(base, metadata);
@@ -217,30 +225,36 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
 
   /**
    * If request is coming from replication process, createUpdateTableRequestBody.tableType should be
-   * REPLICA_TABLE Replication process requests are identified based on difference between table
-   * types and cluster_id between base, metadata
+   * REPLICA_TABLE. Replication process requests are identified based on difference between table
+   * types and cluster_id between base, metadata. Returns {@code null} when either side lacks the
+   * {@code openhouse.tableType} property — this matters for {@link #doCommit}'s dispatch which may
+   * be invoked very early (e.g. by replace transactions) before properties are populated.
    */
   @VisibleForTesting
   CreateUpdateTableRequestBody.TableTypeEnum getTableType(
       TableMetadata base, TableMetadata metadata) {
+    String metaTypeStr = metadata.properties().get(OPENHOUSE_TABLE_TYPE_KEY);
     if (base != null) {
-      CreateUpdateTableRequestBody.TableTypeEnum baseTableType =
-          CreateUpdateTableRequestBody.TableTypeEnum.valueOf(
-              base.properties().get(OPENHOUSE_TABLE_TYPE_KEY));
-      CreateUpdateTableRequestBody.TableTypeEnum metadataTableType =
-          CreateUpdateTableRequestBody.TableTypeEnum.valueOf(
-              metadata.properties().get(OPENHOUSE_TABLE_TYPE_KEY));
-      // check if commit request is from replication case
-      if (baseTableType == CreateUpdateTableRequestBody.TableTypeEnum.REPLICA_TABLE
-          && metadataTableType == CreateUpdateTableRequestBody.TableTypeEnum.PRIMARY_TABLE
-          && !base.properties()
-              .get(OPENHOUSE_CLUSTER_ID_KEY)
-              .equals(metadata.properties().get(OPENHOUSE_CLUSTER_ID_KEY))) {
-        return baseTableType;
+      String baseTypeStr = base.properties().get(OPENHOUSE_TABLE_TYPE_KEY);
+      if (baseTypeStr != null && metaTypeStr != null) {
+        CreateUpdateTableRequestBody.TableTypeEnum baseTableType =
+            CreateUpdateTableRequestBody.TableTypeEnum.valueOf(baseTypeStr);
+        CreateUpdateTableRequestBody.TableTypeEnum metadataTableType =
+            CreateUpdateTableRequestBody.TableTypeEnum.valueOf(metaTypeStr);
+        String baseCluster = base.properties().get(OPENHOUSE_CLUSTER_ID_KEY);
+        String metaCluster = metadata.properties().get(OPENHOUSE_CLUSTER_ID_KEY);
+        // check if commit request is from replication case
+        if (baseTableType == CreateUpdateTableRequestBody.TableTypeEnum.REPLICA_TABLE
+            && metadataTableType == CreateUpdateTableRequestBody.TableTypeEnum.PRIMARY_TABLE
+            && baseCluster != null
+            && !baseCluster.equals(metaCluster)) {
+          return baseTableType;
+        }
       }
     }
-    return CreateUpdateTableRequestBody.TableTypeEnum.valueOf(
-        metadata.properties().get(OPENHOUSE_TABLE_TYPE_KEY));
+    return metaTypeStr != null
+        ? CreateUpdateTableRequestBody.TableTypeEnum.valueOf(metaTypeStr)
+        : null;
   }
 
   @VisibleForTesting
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
@@ -119,12 +119,15 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
       // the metadata being committed is PRIMARY (not a REPLICA_TABLE update).
       boolean isReplicaCommit =
           getTableType(base, metadata) == CreateUpdateTableRequestBody.TableTypeEnum.REPLICA_TABLE;
-      if (metadataUpdated && snapshotsUpdated && base != null && !isReplicaCommit) {
+      if (isReplicaCommit) {
+        if (snapshotsUpdated) {
+          putSnapshots(base, metadata);
+        } else if (metadataUpdated) {
+          createUpdateTable(base, metadata);
+        }
+      } else if (metadataUpdated && snapshotsUpdated && base != null) {
         // Among PRIMARY-table commits, only CTAS and RTAS update both metadata and snapshots at
-        // the same time; when the table already exists this is an RTAS (replace) commit. REPLICA
-        // commits also touch both, but are excluded above via isReplicaCommit and handled by the
-        // regular putSnapshots path so the server preserves newIntermediateSchemas — the replace
-        // path would treat them as a fresh creation and drop the multi-schema deltas.
+        // the same time; when the table already exists this is an RTAS (replace) commit.
         putSnapshotsForReplace(base, metadata);
       } else if (snapshotsUpdated) {
         putSnapshots(base, metadata);
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
@@ -120,8 +120,11 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
       boolean isReplicaCommit =
           getTableType(base, metadata) == CreateUpdateTableRequestBody.TableTypeEnum.REPLICA_TABLE;
       if (metadataUpdated && snapshotsUpdated && base != null && !isReplicaCommit) {
-        // Only CTAS and RTAS can update both metadata and snapshots at the same time.
-        // When the table exists, it will be a replace commit operation.
+        // Among PRIMARY-table commits, only CTAS and RTAS update both metadata and snapshots at
+        // the same time; when the table already exists this is an RTAS (replace) commit. REPLICA
+        // commits also touch both, but are excluded above via isReplicaCommit and handled by the
+        // regular putSnapshots path so the server preserves newIntermediateSchemas — the replace
+        // path would treat them as a fresh creation and drop the multi-schema deltas.
         putSnapshotsForReplace(base, metadata);
       } else if (snapshotsUpdated) {
         putSnapshots(base, metadata);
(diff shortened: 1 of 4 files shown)


integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -111,7 +111,15 @@ public void doCommit(TableMetadata base, TableMetadata metadata) {
...
-      if (metadataUpdated && snapshotsUpdated && base != null) {
+      // REPLICA_TABLE commits from cross-cluster replication can update both
+      // metadata (schema, properties) and snapshots in one commit. They must go through the
+      // regular putSnapshots path so the server's update branch can apply newIntermediateSchemas;
+      // the replace path treats the commit as a fresh table creation and does not preserve
+      // multi-schema delta semantics. Reserve putSnapshotsForReplace for genuine CTAS/RTAS, where
+      // the metadata being committed is PRIMARY (not a REPLICA_TABLE update).
+      boolean isReplicaCommit =
+          getTableType(base, metadata) == CreateUpdateTableRequestBody.TableTypeEnum.REPLICA_TABLE;
+      if (metadataUpdated && snapshotsUpdated && base != null && !isReplicaCommit) {
```
mkuchenbecker: What makes a replica commit different that we need to special case?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -111,7 +111,15 @@ public void doCommit(TableMetadata base, TableMetadata metadata) {
...
     boolean snapshotsUpdated = areSnapshotsUpdated(base, metadata);
     try {
-      if (metadataUpdated && snapshotsUpdated && base != null) {
+      // REPLICA_TABLE commits from cross-cluster replication can update both
+      // metadata (schema, properties) and snapshots in one commit. They must go through the
+      // regular putSnapshots path so the server's update branch can apply newIntermediateSchemas;
+      // the replace path treats the commit as a fresh table creation and does not preserve
+      // multi-schema delta semantics. Reserve putSnapshotsForReplace for genuine CTAS/RTAS, where
+      // the metadata being committed is PRIMARY (not a REPLICA_TABLE update).
+      boolean isReplicaCommit =
```
mkuchenbecker, line 120: what about if replica, then putsnapshots?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -111,7 +111,15 @@ public void doCommit(TableMetadata base, TableMetadata metadata) {
...
+          getTableType(base, metadata) == CreateUpdateTableRequestBody.TableTypeEnum.REPLICA_TABLE;
+      if (metadataUpdated && snapshotsUpdated && base != null && !isReplicaCommit) {
         // Only CTAS and RTAS can update both metadata and snapshots at the same time.
```
mkuchenbecker: this comment is no longer accurate.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -217,30 +225,36 @@ && getTableType(base, metadata)
...
-          CreateUpdateTableRequestBody.TableTypeEnum.valueOf(
-              metadata.properties().get(OPENHOUSE_TABLE_TYPE_KEY));
-      // check if commit request is from replication case
-      if (baseTableType == CreateUpdateTableRequestBody.TableTypeEnum.REPLICA_TABLE
-          && metadataTableType == CreateUpdateTableRequestBody.TableTypeEnum.PRIMARY_TABLE
-          && !base.properties()
-              .get(OPENHOUSE_CLUSTER_ID_KEY)
-              .equals(metadata.properties().get(OPENHOUSE_CLUSTER_ID_KEY))) {
-        return baseTableType;
+      String baseTypeStr = base.properties().get(OPENHOUSE_TABLE_TYPE_KEY);
```
mkuchenbecker, line 247: Why might this not exist?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -217,30 +225,36 @@ && getTableType(base, metadata)
...
-   * types and cluster_id between base, metadata
+   * REPLICA_TABLE. Replication process requests are identified based on difference between table
+   * types and cluster_id between base, metadata. Returns {@code null} when either side lacks the
+   * {@code openhouse.tableType} property — this matters for {@link #doCommit}'s dispatch which may
+   * be invoked very early (e.g. by replace transactions) before properties are populated.
    */
   @VisibleForTesting
   CreateUpdateTableRequestBody.TableTypeEnum getTableType(
       TableMetadata base, TableMetadata metadata) {
+    String metaTypeStr = metadata.properties().get(OPENHOUSE_TABLE_TYPE_KEY);
```
mkuchenbecker, line 245: nit, I would move this to after the if so on the case where base is null its not evaluated.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -217,30 +225,36 @@ && getTableType(base, metadata)
...
+            CreateUpdateTableRequestBody.TableTypeEnum.valueOf(metaTypeStr);
+        String baseCluster = base.properties().get(OPENHOUSE_CLUSTER_ID_KEY);
+        String metaCluster = metadata.properties().get(OPENHOUSE_CLUSTER_ID_KEY);
+        // check if commit request is from replication case
+        if (baseTableType == CreateUpdateTableRequestBody.TableTypeEnum.REPLICA_TABLE
+            && metadataTableType == CreateUpdateTableRequestBody.TableTypeEnum.PRIMARY_TABLE
+            && baseCluster != null
+            && !baseCluster.equals(metaCluster)) {
+          return baseTableType;
+        }
```
mkuchenbecker, line 261: This logic does not makse sense to me. Why can we not trust the table type?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
abhisheknath2011: Does that mean that this excludes RTAS commit for replica table?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
abhisheknath2011: Trying to understand the change here.  Seems like this change is applicable only for replica table to determine table type?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
dushyantk1509: It will go to this block

else if (snapshotsUpdated) {
        putSnapshots(base, metadata);

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
dushyantk1509: This block is supposed to be used for RTAS/CTAS where these is no concept of intermediate schemas but the issue is that this also sources replica commits from gaas into replace snapshots loosing intermediate schemas.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
dushyantk1509: This is needed for the end return statement.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
dushyantk1509: This is added as a safety check for legacy tables.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
dushyantk1509: Just to be clear, this logic is pre-existing and this change only introduce null checks.

Metadata's type can be trusted by default; this branch is the one exception. In a cross-cluster replication commit, the destination replica's metadata is overwritten with the source primary's content, so metadata.tableType says PRIMARY even though this table is a REPLICA. base is the destination's truth, so for that fingerprint (base=REPLICA, metadata=PRIMARY, different cluster) we override to base.

I do agree that it can be collapsed to "base if non-null, else metadata". I would put it out of scope of this PR.

abhisheknath2011 (dismissed):
Looks good to me. Lets get a stamp from Mike as well.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -111,9 +111,20 @@ public void doCommit(TableMetadata base, TableMetadata metadata) {
     boolean metadataUpdated = isMetadataUpdated(base, metadata);
     boolean snapshotsUpdated = areSnapshotsUpdated(base, metadata);
     try {
-      if (metadataUpdated && snapshotsUpdated && base != null) {
-        // Only CTAS and RTAS can update both metadata and snapshots at the same time.
-        // When the table exists, it will be a replace commit operation.
+      // REPLICA_TABLE commits from cross-cluster replication can update both
```
mkuchenbecker, line 114: I don't think detecting that this is a replica seems incorrect vs detecting the type of change and applying it correctly. Why can't we detect the condition and apply it vs detecting the table type?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
mkuchenbecker: what are legacy tables?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
mkuchenbecker: To me the concept of a Replica table as a first class citizen is adding the friction here.  A replica there is nothign special about other than the permissions on who can modify it. Its unclear to me why this is a table type, and why that needs a special case.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
dushyantk1509: The existing implementation already tries to detect a change-type: metadataUpdated && snapshotsUpdated && base != null is meant to mean "this is a replace (RTAS/CTAS)." The problem is a cross-cluster replication commit also updates both metadata and snapshots in one shot, so it matches that same condition — the two operations are indistinguishable by their metadata/snapshot deltas alone.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
mkuchenbecker: I'm saying inverting the logic is easier to follow. Detect "condition -> do thing" is much easier to follow than "detect lack of condition we previously missed."

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
mkuchenbecker: That's what's unclear to me is why this needs a special commit type at all. This is what's unclear by this logic:

1 - Why does Replication have special handling at-all in this case? What is the distinguishing feature of Replication that requires it to be handled specifically
2 - Why is replication table type not enough of  a signal to determine what to do?

I understand we need to preserve history but "replica" type seems like the wrong thing to be detecting, and if it is the thing that we are detecting then it should be enough to have its own dedicated logic "when replica do x"

mkuchenbecker (commented):
If we are keeping replica tables as separately handled, a clear "if replica, do X" will make the logic easier to follow and later remove if the case is no longer needed, and protects replication from changes to other commit types.

cbb330:
+1 to merge. this stops the data loss. 

But, this fix adds a fourth way to detect an intent the client already knows but never states. Server infers "replace vs. not" from comparing before/after states and then corrects that guess with three table properties (is-a-copy, is-an-original, cluster names differ) and now a further check to steer replication back to the update path. Each is a workaround for the same root which is that the caller knows what operation it's doing and doesn't send that operation to the server, so the server reconstructs it. Every new commit shape will need another clue and another branch.

The durable fix is for the client to declare the operation instead of the server deducing it. This is what the open-source Iceberg REST catalog does that we can draw inspiration from. A commit is an explicit list of changes to apply (add this schema, set current schema, add this snapshot, point this branch at it) plus preconditions, with no "replace mode" to detect. Replication stops being special, same change list as anything else, just more "add schema" entries. 

With that in mind, I suggest a not-blocking follow-up: add one field where the caller declares the operation (replication / replace / append) and route the logic on that enum in server. then delete the metadata-comparison guess and the property-sniffing that was bandaided over time. 

Longer term, we can align with the REST spec [0] and replace the two schema fields with one ordered list of schema additions. Today the final schema goes in one field and the in-between schema versions go in a separate list and it is handled as a branched logic on server. Reducing branching logic helps prevent bugs on edge cases. And then openhouse.tableType and openhouse.clusterId can stop being used to pick a code path, as sort of side-affect magic variables. 

[0] - In the Iceberg REST catalog, a commit (`UpdateTableRequest`, sent to `POST /v1/.../tables/{table}`) has two arrays:
* updates: an ordered list of change actions, each tagged with an action name. The relevant ones here: `add-schema` (adds one schema version to the table's schema list), `set-current-schema` (points the table at one of those schema versions by id), `add-snapshot` (adds a data snapshot), `set-snapshot-ref` (points a branch, e.g. main, at a snapshot).
* requirements: preconditions the server checks before applying, e.g. `assert-ref-snapshot-id` ("the main branch must currently point at snapshot X") or `assert-create` ("the table must not already exist").

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -111,9 +111,23 @@ public void doCommit(TableMetadata base, TableMetadata metadata) {
...
+      // multi-schema delta semantics. Reserve putSnapshotsForReplace for genuine CTAS/RTAS, where
+      // the metadata being committed is PRIMARY (not a REPLICA_TABLE update).
+      boolean isReplicaCommit =
+          getTableType(base, metadata) == CreateUpdateTableRequestBody.TableTypeEnum.REPLICA_TABLE;
+      if (isReplicaCommit) {
+        if (snapshotsUpdated) {
+          putSnapshots(base, metadata);
+        } else if (metadataUpdated) {
+          createUpdateTable(base, metadata);
+        }
```
mkuchenbecker, line 127: nit. else not needed, just return short circuit.

mkuchenbecker (approved):

PR #608, opened 2026-05-27
(wip | no review) feat(tables): post-commit stats push to optimizer

mkuchenbecker:
## Summary

After a successful Iceberg snapshot commit, the tables service runs registered
post-commit operations against the freshly-saved `TableDto`. The first operation is
the optimizer stats push (`PUT /v1/optimizer/stats/{tableUuid}`), opting in per table
via `maintenance.optimizer.stats.enabled=true` and gated globally by
`tables.postcommit.enabled` (default `false`).

The framework is one generic dispatcher; individual operations describe payload +
endpoint and let the dispatcher own async / timeout / error swallow / metrics.

- Per-op wall-clock budget enforced by `PostCommitDispatcher`
  (`tables.postcommit.per-op-timeout-ms`, default 3000 ms).
- Best-effort: any operation error is recorded as a Micrometer counter + warn log,
  never propagated to the commit caller.
- No HDFS round-trip — stats sourced from `Snapshot.summary()` already in memory at
  load time, captured on `TableDto.currentSnapshot` as `Optional<CurrentSnapshotInfo>`
  (snapshot id + summary).
- The `OptimizerStatsPostCommitOperation` retains its own per-attempt timeout +
  bounded retry on retryable errors (network, 408, 429, 5xx, `TimeoutException`).

### Why async (the explicit justification)

A synchronous post-commit hook converts an optimizer outage into a tables-service
write outage. The stats push is a best-effort scheduling signal, not a
write-correctness step — its blast radius must not include the write path. The
crash-loss window (JVM dies between commit and HTTP push completion) is acceptable
because operations are cumulative: the next commit re-pushes the current state, and
the consumer self-corrects from missing data.

The complexity stays bounded by living in one place (`PostCommitDispatcher`). New
operations register without re-deriving async plumbing; reviewers reason about
async semantics once.

### Configuration

| Property | Default | Notes |
|---|---|---|
| `tables.postcommit.enabled` | `false` | Master switch for the dispatcher; per-instance rollout. |
| `tables.postcommit.per-op-timeout-ms` | 3000 | Outer wall-clock ceiling per operation. |
| `optimizer.stats.enabled` | `false` | Master switch for the optimizer-stats operation bean. |
| `optimizer.stats.base-uri` | _(unset)_ | Required when both above are `true`. |
| `optimizer.stats.per-attempt-timeout-ms` | 1000 | Bounds each HTTP attempt inside the operation. |
| `optimizer.stats.max-attempts` | 3 | One initial + up to two retries. |

### Metrics (Micrometer, tagged `op={operation name}`)

| Key | Type | Tags |
|---|---|---|
| `postcommit_op_duration` | timer | `op`, `outcome={success, timeout, network_error, server_error, client_error, unknown_error}` |
| `postcommit_op_skipped` | counter | `op` |
| `postcommit_op_failed` | counter | `op`, `outcome={timeout, network_error, server_error, client_error, prepare_threw, unknown_error}` |

### Wiring

`PostCommitDispatcher`, `OptimizerStatsConfig` (WebClient bean), and
`OptimizerStatsPostCommitOperation` are each `@ConditionalOnProperty` — when their
flag is `false` no bean is constructed. `IcebergSnapshotsServiceImpl` consumes the
dispatcher as `Optional<PostCommitDispatcher>` and the on-commit hook is a literal
no-op when absent.

### Follow-up

The wire body already carries `snapshotId`. Server-side dedupe (use the snapshot ID
as an idempotency token on upsert and reject out-of-order replays) is intentionally
deferred to a follow-up change so this PR stays focused on the framework + the
first operation.

## Changes

- [x] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [x] Refactoring
- [ ] Documentation
- [x] Tests

Internal API: new `services/tables/.../services/postcommit/` package
(`PostCommitOperation` interface, `PostCommitDispatcher`, `PostCommitProperties`);
new `services/tables/.../model/CurrentSnapshotInfo`; `TableDto.currentSnapshot`
replaces `currentSnapshotSummary`; new metric keys in `MetricsConstant`;
`TableStatsController` publishes `BASE_PATH` / `TABLE_PATH_TEMPLATE` constants.

New features: opt-in post-commit operation framework + the first operation
(optimizer stats push), each independently gated by config.

Refactor: optimizer-stats client recast as a `PostCommitOperation` impl; outer
timeout, `onErrorResume`, and `.subscribe()` removed from the operation and moved
into the dispatcher.

Tests: new `PostCommitDispatcherTest` (6 cases); `OptimizerStatsClientTest` renamed
to `OptimizerStatsPostCommitOperationTest` (6 cases adapted to the `prepare()`
contract); `IcebergSnapshotsServiceTest` updated; `TableDtoMappingTest` updated.

## Testing Done

`PostCommitDispatcherTest` — 6 cases, all synchronous via `decorate(...).block(BLOCK_MAX)`:

- `dispatch_invokesEveryRegisteredOperationsPrepare`
- `decorate_emptyPrepare_incrementsSkippedAndReturnsEmpty`
- `decorate_successfulWork_recordsDurationWithSuccessOutcome`
- `decorate_workSignalsError_incrementsFailedWithClassifiedOutcomeAndSwallowsError`
- `decorate_prepareThrowsSynchronously_incrementsFailedAndDispatchContinuesToLaterOps`
- `decorate_workExceedsPerOpTimeout_incrementsFailedWithTimeoutOutcome`

`OptimizerStatsPostCommitOperationTest` — 7 cases (1 stable-name + 6 behavior):

- `name_isStableTagValue`
- `prepare_optedIn_emitsRequestWithFullPayload` (asserts wire body field-for-field, including new `snapshotId`)
- `prepare_notOptedIn_returnsEmpty`
- `prepare_noSnapshot_returnsEmpty`
- `prepare_5xxThenSuccess_retriesAndCompletes`
- `prepare_allAttemptsFail_propagatesUnderlyingError`
- `prepare_4xxNonRetryable_doesNotRetry`

`IcebergSnapshotsServiceTest.testPostCommitDispatcherInvokedAfterSuccessfulCommit`
verifies `dispatcher.dispatch(savedDto)` is invoked exactly once after a successful
save (replaces the prior `OptimizerStatsClient.report` assertion).

`TableDtoMappingTest` — `currentSnapshot` added to `FIELDS_UNMAPPABLE`.

Local build: `:services:tables:test` + `:services:optimizer:test` BUILD SUCCESSFUL;
`:services:tables:spotlessCheck`, `:services:common:spotlessCheck`,
`:services:optimizer:spotlessCheck` clean.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

Both `tables.postcommit.enabled` and `optimizer.stats.enabled` default to `false`,
so existing deployments see no behavioural change. The feature only activates when
the global dispatcher flag, the operation's flag, and the per-table opt-in property
are all set.


services/common/src/main/java/com/linkedin/openhouse/common/metrics/MetricsConstant.java
```diff
@@ -63,4 +63,10 @@ private MetricsConstant() {}
   public static final String HTS_LIST_TABLES_REQUEST = "hts_list_tables_request";
   public static final String HTS_LIST_TABLES_TIME = "hts_list_tables_time";
   public static final String HTS_SEARCH_TABLES_TIME = "hts_search_tables_time";
+
+  // Optimizer post-commit stats push (Tables → Optimizer)
+  public static final String OPTIMIZER_STATS_DURATION = "optimizer_stats_duration";
```
mkuchenbecker: max duration timeout

services/common/src/main/java/com/linkedin/openhouse/common/metrics/MetricsConstant.java
```diff
@@ -63,4 +63,10 @@ private MetricsConstant() {}
...
+  public static final String OPTIMIZER_STATS_ATTEMPTS = "optimizer_stats_attempts";
+  public static final String OPTIMIZER_STATS_SKIPPED = "optimizer_stats_skipped";
+  public static final String OPTIMIZER_STATS_FAILED_FINAL = "optimizer_stats_failed_final";
```
mkuchenbecker: Failed vs failed_final.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
```diff
@@ -0,0 +1,245 @@
...
+import org.springframework.web.reactive.function.client.WebClient;
+import org.springframework.web.reactive.function.client.WebClientRequestException;
+import org.springframework.web.reactive.function.client.WebClientResponseException;
+import reactor.core.publisher.Mono;
+import reactor.util.retry.Retry;
+import reactor.util.retry.RetrySpec;
+
+/**
+ * Pushes a stats record to the optimizer's {@code PUT /v1/optimizer/stats/{tableUuid}} endpoint
+ * after a successful Iceberg snapshot commit. Fire-and-forget — the {@link #report} call returns
```
mkuchenbecker: we should see if we get an error. this working is ambiguous and non-useful.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * <p>The bean is only wired when {@code optimizer.stats.enabled=true}.
+ */
+@Slf4j
+@Component
+@ConditionalOnProperty(prefix = "optimizer.stats", name = "enabled", havingValue = "true")
+public class OptimizerStatsClient {
+
+  /** Per-call URL path. */
+  static final String PATH_TEMPLATE = "/v1/optimizer/stats/{tableUuid}";
```
mkuchenbecker: is this the canonical way to do this?

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
```diff
@@ -0,0 +1,245 @@
...
+
+  /** Table-property key that opts a table in for the post-commit stats push. */
+  static final String OPT_IN_PROPERTY = "maintenance.optimizer.stats.enabled";
+
+  /** Iceberg snapshot-summary keys we read. All values are decimal-string longs. */
+  static final String SUMMARY_TOTAL_DATA_FILES = "total-data-files";
```
mkuchenbecker: Done. `TableDto.currentSnapshot` now carries a `CurrentSnapshotInfo(snapshotId, summary)` populated from `Snapshot.snapshotId()` and `Snapshot.summary()` at `convertToTableDto`. `OptimizerStatsRequest.Snapshot` carries `snapshotId`. The server-side dedupe wiring (use snapshot ID as an idempotency token on upsert and reject out-of-order replays) is intentionally deferred to a follow-up change so this PR stays focused on the framework + first operation.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
```diff
@@ -0,0 +1,245 @@
...
+   * <p>The call is skipped (and a {@code skipped} counter is emitted) when:
+   *
+   * <ul>
+   *   <li>The table is not opted in via {@link #OPT_IN_PROPERTY}.
+   *   <li>{@code snapshotSummary} is null or empty (no committed snapshot yet — e.g. CREATE TABLE
+   *       with no rows).
+   * </ul>
+   */
+  public void report(TableDto saved, Map<String, String> snapshotSummary) {
+    reportAsync(saved, snapshotSummary).subscribe();
```
mkuchenbecker: why async? If the thread dies we lose the data, no? There will never be a timeout ever with this impl as we don't actually. 

Help me justify this complexity given we subscribe to the result.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
```diff
@@ -0,0 +1,245 @@
...
+   * block on completion. Production callers use {@link #report}.
+   */
+  Mono<Void> reportAsync(TableDto saved, Map<String, String> snapshotSummary) {
+    if (!isOptedIn(saved)) {
+      meterRegistry
+          .counter(MetricsConstant.OPTIMIZER_STATS_SKIPPED, "reason", "opt_out")
+          .increment();
+      return Mono.empty();
+    }
+    if (snapshotSummary == null || snapshotSummary.isEmpty()) {
```
mkuchenbecker: do not use null guards if this will never be null. In the case its empty we should not sent null to this function, but detect and immedtialty conver to an optional. 

If it may be null, use optional instead. Apply this across the pr.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
mkuchenbecker: If this is fully async timeouts can be substantially relaxed. we can have a timeout on the order of 3s overall then.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
```diff
@@ -0,0 +1,245 @@
...
+      meterRegistry
+          .counter(MetricsConstant.OPTIMIZER_STATS_SKIPPED, "reason", "no_snapshot")
+          .increment();
+      return Mono.empty();
+    }
+
+    OptimizerStatsRequest body = buildRequest(saved, snapshotSummary);
+    String tableUuid = saved.getTableUUID();
+    Timer.Sample sample = Timer.start(meterRegistry);
```
mkuchenbecker: Does the closure capture this variable correctly?

services/tables/src/main/java/com/linkedin/openhouse/tables/model/TableDto.java
```diff
@@ -83,6 +84,18 @@ public class TableDto {
 
   private boolean replaceCommit;
 
+  /**
+   * Iceberg {@code Snapshot.summary()} for the current snapshot at the time the {@code TableDto}
+   * was constructed (post-commit on save; post-load on read). Populated only when an Iceberg {@code
+   * Table} is available — i.e. by {@code InternalRepositoryUtils.convertToTableDto}. Not persisted,
```
mkuchenbecker: don't reference private impls. Reference the conditions this is or is not present.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/IcebergSnapshotsServiceImpl.java
```diff
@@ -83,7 +91,14 @@ public Pair<TableDto, Boolean> putIcebergSnapshots(
...
     }
     try {
-      return Pair.of(openHouseInternalRepository.save(tableDtoToSave), !tableDto.isPresent());
+      TableDto savedDto = openHouseInternalRepository.save(tableDtoToSave);
+      // Fire-and-forget push of the post-commit snapshot summary to the optimizer. Returns
+      // immediately; failures are swallowed inside the client. Skipped at the client when the
+      // table is not opted in via maintenance.optimizer.stats.enabled or when there is no current
+      // snapshot (e.g. CREATE TABLE without a data commit).
+      optimizerStatsClient.ifPresent(
+          client -> client.report(savedDto, savedDto.getCurrentSnapshotSummary()));
```
mkuchenbecker: Gate this block with a config for the tables service so we can enable the check one openhouse instance at a time and not cause a problem.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/IcebergSnapshotsServiceImpl.java
```diff
@@ -32,6 +33,13 @@ public class IcebergSnapshotsServiceImpl implements IcebergSnapshotsService {
 
   @Autowired AuthorizationUtils authorizationUtils;
 
+  /**
+   * Present only when {@code optimizer.stats.enabled=true}. When absent, no post-commit push is
+   * attempted.
+   */
+  @Autowired(required = false)
+  Optional<OptimizerStatsClient> optimizerStatsClient = Optional.empty();
+
```
mkuchenbecker, line 42: This should be modeled as a postcommit operation. Postcommit operations must be bounded and async and are considered best-effort. We ahve one postcommit operation registered to the service which is optimizer stats. That way we have a generic behaviour and a specific impl.

services/common/src/main/java/com/linkedin/openhouse/common/metrics/MetricsConstant.java
mkuchenbecker: Renamed to `POSTCOMMIT_OP_DURATION` (tagged `op` + `outcome`). The timer is bounded by the dispatcher's outer `tables.postcommit.per-op-timeout-ms` (default 3000 ms) — comment on the constant calls that out explicitly. Latest commit.

services/common/src/main/java/com/linkedin/openhouse/common/metrics/MetricsConstant.java
mkuchenbecker: Dropped `_FINAL`. The dispatcher now emits a single `POSTCOMMIT_OP_FAILED` counter tagged with the classified outcome. There is no per-attempt failure counter at the dispatcher; per-attempt retry behavior lives inside each op and is observable through that op's own logging if it needs it.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
mkuchenbecker: Rewrote the javadoc on `OptimizerStatsPostCommitOperation` to state the contract: `prepare()` returns a Mono that completes on HTTP 2xx and signals an error otherwise; the dispatcher owns timeout, subscription, error swallowing, and metric emission. Latest commit.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
mkuchenbecker: There wasn't one. Added `TableStatsController.BASE_PATH` and `TableStatsController.TABLE_PATH_TEMPLATE` as public constants on the optimizer controller (latest commit) and refactored its own `@RequestMapping` to use `BASE_PATH`. The tables-side keeps its own literal because we don't want a tables → optimizer compile dep (optimizer pulls MySQL/JPA); the comment on the tables-side constant now names `TableStatsController.TABLE_PATH_TEMPLATE` as the canonical source.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
mkuchenbecker: Done. `TableDto.currentSnapshot` now carries a `CurrentSnapshotInfo(snapshotId, summary)` populated from `Snapshot.snapshotId()` and `Snapshot.summary()` at `convertToTableDto`. `OptimizerStatsRequest.Snapshot` carries `snapshotId`. The server-side dedupe wiring is intentionally deferred to a follow-up change so this PR stays focused on the framework + first operation.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
mkuchenbecker: Justification, on record: a synchronous push converts an optimizer outage into a tables-service write outage. The push is a best-effort scheduling signal, not a write-correctness step — its blast radius must not include the write path. Crash-loss between commit and HTTP-completion is acceptable because operations are cumulative: the next commit re-pushes the current state and the consumer self-corrects from missing data.

On complexity: the async machinery now lives in one place (`PostCommitDispatcher` — timeout, swallow, subscribe). Operations describe payload + endpoint. Future reviewers reason about async semantics once. The optimizer-stats op itself dropped to `prepare()` + `buildRequest()` + a couple of static helpers — no `.subscribe()`, no outer timeout, no `onErrorResume`.

"There will never be a timeout" — the per-op timeout still fires; it bounds Reactor pipeline runtime (resource freeing, connection pool occupancy, the duration timer) even though the commit thread is no longer watching for it. The dispatcher's `onErrorResume` records `outcome=timeout` deterministically (covered by `decorate_workExceedsPerOpTimeout_incrementsFailedWithTimeoutOutcome`).

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
mkuchenbecker: Applied. `TableDto.currentSnapshot` is now read through `getCurrentSnapshot(): Optional<CurrentSnapshotInfo>` and `InternalRepositoryUtils` converts the nullable `table.currentSnapshot()` at the producer. The operation's `prepare()` no longer has any null/empty-map guards on the summary — it does `savedDto.getCurrentSnapshot().isPresent()` and returns `Optional.empty()` if not, which the dispatcher records as `postcommit_op_skipped{op=optimizer_stats}`.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
mkuchenbecker: Relaxed. Dropped `optimizer.stats.total-timeout-ms`. The dispatcher now owns the outer ceiling via `tables.postcommit.per-op-timeout-ms` (default 3000 ms). The op keeps its per-attempt timeout (default 1000 ms) so retries can fit inside the 3 s budget.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/optimizer/OptimizerStatsClient.java
mkuchenbecker: Yes — `sample` is effectively final and captured by both lambdas; `doOnSuccess` and `onErrorResume` are mutually exclusive on a Mono, so there's no double-stop. After the refactor the `Timer.Sample` lives in `PostCommitDispatcher.decorate(...)` instead, and the same reasoning applies there. Covered by `decorate_successfulWork_recordsDurationWithSuccessOutcome` and `decorate_workSignalsError_incrementsFailedWithClassifiedOutcomeAndSwallowsError`.

services/tables/src/main/java/com/linkedin/openhouse/tables/model/TableDto.java
mkuchenbecker: Rewrote. The javadoc now describes the semantic condition: "Present whenever the underlying table has at least one committed snapshot at construction time; absent for tables with no committed data (e.g. CREATE TABLE with no rows yet)." No mention of `InternalRepositoryUtils` or any other private impl.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/IcebergSnapshotsServiceImpl.java
mkuchenbecker: Done — single gate at the tables-service level: `tables.postcommit.enabled` (default `false`). When false the `PostCommitDispatcher` bean isn't constructed; the service consumes `Optional<PostCommitDispatcher>` and the on-commit hook is a literal no-op. Per-OH-instance rollout is one config flip. The op also keeps its own `optimizer.stats.enabled` switch so we can disable one operation while leaving the framework on.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/IcebergSnapshotsServiceImpl.java
mkuchenbecker: Done. New `services/tables/.../services/postcommit/` package: `PostCommitOperation` interface (`name()` + `Optional<Mono<Void>> prepare(TableDto)`), `PostCommitProperties` (`enabled`, `perOpTimeoutMs`), `PostCommitDispatcher` (per-op timeout, error swallow, metric emission, fire-and-forget subscribe). `OptimizerStatsPostCommitOperation` is the first registered impl. `IcebergSnapshotsServiceImpl` now has one line: `postCommitDispatcher.ifPresent(d -> d.dispatch(savedDto))`. Adding a future operation = implement the interface and let `@ConditionalOnProperty` wire it; no service-layer change.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/postcommit/PostCommitDispatcher.java
```diff
@@ -0,0 +1,147 @@
...
+                  return Mono.empty();
+                });
+    return Optional.of(decorated);
+  }
+
+  /**
+   * Map a terminal error to a small set of outcome tags. Kept here so all operations share the same
+   * taxonomy.
+   */
+  private static String classifyOutcome(Throwable e) {
```
mkuchenbecker, line 118: this entire thing is stupid, remove it and report the error type / code. everyone knows 500 are sys error and 400 are user.

cbb330 (dismissed):
lets pull from the kafka events emitted from the tableauditevent or CDC stream or poll the database periodically for last-updated

otherwise we're editting core openhouse commit() logic to plug table optimizer. this doesn't look right

mkuchenbecker:
@cbb330 this is for testing. NRT stats is the official design. I put up draft PRs for review and tell claude what to fix.

PR #612, opened 2026-05-28
fix(catalog): abort doCommit on stale-base divergence

mkuchenbecker:
## Summary

Adds a catalog-level CAS in `OpenHouseInternalTableOperations.doCommit` that aborts when the writer's declared base (`COMMIT_KEY`) does not match the catalog's current persisted base — closing the `BaseTransaction.applyUpdates` silent-rebase variant of the stale-base lost-update bug (incident-12185).

## Mechanism

1. A writer's transaction stages `COMMIT_KEY = T_X` and a `SNAPSHOTS_JSON_KEY` payload computed against `T_X`.
2. A racing commit advances the catalog `T_X → T_Y`, adding a snapshot.
3. `BaseTransaction.applyUpdates` silently refreshes the in-flight base to `T_Y` and re-applies the staged update — re-stamping `COMMIT_KEY = T_X` on top of `T_Y` while leaving the stale `SNAPSHOTS_JSON_KEY`.
4. `doCommit` runs with `base = T_Y` but `COMMIT_KEY = T_X`. Without the check, the subtractive snapshot merge computes `toRemove = T_Y.snapshots() − stale payload = {racing snapshot}` and silently drops it.

The fix reads `COMMIT_KEY` before `failIfRetryUpdate` strips it, URI-normalizes both paths via Hadoop `Path`, and throws `CommitFailedException` on mismatch so Iceberg retries against the fresh base.

## Scope of the check

- **Aborts** when `COMMIT_KEY` is a concrete location differing from the catalog base, or is `INITIAL_VERSION`.
- **Does not defend** commits that leave `COMMIT_KEY` unset — wholesale replace/create (`replaceTable`, stage-create, stage-replace) are authoritative over the snapshot set, so there is no stale base to compare against.

## Changes

Bug fix + unit test, internal catalog only (2 files). `doCommit` may now throw `CommitFailedException` on a stale-base commit that previously silently dropped a racing snapshot.

## Testing

Unit test `testDoCommitMustAbortStaleBaseRebaseToPreventSnapshotLoss` in `OpenHouseInternalTableOperationsTest` round-trips the post-refresh `TableMetadata` through `TableMetadataParser` so `base.metadataFileLocation()` is non-null (matching a loaded-from-disk base) and the URI-normalized comparison runs. Asserts `CommitFailedException` is thrown and `houseTableRepository.save` is never invoked.

```
./gradlew :iceberg:openhouse:internalcatalog:test --tests "*OpenHouseInternalTableOperationsTest"
```

### Not covered

A Spark concurrent-insert behavior test (`SparkConcurrentInsertFunctionalTest`, PR #614) was explored and removed: it only reproduces against the H2 test fixture, not production MySQL+HTS. A prod-realistic black-box repro would need the real HTS app or a deployed instance.

### DIFF
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
@@ -259,6 +259,9 @@ public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperatio
     try {
       // Now that we have metadataLocation we stamp it in metadata property.
       Map<String, String> properties = new HashMap<>(metadata.properties());
+
+      abortIfWriterBaseDivergedFromCatalog(base, metadata);
+
       failIfRetryUpdate(properties);
       restoreOverriddenProperties(properties);
 
@@ -573,6 +576,55 @@ public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperatio
     }
 
     return builder.build();
+  }
+
+  /**
+   * Catalog-level CAS. Aborts the commit if the writer's declared base ({@code COMMIT_KEY}) does
+   * not match the catalog's current persisted base. Closes the BaseTransaction.applyUpdates
+   * silent-rebase variant of the stale-base bug class, where {@code applyUpdates} re-stamps the
+   * writer's original (non-null) {@code COMMIT_KEY} on top of a concurrently-advanced base.
+   *
+   * <p>Commits that leave {@code COMMIT_KEY} unset (wholesale replace / create — replaceTable,
+   * stage-create, stage-replace) are authoritative over the snapshot set and are intentionally not
+   * defended: there is no stale base to compare against.
+   *
+   * <p>Must run before failIfRetryUpdate, which strips COMMIT_KEY from the doCommit-local
+   * properties copy.
+   *
+   * @throws CommitFailedException when writer and catalog disagree on the base — retriable by
+   *     Iceberg's commit loop after the writer refreshes
+   */
+  private void abortIfWriterBaseDivergedFromCatalog(TableMetadata base, TableMetadata metadata) {
+    if (base == null || base.metadataFileLocation() == null) {
+      // No persisted catalog state to defend (initial CREATE, or mid-CREATE constructed
+      // metadata before any metadata.json has been written).
+      return;
+    }
+    if (!metadata.properties().containsKey(CatalogConstants.SNAPSHOTS_JSON_KEY)) {
+      // Not a snapshot-bearing writer commit (e.g. rename, property-only update, internal
+      // metadata-field write). These paths legitimately have no COMMIT_KEY because they don't
+      // go through OpenHouseInternalRepositoryImpl.save:187, and they don't carry the
+      // stale-snapshot-list payload that the subtractive merge would silently expire.
+      return;
+    }
+    String actualBase = base.metadataFileLocation();
+    String writerClaimedBase = metadata.properties().get(CatalogConstants.COMMIT_KEY);
+
+    if (writerClaimedBase == null) {
+      return;
+    }
+
+    if (CatalogConstants.INITIAL_VERSION.equals(writerClaimedBase)
+        || !new org.apache.hadoop.fs.Path(writerClaimedBase)
+            .toUri()
+            .getPath()
+            .equals(new org.apache.hadoop.fs.Path(actualBase).toUri().getPath())) {
+      throw new CommitFailedException(
+          "Cannot commit: writer's declared base [%s] does not match the catalog's current "
+              + "base [%s] for table %s. A concurrent commit landed between the writer's "
+              + "loadTable and commit. Refresh and retry.",
+          writerClaimedBase, actualBase, tableIdentifier);
+    }
   }
 
   /**
--- a/iceberg/openhouse/internalcatalog/src/test/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperationsTest.java
+++ b/iceberg/openhouse/internalcatalog/src/test/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperationsTest.java
@@ -29,6 +29,7 @@ import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -210,6 +211,107 @@ public class OpenHouseInternalTableOperationsTest {
 
       Assertions.assertTrue(updatedProperties.containsKey(getCanonicalFieldName("tableLocation")));
       Mockito.verify(mockHouseTableRepository, Mockito.times(1)).save(Mockito.eq(mockHouseTable));
+    }
+  }
+
+  /**
+   * Deterministic local reproduction of incident-12185 -- the {@code BaseTransaction.applyUpdates}
+   * silent-rebase variant of the stale-base lost-update class (2026-05-25 WAR fingerprint, where
+   * snapshot 3635817277608242413 was silently rebased out by a concurrent commit).
+   *
+   * <p>Reconstructed at the {@code doCommit} boundary (no threads, no deploy):
+   *
+   * <ul>
+   *   <li>A writer loaded the table at base T_X and staged a commit: {@code COMMIT_KEY=T_X} and a
+   *       {@code SNAPSHOTS_JSON} payload computed against T_X (does NOT contain the racing
+   *       snapshot).
+   *   <li>A concurrent commit landed, advancing the catalog to T_Y, which DOES contain the racing
+   *       snapshot.
+   *   <li>{@code BaseTransaction.applyUpdates} silently refreshed the in-flight base T_X -&gt; T_Y
+   *       and re-applied the staged {@code PropertiesUpdate}: {@code COMMIT_KEY} is re-stamped as
+   *       T_X on top of T_Y, while {@code SNAPSHOTS_JSON} stays the writer's stale
+   *       (racing-snapshot-free) list.
+   *   <li>{@code doCommit} then runs with {@code base=T_Y} (metadataFileLocation non-null) but
+   *       {@code COMMIT_KEY=T_X}.
+   * </ul>
+   *
+   * <p>On the unfixed catalog, {@code doCommit}'s subtractive snapshot merge computes {@code
+   * toRemove = T_Y.snapshots() - stalePayload = {racingSnapshot}} and SILENTLY drops it (no
+   * exception, {@code save()} invoked) -- the lost update. {@code doCommit} MUST instead detect
+   * that the writer's declared base ({@code COMMIT_KEY=T_X}) diverges from the actual base (T_Y)
+   * and abort with {@link CommitFailedException} so Iceberg refreshes and retries against T_Y
+   * (where a recomputed expire/append keeps the racing snapshot).
+   *
+   * <p>This test asserts the required abort, so it FAILS on the current checkout (the racing
+   * snapshot is silently dropped instead) and PASSES once the {@code doCommit} stale-base CAS (OSS
+   * PR #612) is applied.
+   */
+  @Test
+  void testDoCommitMustAbortStaleBaseRebaseToPreventSnapshotLoss() throws IOException {
+    List<Snapshot> testSnapshots = IcebergTestUtil.getSnapshots();
+    Snapshot writerKnown1 = testSnapshots.get(0);
+    Snapshot writerKnown2 = testSnapshots.get(1);
+    // Added concurrently (lands in T_Y); absent from the writer's stale payload.
+    Snapshot racingSnapshot = testSnapshots.get(2);
+
+    // The base the writer THOUGHT it was committing against (T_X) -- different from the catalog
+    // head.
+    String writerClaimedBaseLocation =
+        "/test/openhouse/test_db/test_table/00001-writer-claimed-base.metadata.json";
+
+    // Build T_Y (post-refresh base) with all three snapshots and round-trip it through
+    // TableMetadataParser so metadataFileLocation() is non-null, matching what Iceberg passes into
+    // doCommit for a base loaded from disk after applyUpdates' silent refresh.
+    java.nio.file.Path tmpDir = Files.createTempDirectory("oh-incident-12185-repro");
+    String postRefreshBasePath = tmpDir.resolve("00010-post-refresh-base.metadata.json").toString();
+    TableMetadata buildable =
+        TableMetadata.buildFrom(BASE_TABLE_METADATA)
+            .setBranchSnapshot(writerKnown1, SnapshotRef.MAIN_BRANCH)
+            .setBranchSnapshot(writerKnown2, SnapshotRef.MAIN_BRANCH)
+            .setBranchSnapshot(racingSnapshot, SnapshotRef.MAIN_BRANCH)
+            .build();
+    Path postRefreshBaseFsPath = new Path(postRefreshBasePath);
+    FileSystem fs = postRefreshBaseFsPath.getFileSystem(new Configuration());
+    try (FSDataOutputStream out = fs.create(postRefreshBaseFsPath, true)) {
+      out.write(TableMetadataParser.toJson(buildable).getBytes());
+    }
+    TableMetadata postRefreshBase =
+        TableMetadataParser.read(new HadoopFileIO(new Configuration()), postRefreshBasePath);
+    Assertions.assertNotNull(postRefreshBase.metadataFileLocation());
+
+    // The writer's stale payload: only the snapshots it knew about at T_X (racing snapshot
+    // omitted).
+    List<Snapshot> staleWriterPayload = Arrays.asList(writerKnown1, writerKnown2);
+    Map<String, String> properties = new HashMap<>();
+    properties.put(
+        CatalogConstants.SNAPSHOTS_JSON_KEY, SnapshotsUtil.serializedSnapshots(staleWriterPayload));
+    properties.put(
+        CatalogConstants.SNAPSHOTS_REFS_KEY,
+        SnapshotsUtil.serializeMap(IcebergTestUtil.createMainBranchRefPointingTo(writerKnown2)));
+    properties.put(getCanonicalFieldName("tableLocation"), TEST_LOCATION);
+    // applyUpdates re-stamped the writer's ORIGINAL claimed base here, even though base is now T_Y.
+    properties.put(CatalogConstants.COMMIT_KEY, writerClaimedBaseLocation);
+
+    TableMetadata metadata = postRefreshBase.replaceProperties(properties);
+
+    try (MockedStatic<TableMetadataParser> ignoreWriteMock =
+        Mockito.mockStatic(TableMetadataParser.class)) {
+      CommitFailedException thrown =
+          Assertions.assertThrows(
+              CommitFailedException.class,
+              () -> openHouseInternalTableOperations.doCommit(postRefreshBase, metadata),
+              "REPRO incident-12185: doCommit must abort when the writer's declared base (COMMIT_KEY) "
+                  + "diverges from the catalog's actual base. On the unfixed catalog no exception is "
+                  + "thrown and the subtractive merge silently expires the racing snapshot "
+                  + racingSnapshot.snapshotId()
+                  + " -- reproducing the lost update.");
+
+      Assertions.assertTrue(
+          thrown.getMessage().contains("Cannot commit"),
+          "Expected stale-base abort message, got: " + thrown.getMessage());
+
+      // The racing snapshot must never have been persisted out of existence.
+      Mockito.verify(mockHouseTableRepository, Mockito.never()).save(Mockito.any());
     }
   }
 


iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -582,6 +609,23 @@ static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema)
...
+   */
+  private static boolean isSameMetadataPath(String a, String b) {
+    if (java.util.Objects.equals(a, b)) {
+      return true;
+    }
+    if (a == null || b == null) {
+      return false;
+    }
+    return java.util.Objects.equals(
+        java.net.URI.create(a).getPath(), java.net.URI.create(b).getPath());
```
mkuchenbecker: This could fail, no?  URI throws on invalid but this may be path. I am checking, not saying to change it.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -265,9 +265,36 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+      if (base != null
+          && writerClaimedBase != null
+          && !CatalogConstants.INITIAL_VERSION.equals(writerClaimedBase)
+          && !isSameMetadataPath(writerClaimedBase, base.metadataFileLocation())) {
+        throw new CommitFailedException(
+            "Cannot commit: writer's declared base [%s] does not match the table's current base "
+                + "[%s] for table %s. The catalog advanced after the writer constructed its "
+                + "request. Refresh and retry.",
+            writerClaimedBase, base.metadataFileLocation(), tableIdentifier);
+      }
```
mkuchenbecker: encapsulate this and the getting of the write changed based as a failOnBaseMismatch function.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Good catch — `URI.create` throws `IllegalArgumentException` on malformed input, which would escape the commit path as an unclassified runtime error instead of a retriable `CommitFailedException`. Captured in `memory/tasks/pr-612-pending-comments.md` with a try/catch fallback to literal-equality. Will apply once the in-flight refactor of this file lands so we do not collide.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Agreed — captured in `memory/tasks/pr-612-pending-comments.md` as a `failOnBaseMismatch(base, properties, tableIdentifier)` extraction. Note the ordering constraint: the helper must run before `failIfRetryUpdate` because it reads `COMMIT_KEY` and `failIfRetryUpdate` strips it. Deferring the edit because this file is being refactored in parallel; will apply on top of the refactor.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Applied in c0ee9fdc — wrapped the `URI.create` calls in a try/catch that falls back to literal-inequality on `IllegalArgumentException`, so divergence is still reported via the caller's `CommitFailedException` (retriable) rather than escaping as an unclassified runtime error.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Applied in c0ee9fdc — pulled the `COMMIT_KEY` capture into `abortIfWriterBaseDivergedFromCatalog` itself. Signature is now `(TableMetadata base, Map<String, String> properties)`; `doCommit`'s caller is just one line. Ordering constraint (must run before `failIfRetryUpdate`) is documented on both sides.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -582,6 +589,143 @@ static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema)
...
+    if (a == null || b == null) {
+      return false;
+    }
+    // URI.create throws IllegalArgumentException on inputs that aren't RFC 2396 compliant. A
+    // malformed COMMIT_KEY from a client must not escape doCommit as an unclassified runtime
+    // error — Iceberg's BaseMetastoreTableOperations.commit retry loop only classifies
+    // CommitFailedException as retriable. Fall back to literal-equality (which already returned
+    // false above), so divergence is reported via the caller's CommitFailedException.
+    try {
+      return java.util.Objects.equals(
```
mkuchenbecker: I think this just normalizes for comparison. I don't like that we need to parse this and potentially cause an error to check equality. Path doesn't have the same validation as URI IIRC and can avoid the throw entirely.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -582,6 +589,143 @@ static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema)
...
+  }
+
+  /**
+   * True when the writer's declared base is "no existing table" — either absent in the payload, or
+   * the literal {@link CatalogConstants#INITIAL_VERSION} marker that {@code
+   * OpenHouseInternalRepositoryImpl} sets for {@code CREATE TABLE} requests. Used by {@link
+   * #abortIfWriterBaseDivergedFromCatalog(TableMetadata, TableMetadata)} to detect the
+   * racing-CREATE class of the stale-base bug.
+   */
+  private static boolean writerClaimsNoExistingBase(String writerClaimedBase) {
```
mkuchenbecker: this doesn't need to be a function.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -582,6 +589,143 @@ static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema)
...
+   *   <li><b>Racing update (the production SEV)</b>: writer claims base {@code T_X}; catalog
+   *       advances to {@code T_Y} between {@code OpenHouseInternalRepositoryImpl.save}'s {@code
+   *       catalog.loadTable()} and {@code transaction.commitTransaction()}. Iceberg's {@code
+   *       BaseTransaction.applyUpdates} silently refreshes the in-flight transaction's base to
+   *       {@code T_Y} and re-applies the staged {@code PendingUpdates}; the re-applied {@code
+   *       PropertiesUpdate} restamps the writer's stale {@code COMMIT_KEY=T_X} on top of {@code
+   *       T_Y}. Without this check, the subtractive snapshot-list merge later in {@code doCommit}
+   *       treats catalog-side snapshots not present in the writer's stale {@code
+   *       SNAPSHOTS_JSON_KEY} as deletions and silently expires them.
+   *   <li><b>Racing CREATE</b>: two writers each call {@code CREATE TABLE} on the same identifier.
```
mkuchenbecker: all the formatting `<b> <li> <ul> `is distracting on code review. bias for simple explanations with limited needed structure.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -582,6 +589,143 @@ static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema)
...
+      // existing base" (null COMMIT_KEY or COMMIT_KEY = INITIAL_VERSION) against persisted
+      // catalog state as a divergence.
+      throw new CommitFailedException(
+          "Cannot commit: writer's view shows no existing table (declared base [%s]), but the "
+              + "catalog already has state at [%s] for table %s. A concurrent commit landed "
+              + "between the writer's loadTable and commit. Refresh and retry.",
+          writerClaimedBase, actualBase, tableIdentifier);
+    }
+
+    if (!isSameMetadataPath(writerClaimedBase, actualBase)) {
```
mkuchenbecker: this function does not need to exist.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -582,6 +589,143 @@ static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema)
...
+      // is the racing-CREATE class: writer 2 committed CREATE against an empty view, but writer
+      // 1 already landed real state. Letting this through would silently overwrite writer 1's
+      // snapshots via doCommit's subtractive snapshot-list merge. We treat any claim of "no
+      // existing base" (null COMMIT_KEY or COMMIT_KEY = INITIAL_VERSION) against persisted
+      // catalog state as a divergence.
+      throw new CommitFailedException(
+          "Cannot commit: writer's view shows no existing table (declared base [%s]), but the "
+              + "catalog already has state at [%s] for table %s. A concurrent commit landed "
+              + "between the writer's loadTable and commit. Refresh and retry.",
+          writerClaimedBase, actualBase, tableIdentifier);
```
mkuchenbecker: invert this check.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/model/HouseTable.java
```diff
@@ -24,6 +25,15 @@ public class HouseTable {
 
   @Id private String databaseId;
 
+  /**
+   * Optimistic-lock version, used by Spring Data JPA's merge to detect concurrent updates against
+   * the same HouseTable row. Without this, two writers that both observed the same {@code
+   * tableLocation} at findById time can both succeed in save(), silently overwriting one writer's
+   * commit and orphaning its metadata.json on HDFS — the silent-snapshot-drop bug reproduced by
+   * SparkConcurrentInsertFunctionalTest.
```
mkuchenbecker:
```suggestion
   * commit and orphaning its metadata.json on HDFS.
```

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Switched to `org.apache.hadoop.fs.Path(...).toUri().getPath()`. Hadoop `Path` does not throw on malformed input, so the `try/catch` around `URI.create` is gone.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Inlined.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Stripped. Javadoc is now plain prose: one short paragraph on what the helper does + a one-line ordering constraint note.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: `isSameMetadataPath` is gone. The path comparison is now a single inline expression: `new Path(writerClaimedBase).toUri().getPath().equals(new Path(actualBase).toUri().getPath())`.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Inverted. The two separate "writer claims null/INITIAL → throw" and "paths differ → throw" branches collapsed into one positive check `writerBaseMatchesCatalog`; the single `if (!writerBaseMatchesCatalog) throw` covers both cases. Same semantics, one branch.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -265,6 +265,13 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
     try {
       // Now that we have metadataLocation we stamp it in metadata property.
       Map<String, String> properties = new HashMap<>(metadata.properties());
+
+      // MUST run before failIfRetryUpdate, which strips COMMIT_KEY from the proposed metadata's
+      // properties. The helper reads COMMIT_KEY off of `metadata` internally — keep the
+      // call colocated with the metadata.properties() copy above so the ordering constraint
+      // can't drift if doCommit is later refactored.
+      abortIfWriterBaseDivergedFromCatalog(base, metadata);
```
mkuchenbecker, line 263: is this even needed anymore?

mkuchenbecker: Different race window than the one `@Version` closes.

`@Version` closes the **HTS-save race**: two PUTs both observe `tableLocation = T_X` at `findById` time and both reach `save()`. The optimistic-lock UPDATE serializes them — one wins, one gets 409. This is the PR #614 scenario.

`abortIfWriterBaseDivergedFromCatalog` closes the **`BaseTransaction.applyUpdates` silent-rebase**: a single in-flight transaction refreshes mid-commit. `this.base` advances from `T_X` to `T_Y` internally, the staged `PropertiesUpdate` re-applies and restamps `COMMIT_KEY = T_X` on top of `T_Y`. By the time `doCommit` runs:

- `base.metadataFileLocation() = T_Y`
- `metadata.properties()[COMMIT_KEY] = T_X` (writers stale value)
- writers `SNAPSHOTS_JSON_KEY` payload is also stale (doesnt include the racing winners snapshot)

The subtractive snapshot-list merge in `doCommit` then silently expires the racing snapshot. Importantly, `@Version` does **not** catch this — the houseTable built after rebase has `tableVersion = T_Y` (post-rebase), which matches the HTS rows `metadataLocation = T_Y`, so the optimistic lock sees a legitimate-looking sequence and the save goes through. Data is silently lost.

This is the 2026-05-25 WAR fingerprint (`3635817277608242413 → 7701923871999342588 → 5646807110737817965`, byte-for-byte revert of `total_files_size`). The check fires on `T_X != T_Y` and throws `CommitFailedException` before the subtractive merge can do damage — Icebergs retry then refreshes and recommits cleanly.

So: defense-in-depth, distinct window from `@Version`. Keep.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -265,6 +265,7 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
     try {
       // Now that we have metadataLocation we stamp it in metadata property.
       Map<String, String> properties = new HashMap<>(metadata.properties());
+
```
mkuchenbecker, line 262:
```suggestion
```

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/model/HouseTable.java
```diff
@@ -24,6 +25,14 @@ public class HouseTable {
 
   @Id private String databaseId;
 
+  /**
+   * Optimistic-lock version, used by Spring Data JPA's merge to detect concurrent updates against
+   * the same HouseTable row. Without this, two writers that both observed the same {@code
+   * tableLocation} at findById time can both succeed in save(), silently overwriting one writer's
+   * commit and orphaning its metadata.json on HDFS.
+   */
+  @Version private Long version;
```
abhisheknath2011: Can we add the summary of offline analysis in the PR description on how adding version would help ensure working of optimistic locking? So the queries would have version clause added now.

abhisheknath2011: Ok I see in the PR description. Thanks.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -582,6 +585,44 @@ static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema)
...
+      // No persisted catalog state to defend (initial CREATE, or mid-CREATE constructed
+      // metadata before any metadata.json has been written).
+      return;
+    }
+    String actualBase = base.metadataFileLocation();
+    String writerClaimedBase = metadata.properties().get(CatalogConstants.COMMIT_KEY);
+
+    boolean writerBaseMatchesCatalog =
+        writerClaimedBase != null
+            && !CatalogConstants.INITIAL_VERSION.equals(writerClaimedBase)
```
mkuchenbecker: base non null where claimed base is null should be an error

mkuchenbecker: Split it out. `base != null` + `claimedBase == null || INITIAL_VERSION` now throws with its own message (`"catalog has persisted state at [%s] but writer claims no base (COMMIT_KEY=[%s])"`); path-mismatch is a separate throw below. Test still passes.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -582,6 +586,62 @@ static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema)
...
+    }
+    if (Boolean.parseBoolean(metadata.properties().get(CatalogConstants.IS_STAGE_CREATE_KEY))
+        || Boolean.parseBoolean(metadata.properties().get(CatalogConstants.IS_STAGE_REPLACE_KEY))
+        || Boolean.parseBoolean(
+            metadata.properties().get(CatalogConstants.IS_REPLACE_COMMIT_KEY))) {
+      // Stage-create / stage-replace / replace-commit explicitly replace the table's snapshot
+      // state wholesale — the writer is authoritative, there is no rebase to defend against.
+      // These paths don't stamp COMMIT_KEY either.
+      return;
+    }
```
mkuchenbecker: existing tests were failing without this because claimed base was null, but I don't like injecting the replace commit key into properties.

iceberg/openhouse/internalcatalog/src/test/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperationsTest.java
```diff
@@ -269,6 +366,82 @@ void testDoCommitAppendAndDeleteSnapshots() throws IOException {
     }
   }
 
+  /**
```
mkuchenbecker: test needs dedupe with above

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: according to claude this is idiomatic for catalog server communication. alternative is we allow a race on create table. It may be the simpler alternative as create table may fail for other invariants.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -575,6 +578,58 @@ static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema)
...
+      // stale-snapshot-list payload that the subtractive merge would silently expire.
+      return;
+    }
+    String actualBase = base.metadataFileLocation();
+    String writerClaimedBase = metadata.properties().get(CatalogConstants.COMMIT_KEY);
+
+    if (writerClaimedBase == null) {
+      // Wholesale replace / create commits (replaceTable, stage-create, stage-replace) leave
+      // COMMIT_KEY unset. The writer is authoritative over the snapshot set, so there is no stale
+      // base to defend against — nothing to compare.
```
mkuchenbecker:
```suggestion
```

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -575,6 +578,55 @@ static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema)
...
+    }
+    String actualBase = base.metadataFileLocation();
+    String writerClaimedBase = metadata.properties().get(CatalogConstants.COMMIT_KEY);
+
+    if (writerClaimedBase == null) {
+      return;
+    }
+
+    if (CatalogConstants.INITIAL_VERSION.equals(writerClaimedBase)
+        || !new org.apache.hadoop.fs.Path(writerClaimedBase)
```
mkuchenbecker, line 618:
```suggestion
    if (!new org.apache.hadoop.fs.Path(writerClaimedBase)
```

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -575,6 +578,54 @@ static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema)
...
+    }
+    String actualBase = base.metadataFileLocation();
+    String writerClaimedBase = metadata.properties().get(CatalogConstants.COMMIT_KEY);
+
+    if (writerClaimedBase == null) {
+      return;
+    }
+
+    if (!new org.apache.hadoop.fs.Path(writerClaimedBase)
+            .toUri()
```
mkuchenbecker:
```suggestion
    if (CatalogConstants.INITIAL_VERSION.equals(writerClaimedBase) || !new org.apache.hadoop.fs.Path(writerClaimedBase)
            .toUri()
```

cbb330 (approved):
This works as a mitigation for testing purposes though it'll need further analysis before we can deploy safely.

Short-term, I think it's worth taking a closer look at this code: https://github.com/linkedin/openhouse/pull/612/changes#diff-526cc78ff45a139113a29951f4db0a28242b9a7f3bc9caef9ceec26b04c5aeb8R331

it may point us toward a more targeted, reliable, and race-condition-free solution.

PR #616, opened 2026-05-28
Add branchRefName to TableAuditEvent

cbb330:
## Problem

The Git for Data feature introduces named Iceberg branch refs, but table operation observability is blind to them. Today `TableAuditEvent` records `currentSnapshotId` and `currentSnapshotTimestampMs` for the main branch, with no signal for *which branch* a write went to. Dali and Grid Observability cannot answer "was this commit to main or a named branch?" without ACL-gated access to current table state.

## Solution

Add `branchRefName` to `TableAuditEvent`, populated at commit time from the existing `snapshotRefs` and `jsonSnapshots` fields already present on `IcebergSnapshotsRequestBody`.

**How the committed branch is identified:** Iceberg appends snapshots chronologically, so the last entry in `jsonSnapshots` is always the newly-committed snapshot. We find the ref in `snapshotRefs` whose snapshot-id matches it — that ref is the branch being written.

This correctly handles:
- Main branch commit → `branchRefName = "main"`
- Named branch commit where main did not advance → `branchRefName = "<branch>"`
- No main ref present → `branchRefName = "<branch>"`, `currentSnapshotId = null`

**Why the last-snapshot assumption is safe:** Every request reaching `putIcebergSnapshots` has a non-empty `jsonSnapshots` — the repository layer (`doUpdateSnapshotsIfNeeded`) gates all `snapshotRefs` processing on `jsonSnapshots` being present. A ref-only update is a silent no-op that never commits, so the last-snapshot invariant holds for all reachable code paths.

## Changes

- `TableAuditEvent.java` — add `String branchRefName`
- `TableAuditAspect.java` — populate in `extractSnapshotInfo()` by matching last snapshot to its ref; no separate pass, no special-casing for main
- `IcebergSnapshotsApiHandlerAuditTest.java` — two new tests: main commit sets `"main"`, named branch commit sets the branch name; existing branch-only test updated to assert `branchRefName`
- `TableAuditModelConstants.java` — add `branchRefName("main")` to the three commit-path expected events

## Depends on

#601 (adds `tableProperties` to `TableAuditEvent`; merge that first)

### DIFF
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
@@ -371,12 +371,18 @@ public class TableAuditAspect {
     } else {
       operationType = OperationType.COMMIT;
     }
+    CreateUpdateTableRequestBody createUpdateTableRequestBody =
+        icebergSnapshotRequestBody.getCreateUpdateTableRequestBody();
     TableAuditEvent.TableAuditEventBuilder eventBuilder =
         TableAuditEvent.builder()
             .eventTimestamp(Instant.now())
             .databaseName(databaseId)
             .tableName(tableId)
-            .operationType(operationType);
+            .operationType(operationType)
+            .tableProperties(
+                createUpdateTableRequestBody == null
+                    ? null
+                    : createUpdateTableRequestBody.getTableProperties());
     extractSnapshotInfo(icebergSnapshotRequestBody, eventBuilder);
     TableAuditEvent event = eventBuilder.build();
     try {
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
@@ -17,7 +17,9 @@ import com.linkedin.openhouse.tables.api.spec.v0.response.GetTableResponseBody;
 import com.linkedin.openhouse.tables.audit.model.OperationStatus;
 import com.linkedin.openhouse.tables.audit.model.OperationType;
 import com.linkedin.openhouse.tables.audit.model.TableAuditEvent;
+import com.linkedin.openhouse.tables.config.InternalCatalogProperties;
 import java.time.Instant;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import lombok.extern.slf4j.Slf4j;
@@ -43,6 +45,8 @@ public class TableAuditAspect {
   @Autowired private ClusterProperties clusterProperties;
 
   @Autowired private AuditHandler<TableAuditEvent> tableAuditHandler;
+
+  @Autowired private InternalCatalogProperties internalCatalogProperties;
 
   /**
    * Install the Around advice for getTable() method in OpenHouseTablesApiHandler.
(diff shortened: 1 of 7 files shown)


services/tables/src/main/java/com/linkedin/openhouse/tables/audit/model/TableAuditEvent.java
```diff
@@ -40,4 +41,8 @@ public class TableAuditEvent extends BaseAuditEvent {
   private Long currentSnapshotId;
 
   private Long currentSnapshotTimestampMs;
+
+  private String branchRefName;
+
+  private Map<String, String> tableProperties;
```
mkuchenbecker, line 47: this is a bigger change than the branch ref name.

services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
```diff
@@ -529,6 +548,30 @@ protected ApiResponse<Void> auditUpdateDatabaseAclPolicies(
...
+   * Narrows the committed table properties down to the configured allowlist ({@code
+   * cluster.iceberg.tables.audit.table-properties-allowlist}). Returns {@code null} when there is
+   * nothing to emit so downstream audit handlers can skip the field entirely. Iterates the
+   * allowlist rather than the source so cost is O(|allowlist|) regardless of source size.
+   */
+  private Map<String, String> filterTableProperties(Map<String, String> source) {
+    if (source == null || source.isEmpty()) {
+      return null;
+    }
+    List<String> allowlist = internalCatalogProperties.getAudit().getTablePropertiesAllowlist();
```
mkuchenbecker, line 561: Where is this defined?

services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
```diff
@@ -529,6 +548,30 @@ protected ApiResponse<Void> auditUpdateDatabaseAclPolicies(
...
+      return null;
+    }
+    Map<String, String> filtered = new HashMap<>();
+    for (String key : allowlist) {
+      String value = source.get(key);
+      if (value != null) {
+        filtered.put(key, value);
+      }
+    }
+    return filtered.isEmpty() ? null : filtered;
```
mkuchenbecker, line 572: Optional<Set<String>> allowlist = Optional(internalCatalogProperties.getAudit().getTablePropertiesAllowlist()) // convert to set
allowlist.foreach(allowlist -> source.filter(key -> allowlist.contains(key)))

This can be way more compact I feel if we use stream filtering.

mkuchenbecker (commented):
I would update the PR to reflect table properties more strongly as that seems to be as large a change as the branch ref.

services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
```diff
@@ -378,50 +382,65 @@ protected ApiResponse<GetTableResponseBody> auditPutIcebergSnapshots(
...
+   * <p>currentSnapshotId and currentSnapshotTimestampMs track the main branch ref for backwards
+   * compatibility. They are null when main is absent from snapshotRefs.
    */
   private void extractSnapshotInfo(
       IcebergSnapshotsRequestBody requestBody,
       TableAuditEvent.TableAuditEventBuilder eventBuilder) {
     try {
       Map<String, String> snapshotRefs = requestBody.getSnapshotRefs();
-      if (snapshotRefs == null) {
+      List<String> jsonSnapshots = requestBody.getJsonSnapshots();
```
mkuchenbecker, line 419: More of a style thing, but I perfer to cast

mkuchenbecker (commented):
what is the status of this pr?

PR #406, opened 2025-12-03
Introduce branching and support for spark.wap.branch

cbb330:
## Summary
Implement Iceberg table branching functionality in OpenHouse internal catalog with comprehensive testing infrastructure.

Strategically, we trust that only a valid client performs the operation. This is safe because in fact O.S. model has the client write metadata.json and send the pointer to the catalog server. So, GlueCatalog (e.g. gluetableoperations.java), DynamoDbCatalog, jdbctableoperations.java do client-only metadata operations (no server-side validation exists). The client behavior is also covered extensively by Iceberg compatibility tests applied to openhouse.

When we trust the client, we simply serialize the client request and later deserialize it. Since TableMetadata fields lack public access and only some are exposed via the builder, we start from the base metadata, take the serialized client request, and apply it to base by comparing with request and: truncating snapshot IDs from base, adding new IDs from the request, and moving the current pointer according to the request.

how to test

oh suite tests - this branch
```
cd /Users/chbush/code/li/project_openhouse/openhouse && \
export JAVA_HOME=$(/usr/libexec/java_home -v 17) && \
./gradlew clean \
    :integrations:spark:spark-3.5:openhouse-spark-3.5-itest:catalogTest \
    --tests "com.linkedin.openhouse.spark.catalogtest.BranchTestSpark3_5" \
    :iceberg:openhouse:internalcatalog:test \
    --tests "com.linkedin.openhouse.internal.catalog.SnapshotDiffApplierTest" \
    --tests "com.linkedin.openhouse.internal.catalog.OpenHouseInternalTableOperationsTest" \
    :integrations:spark:spark-3.1:openhouse-spark-itest:catalogTest \
    --tests "com.linkedin.openhouse.spark.catalogtest.WapIdTest" \
    :apps:openhouse-spark-apps-1.5_2.12:test \
    --tests "com.linkedin.openhouse.catalog.e2e.WapIdJavaTest"
```

compatibility tests - includes this branch + new shim files, and also requires this iceberg 
https://github.com/cbb330/openhouse/tree/working-wap-compatibility
https://github.com/cbb330/iceberg/tree/test-harness
```
# 1. Clean up local maven repo to ensure fresh publish
rm -rf /Users/chbush/code/li/project_openhouse/openhouse/build && \
rm -rf /Users/chbush/.m2/repository/com/linkedin/openhouse/tables-test-fixtures* && \

# 2. Build and publish the fixtures jar (containing the new test list)
[...]

PR #620, opened 2026-05-29
 test(tables): stale-base concurrent-commit regression tests

mkuchenbecker:
## Summary

Concurrency regression tests (`StaleBaseLostUpdateTest`, H2-backed, single-JVM) for committing a snapshot set against a stale base version: a writer that still declares an older base must not silently drop a snapshot another writer committed concurrently.

## Scenarios

Each test loads a table at version L1, lets a second writer commit a new snapshot (advancing the catalog to L2), then commits a writer that still declares L1 as its base with a snapshot set computed at L1 — which omits the second writer's snapshot. The concurrently added snapshot must survive, and the stale commit must be rejected.

- `testStaleInsertDropsConcurrentDataCommitOnFreshTable` — two writers race the first insert into a fresh table.
- `testExpireSnapshotsDropsConcurrentDataCommit` — an expiration (keep only the current head) races a concurrent insert.
- `testStaleMetadataUpdateDropsConcurrentDataCommit` — a metadata change that adds no snapshot races a concurrent insert.
- `testConcurrentInsertOnPopulatedTableIsRejected` — two writers insert into a table that already holds data.

## Results across commits

| Test | before #406 `ff9a40c1` | #406 `14fe4603` | before #612 `d4fc9fe7` | #612 |
| --- | --- | --- | --- | --- |
| fresh-table insert | FAIL | FAIL | FAIL | **PASS** |
| expire vs insert | **PASS** | FAIL | FAIL | **PASS** |
| metadata-update vs insert | **PASS** | FAIL | FAIL | **PASS** |
| populated-table insert | PASS | PASS | PASS | PASS |

## Why the results differ

Before #406, the catalog rejected a stale commit that deletes the latest snapshot **only when the commit adds no snapshot of its own** (`SnapshotInspector.validateSnapshotsUpdate`, which exempts snapshot-adding commits). So:

- **expire** and **metadata-update** add no snapshot → caught before #406, then lost from #406 onward when that check was removed; #612 restores rejection. Regressions introduced by #406.
- **fresh-table insert** adds a snapshot → exempted by that check, so it was never caught and lost data on every commit before #612. Pre-existing, not a #406 regression.
- **populated-table insert** never loses data: the stale writer's appended snapshot shares a sequence number with the concurrent insert, so Iceberg's sequence-number validation rejects it on every version.

The single check that covers all of these is #612's base-divergence abort: it rejects any commit whose declared base no longer matches the catalog.

## Testing

```
./gradlew :services:tables:test --tests "*StaleBaseLostUpdateTest"
```

All four pass with #612. Each test asserts the stale commit is rejected (throws) and that the table holds exactly the prior snapshots plus the concurrently committed snapshot. Test-only; no production changes.

### DIFF
--- a/services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
+++ b/services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
@@ -41,14 +41,19 @@ import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.ContextConfiguration;
 
 /**
+ * A stale writer opens an Iceberg transaction at base L1 and stages its L1 snapshot view +
+ * {@code COMMIT_KEY=L1}, held uncommitted. A racing writer commits a new snapshot through the
+ * repository path, advancing the catalog L1 -&gt; L2. Committing the held transaction drives {@code
+ * BaseTransaction.applyUpdates} to silently refresh the in-flight base to L2 (passing the HTS
+ * optimistic-version CAS) while keeping the stale payload + {@code COMMIT_KEY=L1}; {@code doCommit}
+ * then subtracts the racing snapshot. On the buggy catalog the racing commit is silently dropped;
+ * PR #612's stale-base CAS aborts instead. Invariant: the racing data commit always survives.
  *
- * <p>A stale writer opens an Iceberg transaction at base L1 and stages its real L1 snapshot view
- * (the writer's {@code SNAPSHOTS_JSON} payload) plus {@code COMMIT_KEY=L1}, but holds it
- * uncommitted. A racing writer appends a new snapshot via the repository path, advancing the
- * catalog L1 -&gt; L2. When the held transaction is committed, {@code BaseTransaction.applyUpdates}
- * rebases the stale payload + {@code COMMIT_KEY=L1} onto L2, so {@code doCommit} would subtract the
- * racing snapshot. The invariant asserted is that the racing snapshot must survive (PR #612's
- * stale-base CAS aborts the commit; without it the racing snapshot is silently dropped).
+ * <p>Both cases here are <b>subtractive</b> stale commits — the stale payload omits the racing
+ * snapshot and does not add a new snapshot whose sequence number would collide with it. That is the
+ * shape that actually reproduced incident-12185 (an expire/optimizer commit dropping a fresh data
+ * commit). A stale writer that <i>adds</i> its own snapshot on a multi-snapshot base is instead
+ * rejected by Iceberg's snapshot sequence-number validation, so that shape cannot lose data.
  */
 @SpringBootTest
 @ContextConfiguration(initializers = PropertyOverrideContextInitializer.class)
@@ -61,30 +66,12 @@ public class StaleBaseLostUpdateTest {
 
   @Autowired Catalog catalog;
 
-  /** Base table is freshly created with zero committed data snapshots when the race begins. */
-  @Test
-  void testLostUpdateViaStagedTransactionConflictWithNoPriorData() throws Exception {
-    TableDto l1 = createTableWithCommittedDataSnapshots("conflict_staged_no_data", 0);
-    assertRacingSnapshotSurvivesStaleStagedCommit(l1);
-  }
-
   /**
-   * Table already has prior committed data history (two snapshots) before the race begins — the
-   * stale writer's base is a non-trivial, data-bearing table.
-   */
-  @Test
-  void testLostUpdateViaStagedTransactionConflictWithPriorDataSnapshot() throws Exception {
-    TableDto l1 = createTableWithCommittedDataSnapshots("conflict_staged_prior_data", 2);
-    assertRacingSnapshotSurvivesStaleStagedCommit(l1);
-  }
-
-  /**
-   * A snapshot-expiration commit racing a concurrent data insert — the production maintenance
-   * shape. The expire job stages a "keep" subset computed against base L1 (here: keep only the
-   * current head, expiring older snapshots); that subset cannot reference the concurrently-added
-   * data snapshot. When the stale expire rebases onto L2, the subtractive merge would expire the
-   * racing data commit along with the intended old snapshots. Asserts the data commit survives and
-   * the stale expire is rejected wholesale.
+   * Snapshot-expiration commit racing a concurrent data insert — the production maintenance shape,
+   * on a table with prior committed data history. The stale expire keeps only the current head
+   * (expiring older snapshots); its keep-subset cannot reference the concurrently-added data
+   * snapshot, so the subtractive merge would expire the racing data commit. Asserts the data commit
+   * survives and the stale expire is rejected wholesale.
    */
   @Test
   void testExpireSnapshotsDropsConcurrentDataCommit() throws Exception {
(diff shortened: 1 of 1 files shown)


services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
```diff
@@ -0,0 +1,288 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.ContextConfiguration;
+
+/**
+ * Deterministic H2 reproduction of the incident-12185 stale-base silent snapshot drop. Single-JVM,
+ * no threads, no docker.
```
mkuchenbecker:
```suggestion
```

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
```diff
@@ -0,0 +1,260 @@
...
+    return dto;
+  }
+
+  private static List<Snapshot> appended(List<Snapshot> existing, Snapshot extra) {
+    List<Snapshot> all = new ArrayList<>(existing);
+    all.add(extra);
+    return all;
+  }
+
+  private static List<String> snapshotJson(List<Snapshot> existing, Snapshot extra) {
```
mkuchenbecker: remove this function

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
```diff
@@ -0,0 +1,260 @@
...
+    return appended(existing, extra).stream()
+        .map(SnapshotParser::toJson)
+        .collect(Collectors.toList());
+  }
+
+  private static TableIdentifier idOf(TableDto dto) {
```
mkuchenbecker: remove this function

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
```diff
@@ -0,0 +1,260 @@
...
+    return TableIdentifier.of(dto.getDatabaseId(), dto.getTableId());
+  }
+
+  private DataFile dummyDataFile() throws Exception {
+    return IcebergSnapshotsModelTestUtilities.createDummyDataFile(
+        Files.createTempFile("stale-base-conflict-", ".orc").toString(),
+        PartitionSpec.unpartitioned());
+  }
+
+  private static Map<String, String> refs(Snapshot snapshot) {
```
mkuchenbecker: remove this function

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
```diff
@@ -0,0 +1,260 @@
...
+    Assertions.assertEquals(
+        priorSnapshotCount + 1,
+        remaining.size(),
+        "table must hold only the prior snapshots plus the concurrently committed snapshot");
+
+    cleanup(id);
+  }
+
+  /** Commits {@code jsonSnapshots} (main -&gt; {@code head}) declaring base {@code l1}'s version. */
+  private void commitThroughRepository(TableDto l1, List<String> jsonSnapshots, Snapshot head) {
```
mkuchenbecker: remove this function.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
```diff
@@ -0,0 +1,260 @@
...
+    Snapshot racing = catalog.loadTable(id).newAppend().appendFile(dummyDataFile()).apply();
+    commitThroughRepository(l1, snapshotJson(base, racing), racing);
+
+    // Evaluate the stale commit on its base version rather than short-circuit it as a duplicate.
+    clearRetryCache();
+
+    try {
+      staleTxn.commitTransaction();
+    } catch (Exception expected) {
+      LOG.info("stale commit rejected: {}", expected.toString());
```
mkuchenbecker: This should be asserted whether there is an exception.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
```diff
@@ -0,0 +1,260 @@
...
+
+    List<Snapshot> remaining = Lists.newArrayList(catalog.loadTable(id).snapshots());
+    LOG.info(
+        "remaining snapshots after stale commit={}",
+        remaining.stream().map(Snapshot::snapshotId).collect(Collectors.toList()));
+    Assertions.assertTrue(
+        remaining.stream().anyMatch(s -> s.snapshotId() == racing.snapshotId()),
+        "concurrently committed snapshot (" + racing.snapshotId() + ") must not be dropped");
+    Assertions.assertEquals(
+        priorSnapshotCount + 1,
```
mkuchenbecker: this assertion does not make sense to me. We added two snapsnhots, no? Is it because the loser fails now?

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
mkuchenbecker: Done. The table-state assertion now runs unconditionally after the commit attempt; the try/catch no longer logs, and the catch only allows the stale commit to be rejected — the assertion below is the single check whether or not it throws.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
mkuchenbecker: Right — both writers stage a snapshot, but the stale writer (the loser) is rejected, so only the concurrent insert lands; hence prior + 1. Replaced the size check with a set-equality assertion: the table must hold exactly the prior snapshots plus the concurrent snapshot, which makes it explicit that the stale writer's own snapshot is not applied.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
mkuchenbecker: Done — inlined the openHouseInternalRepository.save(...) call at its call site.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
mkuchenbecker: Done — inlined.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
mkuchenbecker: Done — inlined TableIdentifier.of(dto.getDatabaseId(), dto.getTableId()) at the call sites.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
mkuchenbecker: Done — inlined obtainSnapshotRefsFromSnapshot(SnapshotParser.toJson(...)) at the call sites.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/StaleBaseLostUpdateTest.java
```diff
@@ -0,0 +1,250 @@
...
+    // Evaluate the stale commit on its base version rather than short-circuit it as a duplicate.
+    clearRetryCache();
+
+    // The stale commit may be rejected outright; it must not succeed by dropping the concurrent
+    // snapshot. The table-state assertion below holds whether or not the commit throws.
+    try {
+      staleTxn.commitTransaction();
+    } catch (Exception rejected) {
+      // Rejection is an acceptable outcome; the table state is asserted below.
+    }
```
mkuchenbecker: Please actually assert whether or not there is an exception.

mkuchenbecker: Done. Replaced the swallow with assertThrows(CommitFailedException.class, staleTxn::commitTransaction) — the rejection is now explicitly asserted — followed by the set-equality assertion on the resulting table state.

PR #622, opened 2026-05-29
fix(catalog): exclude replica tables from stale-base abort

mkuchenbecker:
## Summary

Exempts replica tables from the `doCommit` stale-base CAS
(`abortIfWriterBaseDivergedFromCatalog`) added for incident-12185.

The CAS compares the writer's declared base (`COMMIT_KEY`) against the
catalog's current persisted base and aborts on divergence. That contract
does not hold for replica tables: the replication job replays the
primary's authoritative snapshot list wholesale rather than committing
incrementally against a loaded base, so its `COMMIT_KEY` does not encode
a "base I read" that the catalog head can be compared against. Enforcing
the CAS there would spuriously abort legitimate replication commits.

## Changes

- Early return in `abortIfWriterBaseDivergedFromCatalog` when the table
  is a replica (`openhouse.isTableReplicated=true`).
- Extract a broad `isReplicatedTable(properties)` helper (flag only, no
  version check) and refactor `isReplicatedTableCreate` to reuse it. The
  exemption uses the broad check intentionally so it covers replica
  **updates**, not just creates.

## Testing

`testDoCommitDoesNotAbortStaleBaseForReplicaTable` reuses the abort
test's `COMMIT_KEY`/base divergence setup but sets the replicated flag
and a non-initial table version (the case `isReplicatedTableCreate`
would miss); asserts `doCommit` does not throw and
`houseTableRepository.save` is invoked.

```
./gradlew :iceberg:openhouse:internalcatalog:test --tests "*OpenHouseInternalTableOperationsTest"
```

Both the new test and the original
`testDoCommitMustAbortStaleBaseRebaseToPreventSnapshotLoss` pass.

## Stacking

Stacked on top of #620 (the stale-base abort + regression tests).
Because that branch is not yet on `main`, this PR's diff includes #620's
commits until it merges — review only the replica-exemption commit here.

### DIFF
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
@@ -588,12 +588,6 @@ public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperatio
    * stage-create, stage-replace) are authoritative over the snapshot set and are intentionally not
    * defended: there is no stale base to compare against.
    *
-   * <p>Replica tables are likewise exempt: the replication job replays the primary's authoritative
-   * snapshot list wholesale rather than committing incrementally against a loaded base, so its
-   * {@code COMMIT_KEY} does not encode a "base I read" contract that the catalog head can be
-   * compared against. Enforcing the CAS there would spuriously abort legitimate replication
-   * commits.
-   *
    * <p>Must run before failIfRetryUpdate, which strips COMMIT_KEY from the doCommit-local
    * properties copy.
    *
@@ -606,7 +600,10 @@ public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperatio
       // metadata before any metadata.json has been written).
       return;
     }
-    if (isReplicatedTable(metadata.properties())) {
+    if (Boolean.parseBoolean(
+        metadata
+            .properties()
+            .getOrDefault(CatalogConstants.OPENHOUSE_IS_TABLE_REPLICATED_KEY, "false"))) {
       // Replica tables are written by the replication job, which replays the primary's
       // authoritative snapshot list wholesale. There is no meaningful writer-vs-catalog base to
       // compare, so the stale-base CAS does not apply.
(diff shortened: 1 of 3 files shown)


iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -588,6 +588,12 @@ static SortOrder rebuildSortOrder(SortOrder originalSortOrder, Schema newSchema)
    * stage-create, stage-replace) are authoritative over the snapshot set and are intentionally not
    * defended: there is no stale base to compare against.
    *
+   * <p>Replica tables are likewise exempt: the replication job replays the primary's authoritative
+   * snapshot list wholesale rather than committing incrementally against a loaded base, so its
+   * {@code COMMIT_KEY} does not encode a "base I read" contract that the catalog head can be
+   * compared against. Enforcing the CAS there would spuriously abort legitimate replication
+   * commits.
+   *
```
mkuchenbecker: This comment is out of place.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -820,14 +832,22 @@ private void updateMetadataFieldForTable(TableMetadata metadata, String tableLoc
    * @return
    */
   private boolean isReplicatedTableCreate(Map<String, String> properties) {
-    return Boolean.parseBoolean(
-            properties.getOrDefault(CatalogConstants.OPENHOUSE_IS_TABLE_REPLICATED_KEY, "false"))
+    return isReplicatedTable(properties)
```
mkuchenbecker: do not refactor, only touch tthe other function. no need to make a helper.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Removed the replica paragraph from the method javadoc in fff2299. The early-return's inline comment already documents the exemption.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
mkuchenbecker: Done in fff2299 — reverted isReplicatedTableCreate to its original form, dropped the helper, and inlined the openhouse.isTableReplicated check directly in abortIfWriterBaseDivergedFromCatalog. Only that function is touched now.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -600,6 +600,15 @@ private void abortIfWriterBaseDivergedFromCatalog(TableMetadata base, TableMetad
       // metadata before any metadata.json has been written).
       return;
     }
+    if (Boolean.parseBoolean(
+        metadata
+            .properties()
+            .getOrDefault(CatalogConstants.OPENHOUSE_IS_TABLE_REPLICATED_KEY, "false"))) {
+      // Replica tables are written by the replication job, which replays the primary's
+      // authoritative snapshot list wholesale. There is no meaningful writer-vs-catalog base to
+      // compare, so the stale-base CAS does not apply.
```
mkuchenbecker:
```suggestion
            .getOrDefault(CatalogConstants.OPENHOUSE_IS_TABLE_REPLICATED_KEY, "false"))) {
```

PR #624, opened 2026-05-30
[dataloader] Don't lower DataFusion execution batch size below its default

ShreyeshArangath:
## Summary

Stops a small read `batch_size` (e.g. 128) from lowering DataFusion's internal `datafusion.execution.batch_size` below its 8192 default. The read `batch_size` is now only ever *raised* into the execution batch size to honor a large requested batch (the case #568 fixed) — never lowered.

Rationale: lowering the engine's internal execution batch below its default can only ever split work into more, smaller batches; it never coalesces rows up, so it cannot help and can only add per-batch overhead. We therefore only raise it for the large-batch case #568 added, and otherwise leave DataFusion at its default.

> **Scope note:** this began as an investigation into a reported training perf regression (`read_batch` stalls, starved prefetch queue). Follow-up with Jonathan showed the regression **persisted even with the execution batch size pinned at 8192**, so that slowdown comes from elsewhere (the read/IO path), not this config. This PR lands as a small **correctness/hardening change**, not as the fix for that regression.

## Changes

- [ ] Internal API Changes
- [x] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

One change in `_create_transform_session`: only set `datafusion.execution.batch_size` when the read `batch_size` **exceeds** DataFusion's 8192 default; a smaller value leaves DataFusion at its default. No new public parameter (per review feedback to avoid extra config surface) — this is a clamp on the existing `batch_size`.

## Testing Done

The existing `test_split_batch_size_honored_with_transform` already covers the only behavior the dataloader exercises — a `batch_size` above the default is propagated so transforms are not fragmented — and still passes under the clamp. The added behavior is purely a guard against lowering the value below the default, which the dataloader's read path never benefits from. Full suite green: `make verify` — ruff + mypy + **281 passed**.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

Backward compatible. The only behavior change is that a read `batch_size` ≤ 8192 no longer shrinks the transform execution batch below DataFusion's default.


hungj (approved):

integrations/python/dataloader/src/openhouse/dataloader/data_loader.py
```diff
@@ -145,6 +146,10 @@ def __init__(
                 Passed to PyArrow's Scanner which produces batches of at most this many
                 rows. Smaller values reduce peak memory but increase per-batch overhead.
                 None uses the PyArrow default (~131K rows).
+            transform_batch_size: Rows per RecordBatch DataFusion targets while executing a
```
robreeves: I think we can do this without a new configuration. Instead we can check if the existing `batch_size` param is less than the default DF batch size `_DATAFUSION_DEFAULT_BATCH_SIZE`. If it is dont update `datafusion.execution.batch_size`.

robreeves (commented):
> Setting it to 128 turns one UDF call over an 8192-row batch into 64 calls over 128-row slices. The per-row work is identical; you just pay the fixed per-batch / per-UDF-call overhead (Python<->Arrow boundary, allocation, poll cycle) 64x more.

When we saw the regression we had `batch_size=128`. So the number of UDF calls should not have changed because `arrow_batch_size / df_batch_size` is <= 1 with both the default DF 8192 and setting the DF batch size to 128.

ShreyeshArangath:
> When we saw the regression we had batch_size=128. So the number of UDF calls should not have changed because arrow_batch_size / df_batch_size is <= 1 with both the default DF 8192 and setting the DF batch size to 128.

You are right, I checked with Jonathan and we tried saw regressions even with this fix; indicating the slowness likely comes from elsewhere. I'll update the PR description to reflect this. That being said, I think it might still be beneficial for us do the value clamping

robreeves (approved):

PR #627, opened 2026-06-01
(wip | no review) InHouse Replication

mkuchenbecker:
## (WIP — do not review)

**InHouse Replication** — a PoC for replacing GaaS (Gobblin-as-a-Service) carbon-copy replication with a
plain Spark Scala function that runs inside our own house: an **eager, full-history (retention-bounded),
physical** copy of an OpenHouse table to a **renamed** destination, committed through the normal OpenHouse
catalog path, with **snapshot-based recovery**.

This is exploratory work in an isolated worktree; it is **not** for review or merge yet.

### What's here
- `replication-poc/{PLAN,CHECKLIST,NOTES}.md` — decision trail, phased plan (P0–P2), iteration loop, oracle.
- `infra/recipes/docker-compose/oh-hadoop-spark/`:
  - `replicate.scala` — the replication function + an in-script P0/P1/P2 scenario and a time-travel oracle.
  - `run_replicate.sh`, `seed_sources.scala` — harness (spark-shell `-i` against the local `openhouse` catalog).
  - `docker-compose.yml` — mounts the recipe dir into `spark-master` so the scala iterates live.

### Approach (decided; rationale in the docs)
- Eager physical copy → self-contained replica; serves DR / read-locality / HA from one mechanism.
- Full history, retention-bounded (floor = full materialization, then per-snapshot deltas) for point-in-time
  recovery against corruption-between-runs.
- No file/UUID/snapshot-ID identity: copy files to new dest paths, relocate via Iceberg API carrying source
  metrics (no re-stat), stamp `source-snapshot-id` in each dest snapshot summary (recovery index + reconcile).

### Status (validated against local Dockerized OpenHouse)
- ✅ Gate: OH accepts a catalog commit referencing files we placed (relocated DataFile, carried metrics).
- ✅ P0 full-history replication; oracle `dest AS OF i == src AS OF i` at every snapshot.
- ✅ P1 idempotent re-run (no-op) + incremental (only-new replays).
- ✅ P2 copy-on-write delete / overwrite / compaction (OH default delete mode is CoW here).
- ⬜ Remaining: P2 MOR positional deletes (force merge-on-read; rewrite delete-file internal paths) +
  equality deletes + schema/partition-spec evolution; P1 resume-after-partial-copy.

Background: GaaS analysis in `~/code/docs/gaas-analysis/` (out-of-repo).

### DIFF
--- a/infra/recipes/docker-compose/oh-hadoop-spark/replicate.scala
+++ b/infra/recipes/docker-compose/oh-hadoop-spark/replicate.scala
@@ -1,0 +1,80 @@
+// Replication PoC — runs in spark-shell against the local OpenHouse catalog (`openhouse`).
+// Loaded via: docker exec local.spark-master /var/config/run_replicate.sh
+//
+// STATUS: SPIKE first (gate). Full `replicateTable` is outlined at the bottom as the next work.
+// The spike answers the one load-bearing unknown: does OpenHouse accept a committed snapshot whose
+// DataFiles point at paths WE placed (relocated, metrics carried), via the normal Iceberg commit path?
+
+import org.apache.hadoop.fs.{FileSystem, FileUtil, Path}
+import org.apache.iceberg.{DataFile, DataFiles, Table}
+import org.apache.iceberg.spark.Spark3Util
+import scala.collection.JavaConverters._
+
+val DB = "db"
+val SRC = s"openhouse.$DB.src_spike"
+val DST = s"openhouse.$DB.dst_spike"
+
+def loadTbl(name: String): Table = Spark3Util.loadIcebergTable(spark, name)
+
+// ---- spike ---------------------------------------------------------------------------------------
+def spike(): Unit = {
+  // fresh state each run (reset without tearing down the stack)
+  spark.sql(s"DROP TABLE IF EXISTS $DST")
+  spark.sql(s"DROP TABLE IF EXISTS $SRC")
+
+  spark.sql(s"CREATE TABLE $SRC (id INT, data STRING) USING iceberg")
+  spark.sql(s"INSERT INTO $SRC VALUES (1,'a'),(2,'b'),(3,'c')")
+
+  val src = loadTbl(SRC)
+  val srcSnap = src.currentSnapshot()
+  val srcFiles: List[DataFile] = src.newScan().planFiles().iterator().asScala.map(_.file()).toList
+  println(s"[spike] source snapshot=${srcSnap.snapshotId} dataFiles=${srcFiles.size}")
+
+  // empty dest with the same schema (rename: different table id)
+  spark.sql(s"CREATE TABLE $DST (id INT, data STRING) USING iceberg")
+  val dst = loadTbl(DST)
+  val hconf = spark.sessionState.newHadoopConf()
+
+  // copy each source data file to a remapped path under the dest table location; relocate the DataFile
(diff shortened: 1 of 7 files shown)


infra/recipes/docker-compose/oh-hadoop-spark/replicate.scala
```diff
@@ -0,0 +1,132 @@
...
+    case ds  => val o = dst.newOverwrite(); adds.foreach(o.addFile); ds.foreach(o.deleteFile); o
+  }
+  stampCommit(op, s, source)
+  val kind = if (dels.isEmpty) "append" else "overwrite"
+  s"$kind +${adds.size} -${dels.size}"
+}
+
+def ensureDest(source: String, dest: String): Table =
+  Try(loadTbl(dest)).getOrElse {
+    spark.sql(s"CREATE TABLE $dest (${spark.table(source).schema.toDDL}) USING iceberg")
```
mkuchenbecker, line 72: This isn't testing openhouse. This is testing iceberg. Are you actually following the openhouse guide for how to create openhouse tables?

infra/recipes/docker-compose/oh-hadoop-spark/replicate.scala
```diff
@@ -0,0 +1,132 @@
...
+
+val DB  = "db"
+val SRC = s"openhouse.$DB.p0_append"
+val DST = s"openhouse.$DB.p0_replica"
+
+def loadTbl(n: String): Table = Spark3Util.loadIcebergTable(spark, n)
+
+// committed history, oldest -> newest (walk parentId from current)
+def chainOf(t: Table): List[Snapshot] = {
+  def walk(s: Snapshot, acc: List[Snapshot]): List[Snapshot] =
```
mkuchenbecker, line 23: walking snapshots can get expensive. really expensive. lets add a limit, configured to default=3. If greater than limit, it makes the final as a single commit. That way you can say (1) to just get latest or 1000 if you want 1000, but "all" is not an option. 

This is because it will be run on the driver is my expectation so our scalability is limited to the number os snapshots. This also prevents malformed tables with too many snopshots from exploding the tool. 

If we could make this run on executors, that would be an option.

infra/recipes/docker-compose/oh-hadoop-spark/replicate.scala
```diff
@@ -0,0 +1,132 @@
...
+    if (s == null) acc else walk(Option(s.parentId()).map(t.snapshot(_)).orNull, s :: acc)
+  walk(t.currentSnapshot(), Nil)
+}
+def rowsAsOf(tbl: String, snapId: Long): List[String] =
```
mkuchenbecker, line 27: no helper

infra/recipes/docker-compose/oh-hadoop-spark/replicate.scala
```diff
@@ -0,0 +1,132 @@
...
+  spark.read.format("iceberg").option("snapshot-id", snapId).load(tbl).collect().map(_.toString).sorted.toList
+
+// copy one source data file to a remapped path under dest's location; relocate the DataFile (metrics carried)
+def relocate(dst: Table, df: DataFile, hconf: org.apache.hadoop.conf.Configuration): DataFile = {
```
mkuchenbecker, line 31: no helper

mkuchenbecker (commented):
any function only called 1-2 times does not need a helper. It decreases readability.

PR #636, opened 2026-06-22
[default values] Send client version via User-Agent header

cbb330:
## Problem

We're rolling out **strip protection** for default-value tables: the server blocks commits that would silently drop a column's declared default. The guard decides from the committed schema — **not** from the client version.

To expand it safely from a pilot group to the whole datalake, we need to watch which client versions are committing, to track two signals:
- failures from non-supported clients hitting the guard;
- commits from non-supported clients that slip *through* it.

The version is observability only; it never gates a commit.

## This PR (observability only)

This does **not** prevent stripping and is **not** an input to any guard; it makes the client version reliable to read so those two signals can be measured. The OpenHouse Java client now sets an explicit `User-Agent: openhouse-java-client/<version>` on every outbound request, set once in the shared `WebClientFactory.getWebClient()` so the tables, jobs, and housetables clients are all covered. Version resolves:
1. an explicit `setClientVersion(...)` override,
2. the secureclient jar manifest `Implementation-Version`,
3. `"unknown"`.

`build.gradle` stamps `Implementation-Version` into the jar manifest so the fallback is meaningful in packaged builds.

The `OpenHouseCatalog` now also **reads a `client-version` catalog property and calls `setClientVersion(...)`** (next to its existing `client-name` handling), so consumers set the version via a property rather than touching the relocated/`compileOnly` factory directly. li-openhouse#2147 sets that property from the build-time-baked runtime version.

The `OpenHouseCatalog` also **reads a `client-version` catalog property and calls `setClientVersion(...)`** (next to its existing `client-name` handling), so consumers set the version via a property instead of touching the relocated/`compileOnly` factory directly. li-openhouse#2147 sets that property from the build-time-baked runtime version.

## Why not rely on the transport's default

Previously these requests carried reactor-netty's *default* `User-Agent: ReactorNetty/<version>`, where `<version>` is `getImplementationVersion()` on whatever jar provides reactor-netty's classes. That depends entirely on how the OpenHouse client is packaged, so one token means different things:
- **thin Java runtime** (reactor-netty resolved as a normal dependency, not shaded) → reactor-netty's own *transport* version (e.g. `1.0.x`) — **never** the OpenHouse client's;
- **shaded Spark runtime** (reactor-netty bundled un-relocated into the runtime fat jar) → that fat jar's `Implementation-Version`, i.e. the *runtime's* version — or, if the fat jar carries no version, **nothing** (`ReactorNetty/dev`).

So you can't tell what a `ReactorNetty/<version>` even refers to, and the thin runtime never reports the OpenHouse version at all. Setting our own `openhouse-java-client/<version>` User-Agent makes it explicit and unambiguous across **both** packagings. For shaded runtimes, callers should additionally pass an explicit `setClientVersion(...)`, since jar-manifest introspection is unreliable once shaded.

## Testing

- Added `TablesApiClientFactoryTest.testSetClientVersionCalled`.
- `:client:secureclient:test` and `:client:secureclient:spotlessCheck` pass.

## Follow-ups (separate changes)

- Consume the header server-side (metric tag / per-commit audit dimension).
- Populate `X-Client-Name` (engine/app), which the server already reads for the `client_name` metric.

### DIFF
--- a/client/secureclient/build.gradle
+++ b/client/secureclient/build.gradle
@@ -17,6 +17,14 @@ dependencies {
 
 }
 
+// Stamp the project version into the jar manifest so WebClientFactory can advertise the client
+// version in the User-Agent header by reading Implementation-Version at runtime.
+jar {
+    manifest {
+        attributes('Implementation-Version': project.version)
+    }
+}
+
 test {
     useJUnitPlatform()
 }
--- a/client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
+++ b/client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
@@ -12,6 +12,7 @@ import javax.net.ssl.SSLException;
 import lombok.NonNull;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.HttpHeaders;
 import org.springframework.http.client.reactive.ReactorClientHttpConnector;
 import org.springframework.web.reactive.function.client.WebClient;
 import reactor.netty.http.client.HttpClient;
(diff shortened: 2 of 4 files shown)


shanthoosh (approved):
Looks pretty good overall. 

Minor clarification: Have we locally tested this patch with OH image simulating the table OPs

PR #639, opened 2026-06-23
Add clientUserAgent and snapshotSummary to the table audit event

cbb330:
## Summary

Adds two purely-additive fields to `TableAuditEvent` for client/commit observability, populated server-side in `TableAuditAspect` (both are **observability only** — neither gates a request):

- **`clientUserAgent`** — the raw `User-Agent` request header, stored verbatim, so the client/runtime version can be derived at query time (e.g. `openhouse-java-client/<version>`). Set in the shared `buildAndSendEvent`, read from the bound servlet request via `RequestContextHolder` (mirroring `ServiceAuditAspect`); best-effort and never disrupts the audited operation.
- **`snapshotSummary`** — the full Iceberg summary map of the snapshot the `main` ref points to (the same snapshot as `currentSnapshotId`), emitted verbatim (`operation`, `spark.app.id`/`trino_query_id`, engine/iceberg versions, and the core-computed file-delta counts). Emitted unaggregated so queries derive operation and write-mode signals downstream without a lossy server-side classification — there is no per-commit field that cleanly declares copy-on-write vs merge-on-read across engines. `Snapshot.summary()` omits the `operation` key (Iceberg parses it out into a separate field), so it is merged back to mirror the on-disk summary. It is hooked onto the snapshot `extractSnapshotInfo` already resolves for `currentSnapshotTimestampMs`; null when the request carries no `main` ref (branch-only commits), consistent with `currentSnapshotId`.

## Changes

- [x] Internal API Changes
- [x] New Features
- [x] Tests

`TableAuditEvent` gains `clientUserAgent` (`String`) and `snapshotSummary` (`Map<String,String>`). `snapshotSummary` is emitted unbounded by design (unlike the allowlist-filtered `auditedTableProperties`) — a snapshot summary is naturally a small, fixed set of keys; `clientUserAgent` is captured on every audited event, not just commits.

## Testing Done

`IcebergSnapshotsApiHandlerAuditTest`:
- `testPutIcebergSnapshotsCapturesClientUserAgent` — sets a `User-Agent` header and asserts verbatim capture.
- `testPutIcebergSnapshotsEmitsRewriteSummaryVerbatim` / `testPutIcebergSnapshotsEmitsDeleteFileSummaryVerbatim` — assert the raw summary counters are emitted for a copy-on-write overwrite and a merge-on-read delete.
- Extended `testPutIcebergSnapshotsContainsSnapshotInfo` and the failed-path test to assert the summary is captured (including on failure), and the branch-only test to assert it is null when there is no `main` ref.

`snapshotSummary` is excluded from the `ReflectionEquals` expected constants and asserted explicitly. Full audit test suite passes (`IcebergSnapshotsApiHandlerAuditTest` 16, plus `TablesApiHandlerAuditTest` / `DatabasesApiHandlerAuditTest`).

# Additional Information

Part of a sliced rollout of server-side commit-audit observability. A follow-up will add `failureType` (the exception class on failed commits) plus isolation of the audit emission so it can never mask the operation being audited.

### DIFF
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
+import javax.servlet.http.HttpServletRequest;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.iceberg.Snapshot;
 import org.apache.iceberg.SnapshotParser;
@@ -36,7 +37,11 @@ import org.aspectj.lang.ProceedingJoinPoint;
 import org.aspectj.lang.annotation.Around;
 import org.aspectj.lang.annotation.Aspect;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpHeaders;
 import org.springframework.stereotype.Component;
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
 
 /**
  * Aspect class to support table operation auditing for all controllers. It enhances the ability of
@@ -59,6 +64,12 @@ public class TableAuditAspect {
    * config is bound once at startup, so a one-shot lazy init is sufficient.
    */
   private volatile List<Pattern> allowlistPatterns;
+
+  // Iceberg serializes a snapshot's operation as a separate top-level field rather than inside the
+  // summary map (SnapshotParser parses it out into Snapshot#operation), so there is no public
+  // constant for this key. We merge it back into the emitted summary under this name to mirror the
+  // on-disk JSON.
+  private static final String SNAPSHOT_OPERATION_KEY = "operation";
 
   /**
    * Install the Around advice for getTable() method in OpenHouseTablesApiHandler.
@@ -455,6 +466,7 @@ public class TableAuditAspect {
         Snapshot snapshot = SnapshotParser.fromJson(snapshotJson);
         if (snapshot.snapshotId() == mainSnapshotId) {
           eventBuilder.currentSnapshotTimestampMs(snapshot.timestampMillis());
+          eventBuilder.snapshotSummary(buildSnapshotSummary(snapshot));
           return;
         }
       }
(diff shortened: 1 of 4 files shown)


services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
```diff
@@ -662,7 +689,28 @@ private void buildAndSendEvent(
...
+    try {
+      RequestAttributes requestAttributes = RequestContextHolder.getRequestAttributes();
+      if (requestAttributes instanceof ServletRequestAttributes) {
+        HttpServletRequest request = ((ServletRequestAttributes) requestAttributes).getRequest();
+        return request.getHeader(HttpHeaders.USER_AGENT);
+      }
+    } catch (Exception e) {
+      log.warn("Failed to read User-Agent header for audit event", e);
+    }
+    return null;
```
mkuchenbecker, line 714: nit. Optional vs null.

services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
```diff
@@ -464,6 +476,21 @@ private void extractSnapshotInfo(
     }
   }
 
+  /**
+   * Builds the emitted snapshot summary: the snapshot's own summary map plus its operation (which
+   * Iceberg parses out of the summary into a separate field), mirroring the on-disk JSON.
+   */
+  private static Map<String, String> buildSnapshotSummary(Snapshot snapshot) {
+    Map<String, String> summary = new HashMap<>();
+    if (snapshot.summary() != null) {
```
mkuchenbecker, line 485: under what conditions is the summary / operation expected to be null? is a partial version of this considered valid?

services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
```diff
@@ -464,6 +476,21 @@ private void extractSnapshotInfo(
...
+      summary.putAll(snapshot.summary());
```
mkuchenbecker, line 486: Seems fine. The contents do not contain column statistics that would subject this to privacy risks.

services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
```diff
@@ -464,6 +476,21 @@ private void extractSnapshotInfo(
...
+    }
+    if (snapshot.operation() != null) {
+      summary.put(SNAPSHOT_OPERATION_KEY, snapshot.operation());
```
mkuchenbecker, line 489: is empty operation / summary tested? I want to understand the expectation when one or both is actually null given the code.

services/tables/src/main/java/com/linkedin/openhouse/tables/audit/TableAuditAspect.java
```diff
@@ -455,6 +466,7 @@ private void extractSnapshotInfo(
         Snapshot snapshot = SnapshotParser.fromJson(snapshotJson);
         if (snapshot.snapshotId() == mainSnapshotId) {
           eventBuilder.currentSnapshotTimestampMs(snapshot.timestampMillis());
+          eventBuilder.snapshotSummary(buildSnapshotSummary(snapshot));
```
mkuchenbecker, line 469: I assume this is fast, <1ms because its a pure function and we are already emitting the data.

PR #644, opened 2026-06-28
[default values] Tables: add a server-stamped per-table config map to GetTableResponseBody and capture it client-side

cbb330:
## Summary

Adds a generic, server-stamped, per-table client **`config`** map to `GetTableResponseBody` and captures it in `OpenHouseTableOperations` so subclasses can read it. It follows the **Iceberg REST `LoadTableResponse.config` convention** — a `Map<String,String>` of client-side behavior overrides the server controls at runtime, without re-rolling the slow-to-upgrade Java client fleet. **No behavior change**: the map is null/empty until a server stamps it, and the base client has nothing to act on.

## Changes

- [x] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [x] Tests

**Client-facing API Changes**
- READ_ONLY, nullable `config` field (`Map<String,String>`) on `GetTableResponseBody`, modeled on the Iceberg REST load-table `config`. Namespaced keys (e.g. `openhouse.read-bridge`); clients ignore keys they do not understand.
- Why a response field: `doRefresh` already fetches a live `GetTableResponseBody` on every table load (and on commit responses) but discarded all but `getTableLocation()` — it is the natural, already-present, server-controlled, zero-staleness delivery channel. The generated client sets `FAIL_ON_UNKNOWN_PROPERTIES = false`, so a new response field cannot break older clients (and unknown config keys are simply carried). Same additive, nullable, READ_ONLY pattern as `sortOrder` etc.

**Internal API Changes**
- `OpenHouseTableOperations.doRefresh` keeps the full response and stashes `config` in an `AtomicReference`, exposed to subclasses via the new `protected currentConfig()`. READ_ONLY + side-channel: never sent back on writes.
- `TablesMapper` (the table DTO serializer into gettableresponsebody) ignores `config` (`@Mapping(target = "config", ignore = true)`) — it is stamped separately, not sourced from `TableDto`.

**New Features**
- A flat string map keyed by namespaced keys means new features become new `config` entries rather than an API/schema change or client regen.

## Testing Done
- [x] Added new tests.

Compile/codegen verified (`:services:tables`, `:client:tableclient`, `:integrations:java:iceberg-1.2:openhouse-java-runtime`); the client regenerates with `getConfig()` returning `Map<String,String>`. `OpenHouseTableOperationsTest` covers: `currentConfig()` null before refresh; `doRefresh` captures/clears `config`; the REST-style string map deserializes from a response and tolerates unknown fields.

## Stack
**#644 (channel)** → #645 (read-bridge mechanism) → [li-openhouse #2166](https://github.com/linkedin-multiproduct/li-openhouse/pull/2166) (li column-default source from avro.schema.literal)

This is the substrate PR — intentionally behaviorless on its own.

## Next steps (this PR)
- [ ] Review + merge first — it is the base of the stack; #645 and #2166 depend on it.

This PR delivers only the channel; it stays inert until two further pieces exist, both **required**, and both delivered by the next PR in the stack:
- **A server stamp** to populate `config` (it is null until something stamps it) — required, delivered by #645 via `ReadBridgeConfigResolver` + the `ColumnDefaultsSource` seam.
- **A client consumer** that reads `currentConfig()` and acts on it — required, delivered by #645's read-bridge metadata overlay, driven by the column defaults that #2166 supplies.


integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -98,12 +113,16 @@ public void doRefresh() {
                 WebClientRequestException.class,
                 e -> Mono.error(new WebClientRequestWithMessageException(e)))
             .blockOptional();
+    // Capture the server-stamped per-table config so subclasses can gate read-time behavior via
+    // currentConfig(); absent => null. Side-channel only: never sent back on writes.
+    this.config.set(tableResponse.map(GetTableResponseBody::getConfig).orElse(null));
+    Optional<String> tableLocation = tableResponse.map(GetTableResponseBody::getTableLocation);
```
shanthoosh, line 119: We update config before refreshFromMetadataLocation(). If metadata refresh subsequently fails (corrupt metadata, I/O failure, etc.), the operation throws but currentConfig() has already been updated. That leaves the object with config corresponding to a metadata snapshot that was never successfully loaded. Should the config update happen only after refreshFromMetadataLocation() completes successfully

cbb330: config is only consumed during the metadata load itself. So if config is set but the metadata load fails, the overlay is thrown away and neither the stale config nor the failed metadata has any effect. The next refresh re-fetches and re-sets config alongside the metadata, so they re-sync.

shanthoosh (approved):

PR #645, opened 2026-06-28
[default values] Tables: apply server-supplied per-column defaults at read time

cbb330:
## Summary
Adds the open-source read-bridge feature on top of the per-table `config` channel (#644): the OH server stamps per-column initial-defaults onto `GetTableResponseBody.config`, and the Java client overlays them at metadata-load time so a column added after data exists reads its declared default instead of NULL (a v2→v3 read-time bridge).

The whole feature is open source; it exposes exactly **one** deployment seam:
- **`ColumnDefaultsSource`** (interface) — the single open-source/closed-source line: `field-id -> Iceberg single-value JSON`. The open-source default is a no-op lambda in `ApiConfig`; a deployment overrides this bean (li-openhouse derives values from `avro.schema.literal`).
- **`ReadBridgeConfigResolver`** (server) — stamps each default as a flat, namespaced config entry: `openhouse.read-bridge.column-default.<fieldId> = <single-value-json>`. No envelope/POJO — the REST `config` string map carries the structure directly.
- **`ReadBridge`** (client) — decodes those entries and applies the read-time overlay in `OpenHouseTableOperations.loadMetadata` (the `withInitialDefault`/`withSchemaOverlay` transform is a marked TODO). `loadMetadata` stays one delegating call.

## Scope (engines)
- **Spark 3.1 (iceberg-1.2) and Spark 3.5 (iceberg-1.5): both covered by this PR.** The client code lives under `integrations/java/iceberg-1.2/openhouse-java-runtime`, but it is not 1.2-only: the 1.5 runtime compiles the same source, because its `build.gradle` adds the 1.2 module's source dirs to its own sourceSet (`srcDirs += project(':integrations:java:iceberg-1.2:openhouse-java-runtime').sourceSets.main.java.srcDirs`). So `ReadBridge` and the `loadMetadata` hook are built into both runtimes, and there is no separate 1.5 port. Neither line applies Iceberg v3 `initial-default` natively, so both need the bridge.
- **Flink: separate follow-up PR** (if/when a Flink read path needs the overlay).

## Stack
#644 (config channel, merged) → **#645 (read-bridge mechanism)** → [li-openhouse #2204](https://github.com/linkedin-multiproduct/li-openhouse/pull/2204) (the `ColumnDefaultsSource` seam) → [li-openhouse #2203](https://github.com/linkedin-multiproduct/li-openhouse/pull/2203) (derive the defaults from `avro.schema.literal`)

Rebased onto `main` now that #644 has merged, so the diff is the read-bridge change only.

## Next steps (this PR)
- [ ] **Bump the linkedin/iceberg forks so the overlay APIs exist on both lines.** Verified with `javap` against the pinned jars: `TableMetadata.withSchemaOverlay` is absent from `1.2.0.19`, and `1.5.2.15` has neither `withSchemaOverlay` nor `NestedField.initialDefault` (the backport #642 brought to the 1.2 line). Nothing in this PR uses them — the decode path is only Jackson + `TableMetadata` — but `ReadBridge.apply` cannot be implemented until they land.
- [ ] Implement the overlay in `ReadBridge.apply`. It must cover **every** schema-id in `schemasById` that carries a bridged field-id, not just the current schema: Iceberg resolves a scan's schema from the snapshot's own `schemaId` (`SnapshotUtil.schemaFor`), so time-travel, tag, and non-main-branch reads would otherwise return NULL while latest reads return the default. `withSchemaOverlay` takes a multi-schema map for exactly this.

## Testing Done
- `:services:tables` + iceberg-1.2 runtime compile (JDK 17), and repo-wide `spotlessCheck` green.
- `ReadBridgeConfigResolverTest` (server): each default round-trips as a flat `openhouse.read-bridge.column-default.<fieldId>` entry; empty when the source supplies nothing.
- `ReadBridgeTest` (client): decodes the flat entries by field-id; **fails loud** (`IllegalStateException`, with the offending `key=value`) on a malformed *known* `column-default.*` entry, since the server encoder guarantees an int field-id and a value that round-trips through `readTree` — so a decode failure is an encoder bug or transport corruption, not an expected input. *Unknown* keys are ignored, preserving forward compatibility.
- `OpenHouseTableOperationsTest`: config capture and deserialization.

### DIFF
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
@@ -128,11 +128,13 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
   }
 
   /**
-   * Loads the table metadata at the given location. Defaults to the stock parser; subclasses may
-   * override to transform the metadata (e.g. attach column defaults) as it loads.
+   * Loads the table metadata from storage, then runs the read-time {@link ReadBridge} over it using
+   * the per-table behavior the server stamped onto {@link #currentConfig()}. Fail-closed:
+   * absent/off/unparseable config leaves the raw metadata untouched.
    */
   protected TableMetadata loadMetadata(String metadataLocation) {
-    return TableMetadataParser.read(io(), metadataLocation);
+    TableMetadata raw = TableMetadataParser.read(io(), metadataLocation);
+    return ReadBridge.apply(raw, currentConfig());
   }
 
   @Override
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
@@ -129,8 +129,9 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
 
   /**
    * Loads the table metadata from storage, then runs the read-time {@link ReadBridge} over it using
-   * the per-table behavior the server stamped onto {@link #currentConfig()}. Fail-closed:
-   * absent/off/unparseable config leaves the raw metadata untouched.
+   * the per-table behavior the server stamped onto {@link #currentConfig()}. Absent config, or
+   * config carrying nothing to bridge, leaves the raw metadata untouched; a malformed known entry
+   * fails loud (see {@link ReadBridge}).
    */
   protected TableMetadata loadMetadata(String metadataLocation) {
     TableMetadata raw = TableMetadataParser.read(io(), metadataLocation);
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
@@ -6,7 +6,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
-import lombok.extern.slf4j.Slf4j;
 import org.apache.iceberg.TableMetadata;
 
 /**
@@ -19,11 +18,19 @@ import org.apache.iceberg.TableMetadata;
  * ReadBridgeConfigResolver} (services/tables). The contract is flat, namespaced config keys (no
  * envelope/POJO): {@code openhouse.read-bridge.column-default.<fieldId> = <single-value-json>}.
  *
- * <p>Kept out of {@link OpenHouseTableOperations} so the read path stays slim. Pure and
- * fail-closed: an entry with a non-integer field-id or an unparseable value is skipped, and nothing
- * to apply returns {@code raw} unchanged.
+ * <p>Kept out of {@link OpenHouseTableOperations} so the read path stays slim. A read-bridge entry
+ * is produced by the server encoder ({@code ReadBridgeConfigResolver}) from typed {@code JsonNode}s
+ * keyed by integer field-id, so its value always round-trips through {@code readTree} and its
+ * suffix always parses as an int. A decode failure on a *known* entry is therefore a bug or
+ * transport corruption, not an expected runtime state, and this fails loud rather than silently
+ * degrading to NULL. An *unknown* key (a newer server feature this client doesn't recognize) is
+ * ignored, preserving forward compatibility. With nothing to apply, {@code raw} is returned
+ * unchanged.
+ *
+ * <p>Note this guarantees only that a stamped value is <em>well-formed</em>, not that it is the
+ * <em>correct</em> default for its column — that semantic (default-to-schema) consistency is a
+ * write-time concern owned by whatever server path sources the defaults.
  */
-@Slf4j
 final class ReadBridge {
 
   /** Mirror of {@code ReadBridgeConfigResolver.COLUMN_DEFAULT_PREFIX}. */
@@ -43,15 +50,19 @@ final class ReadBridge {
       return raw;
     }
     // TODO(read-bridge): overlay columnDefaults onto raw.schemas() via withSchemaOverlay; future V3
-    // features bridged from config are applied here too.
+    // features bridged from config are applied here too. Two failure categories apply there, as
+    // here: a capability gap we don't yet support degrades to NULL, while an invariant violation
+    // (e.g. a default that can't bind to its column) fails loud.
     return raw;
   }
 
   /**
    * Decodes {@code field-id -> initial-default} from the {@code
-   * openhouse.read-bridge.column-default.*} config entries; empty when there are none. Skips any
-   * entry with a non-integer field-id or an unparseable value (fail-closed). Package-visible for
-   * unit testing in isolation.
+   * openhouse.read-bridge.column-default.*} config entries; empty when there are none. On a known
+   * entry, the server encoder guarantees an integer field-id and a value that round-trips through
+   * {@code readTree}, so a non-integer field-id or an unparseable value is an encoder bug or
+   * transport corruption — it throws rather than degrading. Unknown keys are ignored above.
+   * Package-visible for unit testing in isolation.
    */
   static Map<Integer, JsonNode> columnDefaults(Map<String, String> config) {
     if (config == null) {
@@ -66,10 +77,16 @@ final class ReadBridge {
         int fieldId = Integer.parseInt(entry.getKey().substring(COLUMN_DEFAULT_PREFIX.length()));
         byFieldId.put(fieldId, MAPPER.readTree(entry.getValue()));
       } catch (RuntimeException | JsonProcessingException e) {
-        log.warn(
-            "Skipping unparseable read-bridge config entry {}={}",
-            entry.getKey(),
-            entry.getValue(),
+        // The server encoder stamps an int field-id and a JsonNode value that round-trips through
+        // readTree, so reaching here means an encoder bug or transport corruption, not an expected
+        // state. Fail loud so it is caught, rather than silently reading NULL.
+        throw new IllegalStateException(
+            "read-bridge: unusable "
+                + COLUMN_DEFAULT_PREFIX
+                + " entry "
+                + entry.getKey()
+                + "="
+                + entry.getValue(),
             e);
       }
     }
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
@@ -1,0 +1,78 @@
+package com.linkedin.openhouse.javaclient;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.iceberg.TableMetadata;
+
+/**
+ * Read-time bridge: overlays Iceberg V3 read semantics onto loaded metadata for tables/clients that
+ * don't yet carry them natively, using behavior the server delivers in the per-table {@code
+ * config}. Today it applies per-column initial-defaults; further V3 features can be backported
+ * through the same entry point as they are added.
+ *
+ * <p>Client end of the read-bridge wire contract — mirror of the server encoder {@code
+ * ReadBridgeConfigResolver} (services/tables). The contract is flat, namespaced config keys (no
+ * envelope/POJO): {@code openhouse.read-bridge.column-default.<fieldId> = <single-value-json>}.
+ *
+ * <p>Kept out of {@link OpenHouseTableOperations} so the read path stays slim. Pure and
+ * fail-closed: an entry with a non-integer field-id or an unparseable value is skipped, and nothing
+ * to apply returns {@code raw} unchanged.
+ */
+@Slf4j
+final class ReadBridge {
+
+  /** Mirror of {@code ReadBridgeConfigResolver.COLUMN_DEFAULT_PREFIX}. */
+  static final String COLUMN_DEFAULT_PREFIX = "openhouse.read-bridge.column-default.";
+
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  private ReadBridge() {}
+
+  /**
+   * Applies the bridged read-time behavior the server stamped into {@code config} onto {@code raw},
+   * returning the transformed metadata (or {@code raw} when there is nothing to bridge).
+   */
+  static TableMetadata apply(TableMetadata raw, Map<String, String> config) {
+    Map<Integer, JsonNode> columnDefaults = columnDefaults(config);
+    if (columnDefaults.isEmpty()) {
+      return raw;
+    }
+    // TODO(read-bridge): overlay columnDefaults onto raw.schemas() via withSchemaOverlay; future V3
+    // features bridged from config are applied here too.
+    return raw;
+  }
+
+  /**
+   * Decodes {@code field-id -> initial-default} from the {@code
+   * openhouse.read-bridge.column-default.*} config entries; empty when there are none. Skips any
+   * entry with a non-integer field-id or an unparseable value (fail-closed). Package-visible for
+   * unit testing in isolation.
+   */
+  static Map<Integer, JsonNode> columnDefaults(Map<String, String> config) {
+    if (config == null) {
+      return Collections.emptyMap();
+    }
+    Map<Integer, JsonNode> byFieldId = new HashMap<>();
+    for (Map.Entry<String, String> entry : config.entrySet()) {
+      if (!entry.getKey().startsWith(COLUMN_DEFAULT_PREFIX)) {
+        continue;
+      }
+      try {
+        int fieldId = Integer.parseInt(entry.getKey().substring(COLUMN_DEFAULT_PREFIX.length()));
+        byFieldId.put(fieldId, MAPPER.readTree(entry.getValue()));
+      } catch (RuntimeException | JsonProcessingException e) {
+        log.warn(
+            "Skipping unparseable read-bridge config entry {}={}",
+            entry.getKey(),
+            entry.getValue(),
+            e);
+      }
+    }
+    return byFieldId;
+  }
+}
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseTablesApiHandler.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseTablesApiHandler.java
@@ -13,6 +13,7 @@ import com.linkedin.openhouse.tables.api.spec.v0.response.GetTableResponseBody;
 import com.linkedin.openhouse.tables.api.validator.TablesApiValidator;
 import com.linkedin.openhouse.tables.dto.mapper.TablesMapper;
 import com.linkedin.openhouse.tables.model.TableDto;
+import com.linkedin.openhouse.tables.readbridge.ReadBridgeConfigResolver;
 import com.linkedin.openhouse.tables.services.TablesService;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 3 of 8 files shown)


integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -0,0 +1,78 @@
...
+   * returning the transformed metadata (or {@code raw} when there is nothing to bridge).
+   */
+  static TableMetadata apply(TableMetadata raw, Map<String, String> config) {
+    Map<Integer, JsonNode> columnDefaults = columnDefaults(config);
+    if (columnDefaults.isEmpty()) {
+      return raw;
+    }
+    // TODO(read-bridge): overlay columnDefaults onto raw.schemas() via withSchemaOverlay; future V3
+    // features bridged from config are applied here too.
+    return raw;
```
shanthoosh, line 56: This method is currently a no-op. The apply() method currently always returns raw unchanged, and I couldn’t find a call site where this method is invoked during table metadata refresh/load. The server stamps the read-bridge config into LoadTableResponse, but the client never consumes it, so the feature isn’t active end-to-end. Do we plan to extend this further in the future PRs as a follow-up?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -0,0 +1,78 @@
...
+    Map<Integer, JsonNode> byFieldId = new HashMap<>();
+    for (Map.Entry<String, String> entry : config.entrySet()) {
+      if (!entry.getKey().startsWith(COLUMN_DEFAULT_PREFIX)) {
+        continue;
+      }
+      try {
+        int fieldId = Integer.parseInt(entry.getKey().substring(COLUMN_DEFAULT_PREFIX.length()));
+        byFieldId.put(fieldId, MAPPER.readTree(entry.getValue()));
+      } catch (RuntimeException | JsonProcessingException e) {
+        log.warn(
```
shanthoosh: Minor: It would be useful to log the COLUMN_DEFAULT_PREFIX to aid debugging

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+    Map<Integer, JsonNode> byFieldId = new HashMap<>();
+    for (Map.Entry<String, String> entry : config.entrySet()) {
+      if (!entry.getKey().startsWith(COLUMN_DEFAULT_PREFIX)) {
+        continue;
+      }
+      try {
+        int fieldId = Integer.parseInt(entry.getKey().substring(COLUMN_DEFAULT_PREFIX.length()));
+        byFieldId.put(fieldId, MAPPER.readTree(entry.getValue()));
+      } catch (RuntimeException | JsonProcessingException e) {
```
shanthoosh, line 79: Is it safe to silently not consider the (key, value) pairs that fails with exception and chose the other values. Are we expecting a fixed set of configuration key as json in this case? If so, would be reasonable to set default if the values for config-key are malformed?

shanthoosh (dismissed):
Looks good overall. Have few minor clarification questions/comments.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: Yes, by design this #645 is the O.S. substrate and the overlay transform marked TODO, and defaults only arrive once a ColumnDefaultsSource is supplied in closed/source since it relies on linkedin schema object "avro.schema.literal"

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: Thanks, I actually considered the below comment and converted the behavior to fail loud instead of softly with logging.  a decode failure on a column-default.* entry now fails with IllegalStateException with the offending key=value in the message, so it surfaces immediately instead of hiding in a warn.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: Good question, I changed the behavior to address it. Three parts:

1. converted a Silent Skip into Fail Loud (for known keys within openhouse.read-bridge.column-default.*). By construction the value is always a valid field within the schema. That means a decode failure isn't an expected input to the client and it can only be an encoder bug or transport corruption / issue within the ASL. Silently skipping it would hide a real defect (a column silently reads NULL instead of its default), so we now throw instead.

2. "Fixed set of keys as json?" no its not fixed. the keyspace is open: the field-id in the suffix is dynamic per table, so we can't enumerate expected keys. But each key's shape is encoded with int suffix + single-value JSON. 

3. "Set a default if malformed?" We deliberately don't, there's no safe fallback for "the table's default is unreadable," and fabricating one would risk returning a wrong value and hard coding NULL data (worse than a recognizable failure). So we fail loud rather than guess.

Caveat for follow-up work: this PR guarantees the value returned from server is well-formed, but not that it's the correct default for its column. When the schema overlay (e.g. ASL) PR lands, the apply step must fail loud if a default can't bind to its column, and the server must validate default-vs-schema consistency. we can't trust the client to be semantically correct on its own.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -0,0 +1,95 @@
+package com.linkedin.openhouse.javaclient;
```
mkuchenbecker, line 1: This is only 1.2; are we implementing on 1.5?

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
```diff
@@ -0,0 +1,47 @@
...
+  public static final String COLUMN_DEFAULT_PREFIX = "openhouse.read-bridge.column-default.";
+
+  private final ColumnDefaultsSource columnDefaultsSource;
+
+  public ReadBridgeConfigResolver(ColumnDefaultsSource columnDefaultsSource) {
+    this.columnDefaultsSource = columnDefaultsSource;
+  }
+
+  public Map<String, String> resolve(String databaseId, String tableId, TableDto tableDto) {
+    Map<Integer, JsonNode> columnDefaults = columnDefaultsSource.defaults(tableDto);
```
mkuchenbecker, line 37: Does this mean that if the table was to update the defaults we may read the newly changed defaults?

mkuchenbecker, line 37: How does it work with respect to MVCC? time travel?

mkuchenbecker:
> We don't need read-bridge there.
Why does 3.5 not need the read bridge?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: It's authored under iceberg-1.2, but it isn't 1.2-only. the 1.5 runtime also compiles this same source file. 

integrations/java/iceberg-1.5/openhouse-java-runtime/build.gradle adds the 1.2 module's source dirs to its own sourceSet:

`srcDirs += project(':integrations:java:iceberg-1.2:openhouse-java-runtime').sourceSets.main.java.srcDirs`

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
cbb330: Yes. And the defaults are never persisted into Iceberg metadata. there are two ways to prevent it:
1. remove the initial-defualts attribute from the serialization method in iceberg fork
2. sanitize the schema in OH server prior to writing the metadata file

I plan to implement 2. in a follow up PR. i prefer sanitizing the schema on OH server because we can change the runtime behavior easily in the future. Option 1 would tie us to the client version for changing behavior.

With that, a thing to clarify on consistency in the transaction: In upstream iceberg, linkedin's hive reader, and openhouse's iceberg client, the default value only takes action after the table has been committed and subsequently read. In all cases, an existing DataFrame that is evolved with a default column does not acquire a default value until after the evolution has been committed and a subsequent read has created a new dataframe.

cbb330: MVCC works same as other concurrent schema evolution operations -- config and the metadata location come from the same getTable response, and the overlay is applied inside the metadata load, so a given `TableMetadata` is always overlaid with config from that same server version. there isn't mixing across versions. A losing commit throws `CommitFailedException`, and Iceberg's retry calls ops.refresh() on each attempt, which re-fetches config and re-applies the overlay together.

Time travel is a good point to call out and must be handled carefully otherwise the client may fail while applying the overlay. the client overlay is in a follow up PR. It will look like this:
1. The server returns the default-value config together with the metadata location.
2. During metadata loading, the client decodes the config into a field-ID-to-default map, then applies it independently to every schema in the TableMetadata object. 
3. For each schema, the client attaches defaults only to field IDs present in that schema and ignores entries for fields that did not yet exist. 
4. The resulting schema-ID-to-overlaid-schema map is returned to the object, allowing normal latest, time-travel, tag, and branch reads to select the appropriate schema without unexpected errors.

cbb330:
> We don't need read-bridge there.
> > Why does 3.5 not need the read bridge?

@mkuchenbecker we actually need both. there was a mistake in the PR description, corrected it.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
mkuchenbecker: I think the defaults need to be immutable to the table. Otherwise a currently running query reloading the table will get inconsistent results on the same data for the same version. Under what conditions does the default value change and can it be immutable once set?

mkuchenbecker: But basically time traveling will not actually work on these table if the default value changes over the livetime and will give incorrect results. The problem is not there if we don't allow chaning the default.

mkuchenbecker (commented):
Restricting changing the default mitigates a major concern. Is there a problem doing so?

cbb330:
@mkuchenbecker we can do that, no problem. I have a follow up PR #678 that prevents schema from evolving to drop the default values. once they exist on table, they are there forever. 

this is exactly what the upstream datasource has in their contract, as well as Hive.  so no change in behavior for customer needed. 

It is also how the behavior exists in V3: once initial-defaults is set, it cannot be removed or changed while that column is in the table.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
cbb330: answered below. we keep default value immutable in a follow up pr #678

cbb330: yes, we handle time traveling correctly because we keep default value immutable in a follow up pr #678

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -128,11 +128,14 @@ public void doRefresh() {
...
-   * Loads the table metadata at the given location. Defaults to the stock parser; subclasses may
-   * override to transform the metadata (e.g. attach column defaults) as it loads.
+   * Loads the table metadata from storage, then runs the read-time {@link ReadBridge} over it using
+   * the per-table behavior the server stamped onto {@link #currentConfig()}. Absent config, or
+   * config carrying nothing to bridge, leaves the raw metadata untouched; a malformed known entry
+   * fails loud (see {@link ReadBridge}).
    */
   protected TableMetadata loadMetadata(String metadataLocation) {
-    return TableMetadataParser.read(io(), metadataLocation);
+    TableMetadata raw = TableMetadataParser.read(io(), metadataLocation);
```
mkuchenbecker, line 137: can we add a config to fall back to old behabiour so this is config gated?

mkuchenbecker: this may or may not b ethe right spot but the concern is exposing all customers the to codepath when this lands

cbb330: The intent of this PR is already to deploy without a behavior change. This line is a stub, empty config returns raw, and even a stamped config is not overlaid yet (that's #679). Iceberg readers therefore see the same schema as today, without initial-default, so fills don't change.

First I'd smoke test this client jar in spark-shell. and then deploy the client fleet and bake. Overlay is #679, and only for tables the server has stamped, that stamp is gated by the table feature flag in #674.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
```diff
@@ -0,0 +1,47 @@
...
+ * deployment delivers the bridge by overriding only {@link ColumnDefaultsSource}.
+ *
+ * <p><b>Mirror:</b> {@link #COLUMN_DEFAULT_PREFIX} is the shared contract with the client decoder;
+ * keep it in sync. Further V3 features ride the same {@code openhouse.read-bridge.*} namespace as
+ * additional keys.
+ */
+public class ReadBridgeConfigResolver {
+
+  /** Config key prefix for a per-column read-time default; suffixed with the Iceberg field-id. */
+  public static final String COLUMN_DEFAULT_PREFIX = "openhouse.read-bridge.column-default.";
```
abhisheknath2011, line 28: COLUMN_DEFAULT_PREFIX is duplicated in the client and server side code. Is there a common place to define the constants for default values?

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
```diff
@@ -0,0 +1,47 @@
...
+
+  private final ColumnDefaultsSource columnDefaultsSource;
+
+  public ReadBridgeConfigResolver(ColumnDefaultsSource columnDefaultsSource) {
+    this.columnDefaultsSource = columnDefaultsSource;
+  }
+
+  public Map<String, String> resolve(String databaseId, String tableId, TableDto tableDto) {
```
abhisheknath2011, line 36: TableDto should have databaseId and tableId. Do we need the explicit fields for them?

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseTablesApiHandler.java
```diff
@@ -35,15 +36,30 @@ public class OpenHouseTablesApiHandler implements TablesApiHandler {
...
 
+  @Autowired private ReadBridgeConfigResolver readBridgeConfigResolver;
+
+  /**
+   * Stamp the server-resolved, per-table client {@code config} (Iceberg REST {@code
+   * LoadTableResponse.config} convention) onto a freshly mapped response body. The mapper leaves
+   * {@code config} null; it is a request-time decision resolved here.
+   */
+  private GetTableResponseBody withConfig(
+      GetTableResponseBody body, String databaseId, String tableId, TableDto tableDto) {
```
abhisheknath2011, line 47: We could remove databaseId and tableId from here.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -0,0 +1,95 @@
...
+        continue;
+      }
+      try {
+        int fieldId = Integer.parseInt(entry.getKey().substring(COLUMN_DEFAULT_PREFIX.length()));
+        byFieldId.put(fieldId, MAPPER.readTree(entry.getValue()));
+      } catch (RuntimeException | JsonProcessingException e) {
+        // The server encoder stamps an int field-id and a JsonNode value that round-trips through
+        // readTree, so reaching here means an encoder bug or transport corruption, not an expected
+        // state. Fail loud so it is caught, rather than silently reading NULL.
+        throw new IllegalStateException(
```
abhisheknath2011, line 83: Should we add db and table as well in the error?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -128,11 +128,14 @@ public void doRefresh() {
...
+    return ReadBridge.apply(raw, currentConfig());
```
abhisheknath2011, line 138: How to get the current config on the client side?

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
cbb330: yeah i had the same thought, but there is not a common place to define constants which is shared between client and server. 

Because this prefix string is the only thing shared between them, I found that creating a jar that client and server depend on was heavy weight. so instead I opted to add a javadoc and add e2e tests in a follow up PR which asserts that client/server follow the same handshake.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
cbb330: good catch. this is leftover. 

Dropped in #674: resolve(TableDto) only.
https://github.com/linkedin/openhouse/blob/ea956d973a6043691e1f4a47a5cf5d789d38574b/services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java#L35

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseTablesApiHandler.java
cbb330: same thing as above, this is leftover.

Dropped in [https://github.com/linkedin/openhouse/pull/674: resolve(TableDto) only.](https://github.com/linkedin/openhouse/blob/ea956d973a6043691e1f4a47a5cf5d789d38574b/services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseTablesApiHandler.java#L42-L43)

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: Agreed. without the table id this looks like a random client exception.

`ReadBridge` only sees the config map, so I’ll attach it at `OpenHouseTableOperations` (it already has tableName()) rather than threading ids into the decoder.

#668 already wraps this ISE so Iceberg doesn’t retry. I’ll add the table name there. #668 wrap today: https://github.com/linkedin/openhouse/blob/chbush/read-bridge-decode-off-retry-path/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java (the catch (IllegalStateException e) in loadMetadata)

cbb330: committed in 668

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
cbb330: From the table-load response, not from Iceberg metadata.

doRefresh() GETs the table, stores GetTableResponseBody.config on an AtomicReference, then Iceberg loads the metadata file and our client loadMetadata() calls ReadBridge.apply(raw, currentConfig()).

currentConfig() is that last GET's map (null if none). It's a side-channel only, so we do not send it back on writes.

abhisheknath2011 (approved):

PR #648, opened 2026-07-18
Paginate HTS list-tables queries to avoid WebClient buffer limit

kamanavishnu:
# Problem & Solution Overview
`OpenHouseInternalCatalog#listTables(Namespace)` issued a single unpaginated HTS query (`findAllByDatabaseId` / `findAll`) to list every table in a database. For databases with a large number of tables, the serialized HTS response can exceed the WebClient in-memory buffer limit configured in `MainApplicationConfig`, causing a `DataBufferLimitException` that surfaces to callers as an HTTP 500 on `GET /tables?databaseId=...`.

This was observed in production for a database that grew past ~30k active tables — the per-table JSON payload (dominated by `metadataLocation` path length) pushed the total response size over the (then 20MB) ceiling.

**Relationship to #649**: #649 bumped the buffer limit 20MB -> 40MB as a stopgap mitigation, explicitly noting in its description that this is temporary and the durable fix is to paginate this call path. This PR is that durable fix: `listTables(Namespace)` now internally pages through HTS using the already-existing paginated repository methods (`findAll(Pageable)` / `findAllByDatabaseId(databaseId, Pageable)`, page size 1000), accumulating all pages before returning. This bounds every individual HTTP response regardless of how many tables a database has -- independent of whatever the buffer ceiling is set to -- while preserving the existing `List<TableIdentifier>` return type and behavior of the unpaginated API.

# Testing Done
- `./gradlew :iceberg:openhouse:internalcatalog:compileJava :services:tables:compileJava` — compiles cleanly.
- `./gradlew :iceberg:openhouse:internalcatalog:test` — all tests pass (11/11 in `OpenHouseInternalCatalogTest`, including 4 new tests covering single-page, multi-page, empty-database, and empty-namespace pagination paths), no regressions.
- **End-to-end local repro against the real code path**, run via Docker Compose (`infra/recipes/docker-compose/oh-only`, Tables Service + HouseTables Service, H2 in-memory DB):
  1. Built `services:tables`/`services:housetables` jars from this branch and rebuilt the Docker images (Dockerfiles copy prebuilt jars, so this step is required to pick up code changes).
  2. Seeded 55,000 synthetic `UserTable` rows into a single database directly via the HTS `PUT /hts/tables` API (bypassing full Iceberg table creation), with realistic `metadataLocation` path lengths.
  3. Called `POST /v1/databases/{databaseId}/tables/search` (`TablesController.searchTables` -> `OpenHouseInternalCatalog.listTables`) against the **pre-fix** jars first: reproduced the exact production failure, `DataBufferLimitException: Exceeded limit on max bytes to buffer : 20971520`.
  4. Re-ran the identical request against the **post-fix** jars: `HTTP 200`, all 55,000 results returned (18.3MB response, aggregated internally over 55 paginated HTS calls of 1000 rows each) -- confirming the fix resolves the failure regardless of buffer size.
- Risk: low — change is isolated to internal list-tables aggregation logic; public method signature and semantics are unchanged.

### DIFF
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
@@ -21,10 +21,11 @@ import com.linkedin.openhouse.internal.catalog.repository.exception.HouseTableCo
 import com.linkedin.openhouse.internal.catalog.repository.exception.HouseTableNotFoundException;
 import com.linkedin.openhouse.internal.catalog.repository.exception.HouseTableRepositoryException;
 import io.micrometer.core.instrument.MeterRegistry;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Function;
 import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.hadoop.fs.Path;
 import org.apache.iceberg.BaseMetastoreCatalog;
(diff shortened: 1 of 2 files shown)


mkuchenbecker (approved):

mkuchenbecker:
If this is ready for review and fixes the issue, +1. Please update the PR description though.

PR #652, opened 2026-07-22
Merge existing policies on CREATE OR REPLACE (RTAS)

mkuchenbecker:
## Summary

`CREATE OR REPLACE ... AS SELECT` (RTAS) silently dropped the table's policies. The replace path rebuilt the `policies` table property purely from the incoming request, so a replace that omitted policies wiped the existing retention, sharing, PII column tags, replication, and history, even though ordinary user table properties survived.

## Fix

Policies are table metadata that a replace must not silently drop. Before the replace properties are built, the existing table's policies are merged with the request's policies. The merge is based on the existing policies, so any plane that the request does not explicitly provide is carried forward from the existing table, and each plane that the request does provide overrides the existing value.

## Merge behavior

The table below describes how each policy plane behaves during a replace.

| Plane | When the request provides it | When the request omits it |
|---|---|---|
| `retention` | The request value is applied. | The existing value is carried forward. |
| `replication` | The request value is applied. | The existing value is carried forward. |
| `history` | The request value is applied. | The existing value is carried forward. |
| `lockState` | The request value is applied. | The existing value is carried forward. |
| `columnTags` | The request map replaces the existing map in full. This is an overwrite, not a per-key merge. | The existing map is carried forward. An omitted field and an empty map are treated the same way. |
| `sharingEnabled` | Not applicable, because this is a primitive boolean and a provided value cannot be distinguished from an omitted one. | The existing value is carried forward. |

Two consequences follow from this behavior. First, because column tags use overwrite semantics and an empty map is treated the same as an omitted field, a replace cannot clear all column tags. Clearing tags is done with `ALTER TABLE ... MODIFY COLUMN ... UNSET TAG`. Second, because `sharingEnabled` is a primitive boolean with no unset state, its value is always preserved across a replace. Sharing is changed with `ALTER TABLE ... SET POLICY (SHARING=...)`.

Spark RTAS has no policy clause, so it always sends a request with no policies, and the entire existing policies object is carried forward unchanged. A partial policy payload can only arrive from a client that calls the REST API directly.

This behavior is consistent with the intent of RTAS, which should preserve table properties so that a replace does not require re-granting access to the same entity.

## Testing Done

The REST level partial payload path is exercised through `RepositoryTest`, which is the layer that can send a partial `Policies` object. Spark cannot reach this path because it always sends a request with no policies.

- `testReplaceMergesExistingPolicies` replaces a table without policies and asserts that the retention policy survives.
- `testReplaceAppliesRequestedPolicies` asserts that a retention policy provided on the request is applied.
- `testReplaceWithPartialPoliciesPreservesSharing` sends a partial payload containing only retention and asserts that `sharingEnabled` stays true while the new retention is applied.
- `testReplaceWithPartialPoliciesPreservesOmittedPlanes` overrides only retention and asserts that the omitted history plane is carried forward.
- `testReplaceWithPartialPoliciesPreservesColumnTags` sends a payload that provides retention but omits column tags, and asserts that the existing column tag is carried forward.
- `testReplaceOverwritesColumnTags` sends a new column tag map and asserts that it replaces the existing map in full, dropping the previous tag.

Black box coverage is exercised through `RtasPolicyPreservationTest` against an embedded OpenHouse server driven by Spark SQL. It asserts that retention, sharing, the PII column tag, and history all survive a `REPLACE TABLE ... AS SELECT`.

The existing `SnapshotsControllerTest.testPutSnapshotsReplaceCommit` still passes, which confirms that a replace on a table that never had policies still yields none. `./gradlew :services:tables:test` and `:integrations:spark:spark-3.1:openhouse-spark-itest:catalogTest` pass on JDK 17, and Spotless is clean on the module.


services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/RepositoryTest.java
```diff
@@ -155,6 +156,52 @@ public void testUpdateTableWithModifiedPoliciesInTablePropsBlocked() {
     Assertions.assertFalse(openHouseInternalRepository.existsById(primaryKey));
   }
 
+  @Test
+  public void testReplaceMergesExistingPolicies() {
```
mkuchenbecker, line 166: this is not enough tests. We need to test that each policy is merged and what the merge logic is. does a partial merge change the policy? Can you add policies as part of RTAS?  these behaviors need to be captured

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -153,6 +153,10 @@ public TableDto save(TableDto tableDto) {
           System.currentTimeMillis() - startTime);
     } else if (tableDto.isStageReplace() || tableDto.isReplaceCommit()) {
       validateReplaceTable(tableIdentifier);
+      // Policies are governance metadata (retention, sharing, PII column tags, replication,
+      // history, lock) that must survive a replace. Merge the existing table's policies with the
+      // request's so RTAS never silently drops the planes the request omits.
```
mkuchenbecker:
```suggestion
      // request's so RTAS consistently merges properties.
```

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/RepositoryTest.java
mkuchenbecker: Added comprehensive coverage (commit 3a0f334d):

Unit tests over `mergePolicies` (`OpenHouseInternalRepositoryImplMergePoliciesTest`):
- each plane merged; every omitted plane carried forward from the existing table
- a provided plane wins — a partial merge changes only that plane
- policies can be **added** via RTAS (existing has retention; request adds history + PII column tags)
- `columnTags`: empty/absent request carries forward, non-empty request wins
- `sharingEnabled` (primitive boolean) takes the request's value
- null-existing and null-requested edge cases

E2e tests (`RepositoryTest`): retention survives an RTAS that omits policies, and an RTAS that *provides* a retention policy applies it.

services/tables/src/test/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImplMergePoliciesTest.java
```diff
@@ -0,0 +1,130 @@
...
+    assertTrue(merged.getLockState().isLocked());
+  }
+
+  @Test
+  void canAddPolicyPlanesViaRtas() {
+    // Existing has retention only; RTAS request adds history and PII column tags.
+    Policies existing = Policies.builder().retention(RETENTION_A).build();
+    Policies requested = Policies.builder().history(HISTORY_A).columnTags(COLUMN_TAGS_B).build();
+
+    Policies merged = OpenHouseInternalRepositoryImpl.mergePolicies(existing, requested);
```
mkuchenbecker: Do not unit test this function. Make black box tests with SQL. If we need a minimal testing harness to write these tests, do it.

mkuchenbecker: Done (commit b5d19843). Replaced the unit test with black-box SQL tests against a real embedded OpenHouse server (`OpenHouseSparkITest`):

- Set policies via `ALTER TABLE ... SET POLICY (RETENTION.../SHARING.../HISTORY...)` and `MODIFY COLUMN ... SET TAG (PII)`
- `CREATE OR REPLACE TABLE ... AS SELECT`
- Assert via `SHOW TBLPROPERTIES` that retention, sharing, the PII column tag, and history all survive the replace.

Two tests in `RtasPolicyPreservationTest` (runs on Spark 3.1 and 3.5). Also fixed `RTASTest.testRTAS`, which asserted the old buggy behavior (policies wiped on RTAS) — it now asserts the history policy is preserved. Teeth-checked: disabling the merge makes both new tests fail.

Note: Spark RTAS has no policy clause, so the SQL surface can only verify survival/carry-forward. The "provided-wins / add-via-RTAS" merge branch is reachable only via the REST request DTO, so it stays covered by the repository e2e tests in `RepositoryTest`.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -365,6 +369,53 @@ protected void validateReplaceTable(TableIdentifier tableIdentifier) {
...
+   * is a primitive boolean with no "unset" state, so the requested value is kept.
+   */
+  static Policies mergePolicies(Policies existing, Policies requested) {
+    if (existing == null) {
+      return requested;
+    }
+    if (requested == null) {
+      return existing;
+    }
+    return requested
```
cbb330: if any new policy is added, the builder would need to be updated as well, which may result in drift. is there a way to ensure there is never drift?

e..g: 
- deep copy policy object (without serializing into a builder)
- then merge the request object on top (again, without a builder since that would require hard coding the keys)

cbb330 (dismissed):

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
mkuchenbecker: Via reflection, possibly but any complex structure this is also problematic and we need a merge strategy per item. I.e. if we want to preserve individual values and nested structure there isn't going to be a seamless way to merge. 

In the vein of simplicity we don't need to merge. It could be the behaviour that setting new policies overwrites the old ones by treating the policies as a document. That is a potentially correct read but I think slightly confusing.

Side Note: Its unclear to me why policies are one table property with an API vs just a table property. 

I would opt for merging policies despite the additional plumbing work.

mkuchenbecker: https://github.com/linkedin/openhouse/pull/659
Example we can merge

mkuchenbecker:
PTAL @cbb330 , we can add a tripwire to make sure this is updated if we want to guard against silent loss - https://github.com/linkedin/openhouse/pull/659

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
mkuchenbecker: Applied — the comment now reads exactly this (the `// request's so RTAS consistently merges properties.` wording is in the current diff).

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -153,6 +153,10 @@ public TableDto save(TableDto tableDto) {
           System.currentTimeMillis() - startTime);
```
kamanavishnu, line 153: This comment is not related to this specific line. Posting here so conversation thread is retained.

Are users expecting their properties to copy over to the replaced table ? With this change, we are porting all the properties over without providing an option to bypass this port. 

If this is the intended behavior, I'm ok. But otherwise, I feel we should provide a hook to bypass porting over the properties. The default could still be to port over the properties. Thoughts ?

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -365,6 +369,53 @@ protected void validateReplaceTable(TableIdentifier tableIdentifier) {
...
+   */
+  static Policies mergePolicies(Policies existing, Policies requested) {
+    if (existing == null) {
+      return requested;
+    }
+    if (requested == null) {
+      return existing;
+    }
+    return requested
+        .toBuilder()
```
kamanavishnu, line 413: `sharingEnabled` is silently dropped by this merge, not preserved. It is a primitive `boolean` (default `false`) on `Policies`, and this builder chain never copies it from `existing` -- it always takes whatever value `requested` happens to carry.

This cannot manifest from Spark RTAS today, since Spark has no policy clause and always sends `policies == null` (short-circuited by the `requested == null` branch above, which returns `existing` wholesale). But any other client that supplies a *partial* `Policies` payload directly through the REST API -- e.g. only setting `retention` on a replace -- will silently flip `sharingEnabled` back to `false` on an existing table that had sharing enabled. That is exactly the class of bug this PR sets out to fix, just for this one plane.

Worth either merging `sharingEnabled` too (accepting that a `false` on the request can never be distinguished from "not provided", so document that tradeoff explicitly), or adding a test that replaces a table with `sharingEnabled=true` using a partial policies payload (retention only) and asserts sharing survives. None of the new tests cover this today.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -365,6 +369,53 @@ protected void validateReplaceTable(TableIdentifier tableIdentifier) {
...
+        .retention(
+            requested.getRetention() != null ? requested.getRetention() : existing.getRetention())
+        .columnTags(
+            MapUtils.isEmpty(requested.getColumnTags())
+                ? existing.getColumnTags()
```
kamanavishnu, line 418: Same "cannot express absence" problem shows up for `columnTags`: `MapUtils.isEmpty(requested.getColumnTags())` treats an explicit empty map (a client intentionally clearing all column tags on replace) identically to "not provided," and falls back to `existing.getColumnTags()`. So a replace can never remove column tags going forward -- it can only add/replace non-empty tag sets. Might be fine if that is the intended UX (tags are additive-only across a replace), but if a caller ever needs to drop column tags via RTAS, this silently blocks it. Worth a comment noting this is intentional, or a `null` vs `isEmpty()` distinction if removal should be supported.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
mkuchenbecker: >Are users expecting their properties to copy over to the replaced table ? With this change, we are porting all the properties over without providing an option to bypass this port.

Yes this is explicitly the behaviour of RTAS, documented here: https://docs.google.com/document/d/1BaVpSxdpVJp2EE4P7vlr2by29_rsPCWmMN9MRSZAT-I/edit?tab=t.0#bookmark=id.cs6xdy1269nz

Policies are ambiguous because its a single property. If you do not specify a policy, we should keep the existing. If you do specify a policy, we should EITHER: Merge or wholesale overwrite. My suggestion is merge properties because each set policy is a separate command. Its actually not possible to set a policy as part of RTAS with the semantics we have today. So there is no way to have continuity today with policies without a merge in the event we did allow policies to be set, or clear understanding its all-or-nothing for policies. 

The way to avoid porting the policy is to explicitly revoke.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
mkuchenbecker: Good callout, I reversed the base so we start from existing. I don't think we need to allow disabling sharing wiht RTAS, but we do need to preserve sharing with RTAS. RTAS is not actually useful if you need to re-grant permissions afterwards to each user, but a single command can disable. 

So the impact due to the null / false issue is there is no way for RTAS to be disabled, which is a fine trade off IMO.

Documented in the PR description.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
mkuchenbecker: Replace I believe is the intended behaviour, but have documented in this case. It would ahve the same problem of only additive then a person that was to RTAS while changing tags ends up with a superset. We can revisit and get a review on this specific decision as part of PRR. 

So the behaviour is tags act like a document. If you want to replace tags you may replace teh tags. If you want to add a tag, use teh normal API.

kamanavishnu (approved):
LGTM!

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
mkuchenbecker: resolving. We can merge the tripwire to provide protection.

PR #655, opened 2026-07-23
Add gated Iceberg view support to OpenHouseCatalog (iceberg-1.5, off by default) 

aastha25:
## Summary

First increment of native OpenHouse view support — production code, gated and off by default. The iceberg-1.5 build of  `OpenHouseCatalog`  now  extends `BaseMetastoreViewCatalog`  (tables inherited unchanged) and implements  `loadView` / `buildView` / `listViews` / `dropView` / `renameView` , all gated behind an  `iceberg-views-enabled`  catalog property that defaults to off (throws  `UnsupportedOperationException`  when disabled). Views fold into the existing catalog — no separate/parallel view catalog — and the iceberg-1.2 / Spark-3.1 runtime stays table-only.

The only placeholder is the view backend: an in-memory mock store standing in for the OpenHouse Views service until its API and client exist. Everything else is the intended final shape. 

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [ ] Tests

✓ New Features
•  `OpenHouseCatalog`  (iceberg-1.5)  extends `BaseMetastoreViewCatalog` ; view ops gated behind  iceberg-views-enabled  (default  false  → throw). Tables unchanged.
• View backend is an in-memory mock ( `ConcurrentHashMap<TableIdentifier, ViewMetadata>`  via an inline  `newViewOps` ) — the sole placeholder, to be swapped for a Views-service-backed  `newViewOps()` → `OpenHouseViewOperations` → `ViewApi` , mirroring  `newTableOps()` → `OpenHouseTableOperations` → `TableApi` .
✓ Refactoring
•  `OpenHouseCatalog`  now diverges per Iceberg version, so it moves to a per-version source dir  `src/iceberg-version-specific/java`  in both modules. iceberg-1.2 keeps its table-only copy ( extends `BaseMetastoreCatalog` ); iceberg-1.5 inherits all shared  src/main/java  from 1.2 except that dir and supplies its own view-capable copy. Two  `build.gradle`  edits wire the split.

## Testing Done

✓ No tests added or updated.
- View backend is a mock placeholder pending the Views service, so there's no meaningful assert surface yet; tests land with the real backend.
-✓ Manually validated the  buildView(...).create()  →  loadView(...)  round-trip via spark-shell (Spark 3.5) against a locally published snapshot with  iceberg-views-enabled=true ; disabled catalog throws as expected. 

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

• Breaking Changes — none (purely additive; new class, no existing code paths touched).
• Deprecations — none.


integrations/java/iceberg-1.5/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseViewCatalog.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  @Override
+  public ViewBuilder buildView(TableIdentifier identifier) {
+    // The mock backend rejects creation into a missing namespace; auto-create so a bare
+    // CREATE VIEW works without a prior CREATE NAMESPACE (real backend will own this policy).
+    Namespace namespace = identifier.namespace();
+    if (!mockViewCatalog.namespaceExists(namespace)) {
+      mockViewCatalog.createNamespace(namespace);
+    }
```
mkuchenbecker: Should we prioritize database level abstraction in openhouse for views?

aastha25: Agreed that we should prioritize the database-level abstraction for views regardless. Having said that, we likely don't need net-new database machinery for the initial setup and views can reuse the exact db provisioning model tables already follow.

integrations/java/iceberg-1.5/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseViewCatalog.java
```diff
@@ -0,0 +1,115 @@
...
+ * com.linkedin.openhouse.spark.OpenHouseCatalog}, which is not a {@code ViewCatalog} — and
+ * Iceberg's {@code SparkCatalog} enables views only when that configured instance is {@code
+ * instanceof ViewCatalog}. So {@code OpenHouseViewCatalog} is reached only by either (a) repointing
+ * {@code catalog-impl} at a thin Spark wrapper over this class (the config change doubles as a
+ * prototype gate), or (b) making the configured {@code OpenHouseCatalog} itself implement {@code
+ * ViewCatalog} (Spark-3.5 / iceberg-1.5 build only). For prototyping, for noww, (b) has been
+ * adopted to minimize any interference with production code paths.
+ */
+@Slf4j
+public class OpenHouseViewCatalog extends OpenHouseCatalog implements ViewCatalog {
```
wmoustafa: Assuming this will replace Conventional OH catalog? Why not directly extend the conventional one to implement ViewCatalog to maintain backward compatibility and avoid potential drift between adopting table vs view catalog?

integrations/java/iceberg-1.5/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseViewCatalog.java
```diff
@@ -0,0 +1,115 @@
...
+  // MOCK view backend: ephemeral and no persistence. Stand-in until the OpenHouse Views
+  // service and its generated client exist.
+  // Evolution: swap this InMemoryCatalog for an OpenHouse Views-service-backed view catalog that
+  // mirrors how OpenHouseCatalog talks to the Tables service — i.e. a BaseMetastoreViewCatalog
+  // whose
+  // newViewOps() returns an OpenHouseViewOperations calling a generated ViewApi, exactly as
+  // OpenHouseCatalog.newTableOps() returns OpenHouseTableOperations calling TableApi. Only this
+  // field's type + init change; the ViewCatalog delegation below stays identical.
+  // TODO(views-prototype): swap mock -> OH Views service client once the server API + ViewApi land.
+  private final InMemoryCatalog mockViewCatalog = new InMemoryCatalog();
```
wmoustafa: One approach to make this PR production ready is to move the InMemoryCatalog to test classes and keep the APIs here.

integrations/java/iceberg-1.5/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseViewCatalog.java
aastha25: the existing  OpenHouseCatalog  (table catalog) is used in Spark 3.1 with Iceberg 1.2, and this prototype only enables views on Spark 3.5 / Iceberg 1.5, hence the extension + isolation.  It'll be an easy swap to fold  ViewCatalog  into  OpenHouseCatalog  directly if & when we extend support to the 3.1 branch.

mkuchenbecker: I highly suggest that we just target 3.5 first and get it solid before expanding the surface area to 3.1 / iceberg1.2.

integrations/java/iceberg-1.5/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseViewCatalog.java
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.iceberg.view.ViewBuilder;
+
+/**
+ * View-capable {@link OpenHouseCatalog}. Table operations are inherited unchanged; view operations
+ * are added by implementing Iceberg's {@link ViewCatalog}. This makes a single catalog object serve
+ * both tables and views, which is what Iceberg's Spark {@code SparkCatalog} requires to wire {@code
+ * asViewCatalog} (it only enables views when the configured {@code catalog-impl} is {@code
+ * instanceof ViewCatalog}).
+ *
+ * <p>Because it already {@code extends OpenHouseCatalog} (for tables), Java single inheritance
```
mkuchenbecker: Can we do this via composition vs inheritance?

aastha25: Assuming you're asking - can `OpenHouseViewCatalog`  hold  `OpenHouseCatalog`  (composition) instead of extend  it? Yes, thats possible,  `OHViewCatalog`  could implement the table operations and forward those reqs to a composed  `OpenHouseCatalog` . Downsides are that (1) we hand-forward the entire table + namespace + grant surface, and (2) any new method added to  `OpenHouseCatalog`  has to be wired up explicitly in  `OpenHouseViewCatalog`  whereas in the current  approach, we get it for free and its clean.

If we dont foresee a lot of damage from (2), we could do the composition & `OHViewCatalog` extends BaseMetaStoreViewCatalog. But, that also adds a lot of biolerplate code to wire up its (generic) table ops back to OH's REST table logic.

aastha25: we can roll with the existing layout and revisit soon-ish if composed viewCatalog turns out to be tricky / unclean

cbb330 (changes requested):
Thanks for the prototype. 

In production today we already run two catalogs with different responsibilities:

```java
spark.sql.catalog.openhouse.catalog-impl=com.linkedin.openhouse.spark.LiOpenHouseCatalog
spark.sql.catalog.spark_catalog=com.linkedin.coral.spark.catalog.CoralSparkViewCatalog
```
The issue is that Coral catalog has more logic than the OH catalogs, and the divergence has been a lasting production/compliance hole. reconciliation of 2 into 1 is high LoE and high risk (caused one outage) and is stalled. The testing across both catalog's diverges so changes in 1 can break 2 like a CTAS doesn't carry nullability bug OH fixed last year. 

`OpenHouseViewCatalog` (and the likely follow-on `LiOpenHouseViewCatalog`) creates a third catalog type at the Iceberg layer:

```bash
OpenHouseCatalog
├── LiOpenHouseCatalog           # prod tables
└── OpenHouseViewCatalog         # this PR
    └── LiOpenHouseViewCatalog?  # likely LI fork
+ CoralSparkViewCatalog          # prod views (already diverged)
```

Also, Coral handles views for spark_catalog. This PR makes openhouse handle views too. Two paths = two places to keep in sync = another divergence problem.


TLDR; The primary issue is three parallel production catalogs with diverging policies and implementation.

and the ask: don't land this as a new long-lived catalog type on main yet.

Preferred paths:
1. Fold into the existing catalog: implement ViewCatalog on OpenHouseCatalog for the Spark 3.5 / iceberg-1.5 line, instead of a parallel OpenHouseViewCatalog (and a future LiOpenHouseViewCatalog).
2. Or keep the prototype in test code: if the goal is just create/read plumbing, put the mock in test sources / a test module so we don't ship InMemoryCatalog as a production catalog-impl.
3. Otherwise, pause on merging until we decide the Coral relationship in the implementation: either OH becomes Coral’s view backend, or OH replaces Coral for OH-native views with a clear deprecation path.

aastha25:
Thanks @cbb330 , this is the right frame, and I share the core concern. Catalog unification has been top-of-mind for the group and me for a while; we're not looking to add a third top-level catalog to manage, and I'm tracking the unification work with tickets so this doesn't harden into a permanent parallel catalog. This is a **gated alpha**, not a new user-facing production catalog today, but it's also the direction and here's how it converges.

(1) Default catalog stays table-only. The production  `catalog-impl`  remains  `OpenHouseCatalog`  (table-only. I specifically don't want it to be a  `ViewCatalog` , since that would expose create/load view to every  `openhouse.*` calling user. The alpha instead allows pointing `catalog-impl`  at  `OpenHouseViewCatalog` - the same instance serving tables (inherited) and views (implemented)- selected only behind a non-advertised config plus. It's not a second  spark.sql.catalog.*  entry; structurally it's already your path (A), just expressed as a subclass and gated.
(2) Prod path untouched. Every real 2-part  CREATE/LOAD  keeps flowing through the `CoralCatalog` →  HMS  via Spark, and 3-part table ops keep flowing through  `OpenHouseCatalog`  across 3.1 & 3.5. The only new path is 3-part view ops, and it's gated. Freezing the prod paths is deliberate — it keeps ops simple and lets us ship something functional and safe instead of iterating on edge cases for safe deployments.

Mapping to your three suggestions:
(A) Fold  ViewCatalog  into the catalog (1.5 path): agree in spirit on unification / no parallel catalog but for the alpha the default catalog stays table-only by design, so views aren't broadly exposed; the view-enabled subclass is the gated interim, and unification is tracked so it doesn't fall off the radar. 
(B) Keep it in test code: the gated alpha needs to run end-to-end in Spark for devs, so test sources won't do but the config gives the same guarantee you're after: it never reaches general users.
(C) Decide the Coral relationship first: agreed it must be settled before GA, and we're actively exploring integration recipes for Spark 3.5 view-catalog <-> Coral. Since the alpha is gated and the prod  CoralCatalog  ->  HMS  path is untouched, it doesn't need to block progress here.

**Ask**: you've made real strides with the  `catalog-impl`  — could you point me to (1) documentation on how it was added, where it's all configured, and (2) the CTAS-nullability fix you mentioned? I want the view catalog to close those same gaps and budget bandwidth up front rather than rediscover them.

cbb330:
Thanks — the gate and keeping the default catalog table-only both make sense.

I still don't want a new catalog class on main. Swapping `catalog-impl` to a subclass is the same split, just not a second `spark.sql.catalog.*` name. Anyone who points at `OpenHouseViewCatalog` gets a second view path.

Not putting `ViewCatalog` on `OpenHouseCatalog` so every `openhouse.*` user sees views is a real concern. That can be a flag on the existing catalog (views off by default) rather than a new type. That is what I meant by folding it in.

I am fine with an alpha that needs a real Spark session. I am not fine with `InMemoryCatalog` as a production `catalog-impl`, or with Coral waiting on tickets after this class has landed.

On the two pointers you asked for:

- Nullability / CTAS: https://github.com/linkedin-multiproduct/li-openhouse/pull/1289
- Where `catalog-impl` is set: https://github.com/linkedin-multiproduct/grid-template-platform/blob/23485d60cf2bac63984334ca92ad50ec4e9bc7f4/rendered/spark-311/grid1-k8s-0/conf/spark-defaults.conf#L231
- The class that conf loads: https://github.com/linkedin-multiproduct/li-openhouse/blob/master/integrations/spark/spark-3.1/li-openhouse-spark-runtime/src/main/java/com/linkedin/openhouse/spark/LiOpenHouseCatalog.java

aastha25:
thanks @cbb330,  this is exactly the shape I have adopted now. Pushed an update that folds views into  `OpenHouseCatalog`  itself and drops the standalone class.

(1) `ViewCatalog`  is now implemented on  `OpenHouseCatalog`  directly — on the iceberg-1.5 build. it extends `BaseMetastoreViewCatalog` , with tables inherited unchanged. 

(2) Flag, off by default. View ops are gated behind  `iceberg-views-enabled`  (default  false ). When off, every view call throws and the default catalog is table-only for all  `openhouse.*`  users. Turning it on needs a real Spark session with the flag set — the `alpha` path you're fine with.

(3) No InMemoryCatalog. Dropped — it's not shipping as a production  catalog-impl . I only need minimum create→read functionality until the Views service and  ViewApi  exist, so the gated backend is a small in-memory  `Map<TableIdentifier, ViewMetadata>`  behind  `newViewOps` . That swaps for the Views-service-backed  newViewOps → OpenHouseViewOperations → ViewApi  later, mirroring  newTableOps → TableApi . 

(4) Source duplication, not a runtime split.  OpenHouseCatalog  is now duplicated per iceberg version — table-only on iceberg-1.2, view-capable on iceberg-1.5 — since divergence between those runtimes is inevitable as they drift. This is a build-time source copy . The duplication mechanism is easy to adjust later, but the current split builds clean on both.

cbb330:
looks good @aastha25 one question remaining on the test infra

Could you add one shared OpenHouseSparkITest suite that runs unchanged in both the Spark 3.1 and Spark 3.5 itest modules? It should execute the same view SQL and assert the same observable outcomes: success/failure, returned rows, and compatible exception/error behavior.

Please cover CREATE/SHOW/DROP/ALTER VIEW, missing-view reads, and normal table create/read/list/drop behavior. If Iceberg 1.5 emits different exceptions, normalize that in the Spark catalog so existing user code does not break.

aastha25:
@cbb330 , Done, added shared parity suite + exception normalization in `OpenHouseViewSparkITest`. Plus a 3.5-only  `OpenHouseViewEnabledTestSpark3_5`  for the enabled  `buildView`→`loadView`→`listViews`→`dropView`  round-trip.

```
┌───────────────────────────────┬────────────────────────┬────────────────────┐
│ Op (views disabled = default) │ 3.1 table-only         │ 3.5 views-off      │
├───────────────────────────────┼────────────────────────┼────────────────────┤
│ Table create/read/list/drop   │ identical rows+listing │ identical          │
├───────────────────────────────┼────────────────────────┼────────────────────┤
│ Read missing relation         │ AnalysisException      │ AnalysisException  │
├───────────────────────────────┼────────────────────────┼────────────────────┤
│ CREATE / ALTER / DROP VIEW    │ AnalysisException      │ AnalysisException  │
├───────────────────────────────┼────────────────────────┼────────────────────┤
│ SHOW VIEWS                    │ empty, no raw leak     │ empty, no raw leak │
└───────────────────────────────┴────────────────────────┴────────────────────┘
```
`For OSS parity`: when view support is disabled,  `buildView`  throws  `NoSuchNamespaceException`  instead of  `UnsupportedOperationException` , so stock  `SparkCatalog.createView`  maps it to `AnalysisException`  — same interpretation as the 3.1 table-only catalog.`Why only CREATE VIEW needed this`: Spark probes `loadView`  on every unresolved identifier and catches only `NoSuchViewException`  to fall back — so `read`/`ALTER`/`DROP` normalize for free via my  `loadView` → `NoSuchViewException` .  `CREATE VIEW`  alone hits  `SparkCatalog.createView` → `buildView`  directly (no  `loadView`  probe), and  `createView`  catches only  `NoSuchNamespaceException` / `AlreadyExistsException`  — hence that's the type  `buildView` must throw. We could've overridden the `CREATE VIEW` test in 3.5 to assert its divergent exception, but instead we normalized the exception so both runtimes behave identically (preserving the pre-view-support behavior for existing users) rather than codifying the divergence — with view-enablement covered by a separate 3.5-only test.


`Where does normalization belong`: This PR is OSS (`Iceberg` + embedded OH + OSS Spark); the Spark catalog here is stock  `SparkCatalog` . The LI integration is the internal Spark wrapper (a  `SparkCatalog` /`ViewCatalog` subclass) — that's the right home for user-facing exception massaging, which I'll do in a separate internal PR with a precise "views not enabled" message. So **follow up work** -  override  `createView`  in internal OHSparkCatalog  to throw a Spark `AnalysisException`  with an actionable "OpenHouse views not enabled" message when views are disabled, replacing the OSS  NoSuchNamespaceException  type-shim with a precise user-facing error.

cbb330:
@aastha25 thanks, the shared tests look mostly correct and ensures safety. 

One leftover question, your table says `SHOW VIEWS` is empty on both 3.1 and 3.5. But the test still accepts either an empty list or `AnalysisException`. If both really return empty, please assert empty only so that cannot drift. Same for `DROP VIEW IF EXISTS`

aastha25:
@cbb330 , took a deeper look and the earlier per-op matrix was a simplified view and conflated distinct code paths. 

The root asymmetry is in Spark itself, independent of this PR: Spark 3.1 has no v2  `ViewCatalog`  SPI, so view DDL and  `SHOW VIEWS`  throw  `AnalysisException` ; Spark 3.5 has it, so  `SHOW VIEWS`  returns empty. That difference exists even without this PR, and it naturally propagates into any single suite that validates the negative cases (3.1 throws, 3.5 returns empty). Rather than fork the suite, the shared  `OpenHouseViewSparkITest`  keeps one suite and double-handles the two divergent commands ( `SHOW VIEWS` , ` DROP VIEW IF EXISTS` ) — asserting one deterministic outcome per runtime (gated on  instanceof ViewCatalog , so it can't drift).

I've verified no behavioral change / regression vs. the pre-PR baseline across all four environments: li-spark 3.1, li-spark 3.5, and the OSS OH integration tests on Spark 3.1 and Spark 3.5. **These changes are non-blocking for both OSS OH and li-OH.** 

Revised matrix ( before → after  this PR; views disabled = default):

```
┌─────────────────────────────┬───────────────────────────────────────┬───────────────────────────────────────────────────┐
│ Op                          │ OSS Spark 3.1                         │ OSS Spark 3.5                                     │
├─────────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────────────────┤
│ Table create/read/list/drop │ works → works                         │ works → works                                     │
├─────────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────────────────┤
│ Read missing relation       │ AnalysisException → AnalysisException │ AnalysisException → AnalysisException             │
├─────────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────────────────┤
│ CREATE VIEW                 │ AnalysisException → AnalysisException │ UnsupportedOperationException → AnalysisException │
├─────────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────────────────┤
│ ALTER VIEW                  │ AnalysisException → AnalysisException │ AnalysisException → AnalysisException             │
├─────────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────────────────┤
│ DROP VIEW                   │ AnalysisException → AnalysisException │ AnalysisException → AnalysisException             │
├─────────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────────────────┤
│ SHOW VIEWS                  │ AnalysisException → AnalysisException │ empty → empty                                     │
├─────────────────────────────┼───────────────────────────────────────┼───────────────────────────────────────────────────┤
│ DROP VIEW IF EXISTS         │ AnalysisException → AnalysisException │ no-op → no-op                                     │
└─────────────────────────────┴───────────────────────────────────────┴───────────────────────────────────────────────────┘
```

The only cell that changes is OSS 3.5  CREATE VIEW  — a raw  UnsupportedOperationException  becomes an  AnalysisException  (the normalization in this PR), i.e. a strict improvement. Everything else is  X → X ; Spark 3.1 is untouched end-to-end.

cbb330 (approved):

integrations/java/iceberg-1.5/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseViewCatalog.java
aastha25: removed the inmemorycatalog.

PR #656, opened 2026-07-23
Disallow enabling RTAS alongside WAP or replication

mkuchenbecker:
## Summary

RTAS (`replace.enabled`) could be enabled on a table that also had WAP (`write.wap.enabled`) or
replication configured. RTAS is incompatible. This stops them from being enabled together. Its implemented as a generic "these features are incompatible" set of tuples so defining new exclusions or cleaning them up is cheap / easy.

## Fix

Don't allow enabling RTAS in conjunction with WAP and  replication at the same time.

Reject, on both create and update, any request whose resulting metadata would enable RTAS alongside
WAP or replication (`OpenHouseInternalRepositoryImpl.validateFeatureCompatibility`, called at the top
of `save`). 

## Testing Done

- **Integration (black-box SQL, real embedded server + Spark):** `RtasWapExclusivityTest`
  - enabling `replace.enabled` on a WAP table is rejected,
  - enabling `write.wap.enabled` on an RTAS table is rejected,
  - creating a table with both RTAS and WAP is rejected,
  - enabling `replace.enabled` on a table with a replication policy is rejected.
- Full `catalogTest` suite green; Spotless clean.


services/common/src/main/java/com/linkedin/openhouse/common/exception/UnsupportedClientOperationException.java
```diff
@@ -22,5 +22,6 @@ public enum Operation {
     GRANT_ON_LOCKED_TABLES,
     LOCKED_TABLE_OPERATION,
     RTAS_DISABLED,
+    INCOMPATIBLE_TBLPROPS,
```
mkuchenbecker: INCOMPATIBLE_FEATURES

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -316,6 +317,48 @@ protected void creationEligibilityCheck(TableDto tableDto) {
...
+   * RTAS ({@value CatalogConstants#RTAS_ENABLED_TABLE_PROP}) cannot coexist with WAP ({@value
+   * CatalogConstants#WAP_ENABLED_TABLE_PROP}) or replication on the same table: a staged WAP write
+   * or a replicated table and a whole-table replace do not compose. Reject any create or update
+   * whose resulting metadata would enable RTAS alongside either.
+   *
+   * @param tableDto container of the requested table metadata
+   */
+  private void validateFeatureCompatibility(TableDto tableDto) {
+    Map<String, String> props = tableDto.getTableProperties();
+    boolean rtasEnabled = props != null && Boolean.parseBoolean(props.get(RTAS_ENABLED_TABLE_PROP));
```
mkuchenbecker: the compatilbility were seems to be scoped to RTAS vs a generic "these two features are not compatible." I owuld change that so we can just declare which properties are mutually exclusive, and the message.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -316,6 +317,48 @@ protected void creationEligibilityCheck(TableDto tableDto) {
...
+    if (!rtasEnabled) {
+      return;
+    }
+    boolean wapEnabled = props != null && Boolean.parseBoolean(props.get(WAP_ENABLED_TABLE_PROP));
+    if (wapEnabled) {
+      throw new UnsupportedClientOperationException(
+          UnsupportedClientOperationException.Operation.INCOMPATIBLE_TBLPROPS,
+          String.format(
+              "Table %s.%s cannot enable both RTAS ('%s'='true') and WAP ('%s'='true') at the same time; they are mutually exclusive.",
```
mkuchenbecker: instead of saying they are mux, just say to disable the currently enabled fieature to enable the feature that is being rejected. It should be generic.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -316,6 +317,48 @@ protected void creationEligibilityCheck(TableDto tableDto) {
...
+      throw new UnsupportedClientOperationException(
+          UnsupportedClientOperationException.Operation.INCOMPATIBLE_TBLPROPS,
+          String.format(
+              "Table %s.%s cannot enable RTAS ('%s'='true') while replication is configured; they are mutually exclusive.",
+              tableDto.getDatabaseId(), tableDto.getTableId(), RTAS_ENABLED_TABLE_PROP));
+    }
+  }
+
+  private boolean isReplicationConfigured(
+      com.linkedin.openhouse.tables.api.spec.v0.request.components.Policies policies) {
```
mkuchenbecker: Is there a reason we need this to be fully qualified?

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -316,6 +317,48 @@ protected void creationEligibilityCheck(TableDto tableDto) {
...
+    return policies != null
```
mkuchenbecker: do not supply nullable vs optional

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -316,6 +317,48 @@ protected void creationEligibilityCheck(TableDto tableDto) {
...
+        && policies.getReplication() != null
+        && policies.getReplication().getConfig() != null
+        && !policies.getReplication().getConfig().isEmpty();
```
mkuchenbecker: Is this a valid condition? Is it independently tested? When does it occur, after replicaiton is disabled?

services/common/src/main/java/com/linkedin/openhouse/common/exception/UnsupportedClientOperationException.java
mkuchenbecker: Done — renamed the operation to `INCOMPATIBLE_FEATURES` (e6dce225). It's about incompatible features, not just table properties.

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
mkuchenbecker: Generalized. `validateFeatureCompatibility` no longer hardcodes RTAS-vs-WAP/replication branches — it iterates a declarative `MUTUALLY_EXCLUSIVE_FEATURES` list of feature pairs, where each `Feature` is a label + an "is-enabled" predicate. Declaring a new incompatibility is now just adding a pair. (e6dce225)

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
mkuchenbecker: Reworded to be generic and actionable: "Table X.Y cannot enable %s while %s is enabled. Disable %s to enable %s." — it tells the client which enabled feature to disable, instead of saying they're "mutually exclusive". (e6dce225)

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
mkuchenbecker: No — `Policies` is already imported in this file, so the fully-qualified name was unnecessary. Removed it; the helper now takes `Optional<Policies>`. (e6dce225)

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
mkuchenbecker: Fixed — the helper now takes `Optional<Policies>` instead of a nullable `Policies`, and reads it with `map(...).orElse(false)`. (e6dce225)

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
mkuchenbecker: Good questions — I dug in and it is a valid condition, and it's now independently tested.

- Validity/when it occurs: a `Policies` object can carry a non-null `replication` block whose `config` (destination list) is null or empty — e.g. a table that never configured replication, or where the replication block exists without destinations. So `getReplication() != null` alone would over-report; the `!getConfig().isEmpty()` check is what confirms replication is *actually* configured (has ≥1 destination). This matches how replication is treated elsewhere (a destination is required — see `ReplicationConfigValidator`).
- Consolidation: this exact check was duplicated in `validateReplaceTable`; I extracted it into the single `isReplicationConfigured(Optional<Policies>)` helper and reuse it from both call sites.
- Testing: added `OpenHouseInternalRepositoryImplTest.isReplicationConfigured*` covering empty Optional, no replication block, null/empty config (all false) and a configured destination (true). The end-to-end RTAS+replication rejection also stays covered by the black-box `RtasWapExclusivityTest.testCannotEnableRtasWhenReplicationConfigured`. (e6dce225)

services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
```diff
@@ -316,6 +320,93 @@ protected void creationEligibilityCheck(TableDto tableDto) {
...
+  }
+
+  private static final Feature RTAS_FEATURE =
+      new Feature("RTAS", tableDto -> isTablePropEnabled(tableDto, RTAS_ENABLED_TABLE_PROP));
+  private static final Feature WAP_FEATURE =
+      new Feature("WAP", tableDto -> isTablePropEnabled(tableDto, WAP_ENABLED_TABLE_PROP));
+  private static final Feature REPLICATION_FEATURE =
+      new Feature(
+          "replication",
+          tableDto -> isReplicationConfigured(Optional.ofNullable(tableDto.getPolicies())));
```
mkuchenbecker: Define incompatible features as a list of tuples. Then the code is generic.

mkuchenbecker: Done — the incompatibilities are now a list of tuples: added an `IncompatibleFeatures(feature, conflictsWith)` type and `MUTUALLY_EXCLUSIVE_FEATURES` is `List<IncompatibleFeatures>`, replacing the `Feature[]` array pairs. The validation loop is fully generic over the tuples. (999304f4)

integrations/spark/spark-3.1/openhouse-spark-itest/src/test/java/com/linkedin/openhouse/spark/catalogtest/RtasWapExclusivityTest.java
```diff
@@ -0,0 +1,114 @@
...
+  @Test
+  public void testCannotEnableRtasWhenReplicationConfigured() throws Exception {
+    try (SparkSession spark = getSparkSession()) {
+      String table = "openhouse.dbRtasWap.replicationFirst";
+      spark.sql("DROP TABLE IF EXISTS " + table);
+      spark.sql("CREATE TABLE " + table + " (id bigint, data string) USING iceberg");
+      spark.sql(
+          "ALTER TABLE " + table + " SET POLICY (REPLICATION=({destination:'WAR', interval:12h}))");
+
+      BadRequestException exception =
```
mkuchenbecker, line 101: is this a 4xx error?

mkuchenbecker: Yes — 400 Bad Request (user error), not a 5xx. The chain:

1. The check throws `UnsupportedClientOperationException` (a client-operation error).
2. `OpenHouseExceptionHandler.handleUnsupportedClientOperationException` maps that to `HttpStatus.BAD_REQUEST` (400).
3. The Iceberg REST client turns a 400 into `org.apache.iceberg.exceptions.BadRequestException`, which is exactly what this test asserts via `assertThrows(BadRequestException.class, ...)`.

So the test is also the proof that it surfaces as a 4xx: if it were reported as a 5xx it would come back as a different Iceberg exception (e.g. a server/`5xx` error) and this assertion would fail.

mkuchenbecker: Follow-up: I changed the exception from `UnsupportedClientOperationException` to `RequestValidationFailureException` (still HTTP 400 → `BadRequestException`, which this test asserts). ALTER/CREATE are supported operations; it's the requested property combination that's invalid, so "request validation failure" is the accurate semantic.

Worth noting for anyone touching this path: a raw `IllegalArgumentException`/`IllegalStateException` can NOT be used here. `validateFeatureCompatibility` runs inside the `@Repository`-proxied `save()`, so Spring's persistence-exception translation (`EntityManagerFactoryUtils.convertJpaAccessExceptionIfPossible`) rewraps those two into `InvalidDataAccessApiUsageException`, which is unhandled → HTTP 500 → the client maps 500 to `CommitStateUnknownException` (a system error). `RequestValidationFailureException` is a plain domain exception that isn't translated, so it reaches its 400 handler. (d15662e5)

services/tables/src/main/java/com/linkedin/openhouse/tables/api/validator/impl/OpenHouseTablesApiValidator.java
```diff
@@ -148,6 +151,7 @@ public void validateCreateTable(
       throw new RequestValidationFailureException(validationFailures);
     }
     validatePolicies(createUpdateTableRequestBody);
+    validateFeatureCompatibility(createUpdateTableRequestBody);
```
mkuchenbecker: Lets make the features that are not compatible a parameter of this function as a literal.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/TablesControllerTest.java
```diff
@@ -942,97 +941,14 @@ public void testStagedReplaceFailsWhenRtasDisabled() {
...
-    RequestAndValidateHelper.deleteTableAndValidateResponse(mvc, table);
-  }
+  // The replace-time gate in validateReplaceTable also rejects a REPLACE on a table that has WAP or
+  // replication enabled. That branch is defense-in-depth for when the 'replace.enabled' opt-in flag
+  // is removed at GA (at which point every table implicitly allows RTAS and the replace-time check
+  // becomes the durable protection). It is intentionally not exercised end-to-end here: the
+  // enable-time feature-compatibility validator forbids ever creating a table that has RTAS
+  // enabled alongside WAP or replication, so that coexistence state is not reachable through the
+  // API. Enable-time rejection is covered by TablesValidatorTest and the black-box
+  // RtasWapExclusivityTest.
```
mkuchenbecker: Why did you add a stateful comment to this file? You are justifying a delta with a durable artifact.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/TablesControllerTest.java
```diff
@@ -942,97 +941,14 @@ public void testStagedReplaceFailsWhenRtasDisabled() {
     RequestAndValidateHelper.deleteTableAndValidateResponse(mvc, table);
   }
 
-  @SneakyThrows
```
mkuchenbecker, line 945: Leave these tests. Why you removed them is beyond me. Defense in depth, we added API validation which is handlebars on this function. DO not expand the problem beyond the description. DRY does not apply with respect to validation.

cluster/configs/src/main/java/com/linkedin/openhouse/cluster/configs/ClusterProperties.java
```diff
@@ -96,4 +96,11 @@ public class ClusterProperties {
   // string
   @Value("${cluster.tables.allowed-client-name-values:}")
   private List<String> allowedClientNameValues;
+
+  // Escape hatch for the enable-time feature-compatibility validation (for example the rule that a
+  // table cannot enable RTAS alongside WAP or replication). Enabled by default; set to false to
+  // allow requests that enable otherwise mutually-exclusive features, e.g. to exercise downstream
+  // behavior that can only be reached once such a table exists.
+  @Value("${cluster.tables.feature-compatibility-validation-enabled:true}")
+  private boolean featureCompatibilityValidationEnabled;
```
mkuchenbecker: This should be a table property not a cluster property.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/CatalogConstants.java
```diff
@@ -34,6 +34,9 @@ public final class CatalogConstants {
 
   public static final String WAP_ENABLED_TABLE_PROP = "write.wap.enabled";
 
+  public static final String FEATURE_COMPATIBILITY_VALIDATION_ENABLED_TABLE_PROP =
+      "feature.compatibility.validation.enabled";
```
mkuchenbecker: Reverse this to disabled. This is rolling out default enabled and you must explicitly disable validation. Either that or we need a clear default resolution.

PR #660, opened 2026-07-24
Refactor(spark-3.5): Make OpenHouse SQL extensions standalone 

mkuchenbecker:
## Summary
Gives the spark-3.5 module its own copy of the OpenHouse SQL-extension machinery so the grammar can evolve for spark-3.5. 

Previously spark-3.5 took a dependency on the 3.1 code to avoid duplication. That coupling meant the SQL grammar could not change for spark-3.5 without also changing spark-3.1.

This change will mean that we will explicitly release grammer on each version. 

## Changes
- [ ] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [x] Refactoring
- [ ] Documentation

1. Copy the grammar + 7 borrowed classes into spark-3.5.
2. Add the ANTLR generation task (antlr4 4.7.1, matching spark-3.1) and point `srcDirs`
  at the module's own generated dir.
3. Drop the spark-3.1 module dependency.

## Testing Done
- [x] No behavior change — existing spark-3.5 statement tests pass unmodified against
  the standalone module.

```
./gradlew :integrations:spark:spark-3.5:openhouse-spark-3.5-itest:statementTest
# BUILD SUCCESSFUL — all pre-existing statement tests pass
```

# Additional Information
No breaking changes; no client-facing changes.

### DIFF
--- a/integrations/spark/spark-3.5/openhouse-spark-runtime/build.gradle
+++ b/integrations/spark/spark-3.5/openhouse-spark-runtime/build.gradle
@@ -8,6 +8,11 @@ plugins {
 ext {
   icebergVersion = rootProject.ext.iceberg_1_5_version
   sparkVersion = '3.5.2'
+  // Antlr grammar sources are owned by this module (spark-3.5 keeps its own copy of the
+  // grammar so it can evolve independently of spark-3.1).
+  antlrPackageDirPrefix = "com/linkedin/openhouse/spark/sql/catalyst/parser/extensions/"
+  antlrMainDir = "${projectDir}/src/main/antlr/${antlrPackageDirPrefix}"
+  antlrMainGeneratedSrcDir = "${project.buildDir}/generated-src/antlr/main/"
 }
 
 configurations {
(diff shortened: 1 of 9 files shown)


integrations/spark/spark-3.5/openhouse-spark-runtime/build.gradle
```diff
@@ -16,18 +21,40 @@ configurations {
...
+
+  antlr
 }
 
 // Set source for antlr generated directory
 sourceSets {
   main {
     java {
-      srcDirs += "${project(':integrations:spark:spark-3.1:openhouse-spark-runtime_2.12').buildDir}/generated-src/antlr/main"
+      srcDirs antlrMainGeneratedSrcDir
```
mkuchenbecker, line 32: The difference is directly compiling rather than using the 3.1 generation.

mkuchenbecker: Yes — that's intentional. The Spark 3.5 module now owns and runs ANTLR over its own `src/main/antlr/.../OpenhouseSqlExtensions.g4`, with `sourceSets.main.java.srcDirs` pointing at its own `build/generated-src/antlr/main` and `compileJava.dependsOn runAntlr`. Reusing the Spark 3.1 generated sources would keep Spark 3.5 coupled to the 3.1 module's build output/task ordering and would prevent the 3.5 grammar from diverging for 3.5-specific SQL extensions. This keeps the runtime self-contained while preserving the same generated parser package/shape.

integrations/spark/spark-3.5/openhouse-spark-runtime/src/main/java/com/linkedin/openhouse/spark/OpenHouseCatalog.java
```diff
@@ -0,0 +1,18 @@
...
+ * <p>Catalog can be instantiated as a Iceberg catalog, with following configurations:
+ * spark.sql.catalog.openhouse=org.apache.iceberg.spark.SparkCatalog
+ * spark.sql.catalog.openhouse.catalog-impl=com.linkedin.openhouse.spark.OpenHouseCatalog
+ * spark.sql.catalog.openhouse.metrics-reporter-impl=com.linkedin.openhouse.javaclient.OpenHouseMetricsReporter
+ * spark.sql.catalog.openhouse.uri=http://[openhouse service host]:[openhouse service port]
+ * spark.sql.catalog.openhouse.cluster=[openhouse cluster name]
+ *
+ * <p>It can be used in spark shell as follows: spark.sql("USE openhouse")
+ */
+public class OpenHouseCatalog extends com.linkedin.openhouse.javaclient.OpenHouseCatalog {}
```
aastha25, line 18: general comment - lets follow the same structure as the [java integration modules](https://github.com/linkedin/openhouse/tree/main/integrations/java/iceberg-1.5/openhouse-java-runtime/src/iceberg-version-specific/java/com/linkedin/openhouse/javaclient). any iceberg / spark version specific overrides sit in  `*-version-specific` directory structure. the classes would still have the same package name and can be bundled in the uber jar using [this gradle block](https://github.com/linkedin/openhouse/blob/main/integrations/java/iceberg-1.5/openhouse-java-runtime/build.gradle#L39)

integrations/spark/spark-3.5/openhouse-spark-runtime/build.gradle
teamurko:
```
ANTLR pinned at 4.7.1 while Spark 3.5 bundles a newer runtime (4.9.3).
The generated parser is produced by the 4.7.1 tool, and  fatJarPackagedDependencies  excludes  org.antlr  ("included in spark"),
so at runtime it will use Spark 3.5's ANTLR (4.9.3). 
A 4.7.1-generated serialized ATN can be incompatible with a 4.9.3 runtime.

Mitigating fact: this is not a regression — pre-PR, spark-3.5 already consumed spark-3.1's 4.7.1-generated sources, so behavior is unchanged.
Still worth confirming the itests actually exercise the parser on Spark 3.5
before merge (bumping the tool to match Spark's runtime would be the cleaner long-term fix).
```

teamurko (approved):

integrations/spark/spark-3.5/openhouse-spark-runtime/src/main/java/com/linkedin/openhouse/spark/OpenHouseCatalog.java
mkuchenbecker: That would share code that happened to be the same and every modification would propagate to both. The goal  is that 3.5 can float from 3.1 and any shared code is incidentally shared until the moment its shanged.

PR #664, opened 2026-07-26
Honor table feature toggles during creation

cbb330:
## Summary

Issue: table toggles only work for "alter table" ddl while we want them to work on "create table" ddl as well.

Table creation filtered feature-gated preserved properties because `allowKeyInCreation` called the advised `isKeyPreservedForTable` method through Spring self-invocation, bypassing the feature-toggle aspect. Invoke the table-aware check from the repository's Spring proxy before applying the create-only allowlist fallback.

This makes CREATE consistent with ALTER while retaining support for extension-defined properties that are writable only during creation.

## Changes

- [x] Bug Fixes
- [x] Tests

- Evaluate `isKeyPreservedForTable` through the proxied `PreservedKeyChecker` during creation.
- Retain `allowKeyInCreation` as the fallback for create-only preserved properties.
- Add a Spring integration regression test using a real active table feature toggle.
- Update the default-file-format test to model preserved/toggled state on the table-aware method.

## Testing Done

- Verified the new regression test fails without the production change and passes with it.
- `:services:tables:test` — 463 tests passed.
- `:services:tables:spotlessCheck`
- `:services:tables:checkstyleMain`
- `:services:tables:checkstyleTest`
- Manually reproduced the original behavior against a live OpenHouse namespace: feature-gated properties survived ALTER but were silently filtered during CREATE.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/repository/impl/OpenHouseInternalRepositoryImpl.java
@@ -469,7 +469,10 @@ public class OpenHouseInternalRepositoryImpl implements OpenHouseInternalReposit
     // Populate non-preserved keys, mainly user defined properties.
     Map<String, String> propertiesMap =
         tableDto.getTableProperties().entrySet().stream()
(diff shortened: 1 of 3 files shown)


mkuchenbecker:
Can you articulate the bug / issure more strongly? Create and Alter user a different path?

kamanavishnu (approved):

PR #666, opened 2026-07-31
Add self-service overrides to table feature toggles

cbb330:
## Summary
Problem: I have a feature which I want to ramp on the server. but I also don't want to prevent table owners from self-serve opting in. A generic function to handle that overlap doesn't exist today. 

Extend the existing `TableFeatureToggle` with self-service table overrides while preserving its server-managed targeting API.

An explicit `<featureId>.enabled=true|false` table property wins. When the property is absent, activation delegates to the server toggle. Server rules now support trailing-`*` prefix matching independently for database and table names.

## Changes

- [x] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

Adds a binary-compatible default method to `TableFeatureToggle`:

```java
isFeatureActivatedWithOverride(TableDto tableDto, String featureId)
```

It is deliberately **not** an overload of `isFeatureActivated`. The two carry different safety contracts, and a distinct name makes the difference visible at the call site: authorization gates such as `enable_mor` decide whether a user may write a preserved table property, so they must keep using the server-only `isFeatureActivated(String, String, String)`. The override-honoring form reads a property the gated user can write.

An override that is neither `true` nor `false` fails closed: it is logged and the feature is treated as inactive. The gate is evaluated on the table-load path, so throwing would turn a typo like `read-bridge.enabled=flase` into a `400` and make the table unloadable.

Extends the existing toggle rule matcher while preserving exact and `*` matching:

- `tracking.events` matches exactly.
- `tracking_*.events_*` matches database and table prefixes.
- `*.*` matches every table.

## Testing Done

Ran:

```shell
JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew \
  :services:tables:test \
  --tests 'com.linkedin.openhouse.tables.toggle.TableFeatureToggleTest' \
  :services:housetables:test \
  --tests 'com.linkedin.openhouse.housetables.mock.WildcardTableToggleRuleMatcherTest' \
  -x CopyGitHooksTask
```

All 12 focused tests passed, covering server fallback, explicit opt-in and opt-out, fail-closed handling of unparseable overrides, exact matching, wildcard matching, and paired database/table prefix matching.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

This is an independent OSS foundation for the read-bridge stack in #645 and the corresponding `li-openhouse` implementation PRs. Feature-specific default derivation remains outside this PR.

Note for reviewers: the matcher change widens any existing `table_toggle_rule` row whose pattern ends in `*` but is not exactly `*`. Those previously matched nothing. Worth auditing HTS before merge.

### DIFF
--- a/services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/WildcardTableToggleRuleMatcher.java
+++ b/services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/WildcardTableToggleRuleMatcher.java
@@ -2,17 +2,24 @@ package com.linkedin.openhouse.housetables.services;
 
 import com.linkedin.openhouse.housetables.model.TableToggleRule;
 import org.springframework.stereotype.Component;
+import org.springframework.util.AntPathMatcher;
+import org.springframework.util.PathMatcher;
 
-/** An implementation of {@link TableToggleRuleMatcher} that supports '*' to match any entities */
+/**
+ * A {@link TableToggleRuleMatcher} matching database and table names as case-sensitive globs, so
+ * {@code *} matches any run of characters wherever it appears and {@code ?} matches one.
+ *
+ * <p>TODO: patterns are unvalidated because rules are inserted straight into MySQL, so a malformed
+ * pattern is only discovered by matching nothing. Validation belongs with a rule-write path, which
+ * does not exist yet.
+ */
 @Component
 public class WildcardTableToggleRuleMatcher implements TableToggleRuleMatcher {
+  private static final PathMatcher MATCHER = new AntPathMatcher();
+
   @Override
   public boolean matches(TableToggleRule rule, String tableId, String databaseId) {
-    boolean tableMatches =
-        rule.getTablePattern().equals("*") || rule.getTablePattern().equals(tableId);
-    boolean databaseMatches =
-        rule.getDatabasePattern().equals("*") || rule.getDatabasePattern().equals(databaseId);
-
-    return tableMatches && databaseMatches;
+    return MATCHER.match(rule.getTablePattern(), tableId)
+        && MATCHER.match(rule.getDatabasePattern(), databaseId);
   }
 }
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/TableFeatureToggle.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/TableFeatureToggle.java
@@ -1,14 +1,65 @@
 package com.linkedin.openhouse.tables.toggle;
 
-/** Interface to check if a feature is toggled-on for a table */
+import com.linkedin.openhouse.tables.model.TableDto;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Interface to check if a feature is toggled-on for a table.
+ *
+ * <p>TODO: the two forms below differ in who may influence the decision, which today is conveyed by
+ * their names and javadoc rather than enforced. The intended model declares that per feature
+ * instead of per call site — {@code TableFeature.capability(id)} vs {@code
+ * TableFeature.rollout(id)} — so a permission-bearing feature cannot acquire self-service opt-in by
+ * calling the wrong method. The same change would carry a decision's cause for metrics, give rules
+ * an explicit effect, priority and expiry rather than presence-implies-active, and evaluate rules
+ * from a locally replicated snapshot so a table read no longer blocks on HouseTables.
+ */
 public interface TableFeatureToggle {
+  Logger LOG = LoggerFactory.getLogger(TableFeatureToggle.class);
+
+  /** Suffix appended to a feature id to form its self-service table property. */
+  String ENABLED_PROPERTY_SUFFIX = ".enabled";
+
   /**
-   * Determine if given feature is activated for the table.
+   * Determines the server-side activation decision for a table.
    *
-   * @param databaseId databaseId
-   * @param tableId tableId
-   * @param featureId featureId
-   * @return True if the feature is activated for the table.
+   * <p>Authorization gates — features deciding whether a user may write an otherwise preserved
+   * property, like {@code enable_mor} — must use this form, since the table property honored by
+   * {@link #isFeatureActivatedWithOverride(TableDto, String)} is writable by the user being gated.
    */
   boolean isFeatureActivated(String databaseId, String tableId, String featureId);
+
+  /**
+   * Determines activation, letting the table override the server-side decision.
+   *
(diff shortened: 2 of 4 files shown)


services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/WildcardTableToggleRuleMatcher.java
```diff
@@ -3,16 +3,22 @@
...
-    boolean databaseMatches =
-        rule.getDatabasePattern().equals("*") || rule.getDatabasePattern().equals(databaseId);
+    return matchesPattern(rule.getTablePattern(), tableId)
+        && matchesPattern(rule.getDatabasePattern(), databaseId);
+  }
 
-    return tableMatches && databaseMatches;
+  private static boolean matchesPattern(String pattern, String value) {
+    if (pattern.endsWith("*")) {
+      return value.startsWith(pattern.substring(0, pattern.length() - 1));
```
kamanavishnu: Do we need  a case sensitive match here ? Given we are opening this up to users

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/TableFeatureToggle.java
```diff
@@ -1,14 +1,55 @@
...
-   * Determine if given feature is activated for the table.
+   * Determines the server-side activation decision for a table.
    *
-   * @param databaseId databaseId
-   * @param tableId tableId
-   * @param featureId featureId
-   * @return True if the feature is activated for the table.
+   * <p>Authorization gates — features deciding whether a user may write an otherwise preserved
+   * property, like {@code enable_mor} — must use this form, since the table property honored by
+   * {@link #isFeatureActivatedWithOverride(TableDto, String)} is writable by the user being gated.
```
kamanavishnu, line 30: The javadoc correctly warns that authorization-gating features (e.g. `enable_mor`) must not use `isFeatureActivatedWithOverride` since the backing property is user-writable. Is there any code-level enforcement (allowlist/denylist of featureIds, a marker annotation, etc.) preventing a future caller from accidentally wiring a security-sensitive toggle through this override path, or is the javadoc the only safeguard today?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/WildcardTableToggleRuleMatcher.java
```diff
@@ -3,16 +3,22 @@
...
-    boolean tableMatches =
-        rule.getTablePattern().equals("*") || rule.getTablePattern().equals(tableId);
-    boolean databaseMatches =
-        rule.getDatabasePattern().equals("*") || rule.getDatabasePattern().equals(databaseId);
+    return matchesPattern(rule.getTablePattern(), tableId)
+        && matchesPattern(rule.getDatabasePattern(), databaseId);
+  }
 
-    return tableMatches && databaseMatches;
+  private static boolean matchesPattern(String pattern, String value) {
```
kamanavishnu: Only a trailing `*` is treated as a wildcard here — a `*` anywhere else in the pattern (e.g. `foo*bar`) falls through to the exact-match branch and is treated as a literal character, which could surprise a rule author expecting glob-style matching. Should rule creation validate/reject patterns with a `*` that isn't at the end, so this silent literal-match fallback can't happen?

cbb330: Good catch — this is now moot: the hand-rolled parsing is replaced with Spring `AntPathMatcher`, so `*` and `?` are honored anywhere in the pattern and `foo*bar` matches as a glob rather than silently falling back to a literal. Added a test for a mid-pattern wildcard.

On validating at rule creation: there is no write path to hook. `ToggleStatusesController` exposes only `@GetMapping`, and `ToggleStatusesRepositoryImpl.save` throws `UnsupportedOperationException` — rules are inserted directly into `table_toggle_rule`. I left a TODO on the matcher noting that pattern validation belongs with a rule-write API when one exists.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/WildcardTableToggleRuleMatcher.java
cbb330: Keeping it case-sensitive, deliberately. That is the pre-existing behavior (`pattern.equals(value)`), and `AntPathMatcher` is case-sensitive by default, so this PR does not change it.

Loosening it would silently widen every existing rule, including `enable_mor`, which is exactly the kind of change that should be explicit rather than a side effect of a matcher rewrite. Table and database ids are also matched exactly elsewhere (HTS keys), so case-insensitive matching here would be inconsistent with how identity is treated. Added a test pinning the behavior.

services/tables/src/main/java/com/linkedin/openhouse/tables/toggle/TableFeatureToggle.java
cbb330: Correct — today the javadoc and the method name are the only safeguards. I considered enforcing it in this PR and decided against it, because the enforcement worth having is a change to how features are declared, not a check bolted onto this method.

The intended model, which I have added as a TODO on the interface: a feature declares its kind once, at declaration rather than at each call site.

```java
TableFeature.capability("enable_mor")   // server decides alone
TableFeature.rollout("read-bridge")     // owner may opt in or out
```

Resolution then takes a `TableFeature`, so a capability has no override path to reach — the hazard becomes unrepresentable rather than merely discouraged, and switching a feature would be a one-line, obviously-suspicious diff.

A runtime allowlist is possible sooner: `TblPropsToggleRegistry` already enumerates exactly the features that gate preserved properties, so the override path could reject any feature id registered there. That needs a reverse lookup on the registry and moves resolution into a bean, so I would rather do it together with the typed features than half of it now.

The same follow-up would carry the decision cause for metrics (server ramp vs self-service opt-in), give rules an explicit effect/priority/expiry instead of presence-implies-active, and evaluate from a locally replicated rule snapshot so table reads stop blocking on HouseTables. Happy to open that as the next PR if the direction looks right to you.

kamanavishnu (approved):
Re-reviewed after the latest commit. All three open threads are resolved:
- Wildcard matching now uses Spring `AntPathMatcher`, fixing the mid-pattern `*` fallback and adding a pinning test.
- Case-sensitivity is intentionally preserved (consistent with pre-existing behavior and HTS key matching), with a test locking it in.
- The authorization-gate safeguard is javadoc-only today; agreed this is fine for now given the proposed typed `TableFeature.capability()` vs `.rollout()` follow-up.

LGTM.

PR #668, opened 2026-08-03
[default values] ReadBridge: mark bridge failures unrecoverable on Iceberg's retry path

cbb330:
## Summary

> [!IMPORTANT]
> **Stacked on #645** — please review that first. Until it merges, the diff
> here also shows its commits; the only commit belonging to this PR is
> `ReadBridge: mark bridge failures unrecoverable on Iceberg's retry path`.
> Once #645 merges this collapses to a small client-runtime + test diff. Draft until then.

`loadMetadata` is the loader `BaseMetastoreTableOperations` wraps in
`Tasks.retry(20)` with exponential backoff. That retry exists for the metadata
**file read**, which fails transiently — a network blip, a file not yet visible —
and can succeed on a later attempt.

Decode and apply are deterministic: a malformed config fails the same way every
attempt. Leaving those failures retryable therefore gave a deterministic error the
retry policy of a transient one:

- 21 attempts, ~87s of backoff (100 + 400 + 1600 + 5000×17 ms)
- 21 re-reads and re-parses of the metadata file from storage
- to reproduce the error already available on the first attempt

Every reader of the table pays it simultaneously, so one malformed entry becomes
a read stall plus a storage read storm instead of a fast, clear error. The retry
loop also *loses* the real error: it reports whatever the last attempt threw.

## Changes

- [x] Internal API Changes
- [x] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [x] Refactoring
- [ ] Documentation
- [x] Tests

Keep `from` / `apply` inside `loadMetadata` (decode **before** the file read so a
bad config never touches storage), and wrap `IllegalStateException` as Iceberg's
`Tasks.UnrecoverableException`. Iceberg already stops `Tasks.retry` on that type,
so no `doRefresh` decode field or `AtomicReference<ReadBridge>` is needed.

| | when | retryable |
|---|---|---|
| `ReadBridge.from(config)` | before file IO in `loadMetadata` | no (`UnrecoverableException`) |
| metadata file read / parse | inside `loadMetadata` | yes |
| `bridge.apply(metadata)` | after file IO in `loadMetadata` | no (`UnrecoverableException`) |

## Testing Done

- [ ] Manually Tested on local docker setup.
- [ ] No tests added or updated.
- [ ] Some other form of testing like staging or soak time in production.

New regression test `testMalformedConfigFailsBeforeTouchingStorage` asserts a
`Tasks.UnrecoverableException` whose cause is `IllegalStateException`, **and**
`verifyNoInteractions(mockFileIO)` — failing before any storage access is the
property that matters.

```shell
JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew \
  :integrations:java:iceberg-1.2:openhouse-java-itest:test \
  --tests '*ReadBridge*' \
  --tests '*OpenHouseTableOperationsTest.testMalformed*' \
  --tests '*OpenHouseTableOperationsTest.testDoRefresh*'
```

## Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

Kept out of #645 deliberately, so that PR stays as reviewed and this defect is
reviewable on its own.

Downstream stack (rebased onto this tip): #674 → #675.

### DIFF
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
@@ -128,11 +128,13 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
   }
 
   /**
-   * Loads the table metadata at the given location. Defaults to the stock parser; subclasses may
-   * override to transform the metadata (e.g. attach column defaults) as it loads.
+   * Loads the table metadata from storage, then runs the read-time {@link ReadBridge} over it using
+   * the per-table behavior the server stamped onto {@link #currentConfig()}. Fail-closed:
+   * absent/off/unparseable config leaves the raw metadata untouched.
    */
   protected TableMetadata loadMetadata(String metadataLocation) {
-    return TableMetadataParser.read(io(), metadataLocation);
+    TableMetadata raw = TableMetadataParser.read(io(), metadataLocation);
+    return ReadBridge.apply(raw, currentConfig());
   }
 
   @Override
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import lombok.AccessLevel;
@@ -64,8 +65,8 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
   private String cluster;
 
   /**
-   * Config last applied to {@code current()}, or {@code null}. Bound in {@link #loadMetadata} after
-   * apply so skip-reload cannot desync stamps from overlays.
+   * Config last applied to {@code current()}, or {@code null}. Bound after Iceberg accepts a reload
+   * so skip-reload and a failed UUID check cannot desync stamps from overlays.
    */
   private final AtomicReference<Map<String, String>> config = new AtomicReference<>();
 
@@ -116,10 +117,21 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
           "Cannot find table %s after refresh, maybe another process deleted it", tableName());
     }
     Map<String, String> fetched = tableResponse.map(GetTableResponseBody::getConfig).orElse(null);
-    // Bind config in loadMetadata after apply. Iceberg skips the loader when tableLocation is
-    // unchanged; leaving config alone keeps stamps paired with the in-memory overlay.
+    AtomicBoolean loaded = new AtomicBoolean();
+    // Iceberg skips the loader when tableLocation is unchanged. UUID is checked after the loader
+    // returns; bind only if this call actually accepted a reload.
     super.refreshFromMetadataLocation(
-        tableLocation.orElse(null), null, 20, location -> loadMetadata(location, fetched));
+        tableLocation.orElse(null),
+        null,
+        20,
+        location -> {
+          TableMetadata bridged = loadMetadata(location, fetched);
+          loaded.set(true);
+          return bridged;
+        });
+    if (loaded.get()) {
+      config.set(fetched);
+    }
     log.debug("Calling doRefresh succeeded");
   }
 
@@ -127,8 +139,7 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
    * Decode {@code fetched}, then read the metadata file, then overlay.
    *
    * <p>Decode first so a bad config never hits storage. {@link IllegalStateException} is wrapped as
-   * {@link Tasks.UnrecoverableException} so Iceberg's retry loop does not re-read the file. {@link
-   * #currentConfig()} is updated only after apply succeeds.
+   * {@link Tasks.UnrecoverableException} so Iceberg's retry loop does not re-read the file.
    */
   protected TableMetadata loadMetadata(String metadataLocation, Map<String, String> fetched) {
     final ReadBridge bridge;
@@ -139,9 +150,7 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
     }
     TableMetadata raw = TableMetadataParser.read(io(), metadataLocation);
     try {
-      TableMetadata bridged = bridge.apply(raw);
-      config.set(fetched);
-      return bridged;
+      return bridge.apply(raw);
     } catch (IllegalStateException e) {
       throw new Tasks.UnrecoverableException(e);
     }
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
@@ -129,8 +129,9 @@ public class OpenHouseTableOperations extends BaseMetastoreTableOperations {
 
   /**
    * Loads the table metadata from storage, then runs the read-time {@link ReadBridge} over it using
-   * the per-table behavior the server stamped onto {@link #currentConfig()}. Fail-closed:
-   * absent/off/unparseable config leaves the raw metadata untouched.
+   * the per-table behavior the server stamped onto {@link #currentConfig()}. Absent config, or
+   * config carrying nothing to bridge, leaves the raw metadata untouched; a malformed known entry
+   * fails loud (see {@link ReadBridge}).
    */
   protected TableMetadata loadMetadata(String metadataLocation) {
     TableMetadata raw = TableMetadataParser.read(io(), metadataLocation);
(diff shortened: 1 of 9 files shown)


cbb330:
**Stack update:** this PR is step 2 of the read-bridge follow-ups (base retargeted onto #645).

1. #645 — substrate
2. **This PR (#668)** — decode off Iceberg retry path
3. #674 — policy/ramp in OpenHouse
4. apply follow-up — `ReadBridge.apply` (stacks on #674)

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -77,9 +68,7 @@ static Map<Integer, JsonNode> columnDefaults(Map<String, String> config) {
         int fieldId = Integer.parseInt(entry.getKey().substring(COLUMN_DEFAULT_PREFIX.length()));
         byFieldId.put(fieldId, MAPPER.readTree(entry.getValue()));
       } catch (RuntimeException | JsonProcessingException e) {
-        // The server encoder stamps an int field-id and a JsonNode value that round-trips through
-        // readTree, so reaching here means an encoder bug or transport corruption, not an expected
-        // state. Fail loud so it is caught, rather than silently reading NULL.
+        // Known keys are stamped as int field-id + JSON; anything else is a bug.
```
mkuchenbecker, line 71: why did the comment change?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -9,62 +9,53 @@
...
-   * entry, the server encoder guarantees an integer field-id and a value that round-trips through
-   * {@code readTree}, so a non-integer field-id or an unparseable value is an encoder bug or
-   * transport corruption — it throws rather than degrading. Unknown keys are ignored above.
-   * Package-visible for unit testing in isolation.
-   */
-  static Map<Integer, JsonNode> columnDefaults(Map<String, String> config) {
+  Map<Integer, JsonNode> columnDefaults() {
+    return columnDefaults;
+  }
+
```
mkuchenbecker, line 57: why did the java doc drop?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -113,29 +111,49 @@ public void doRefresh() {
                 WebClientRequestException.class,
                 e -> Mono.error(new WebClientRequestWithMessageException(e)))
             .blockOptional();
-    // Capture the server-stamped per-table config so subclasses can gate read-time behavior via
-    // currentConfig(); absent => null. Side-channel only: never sent back on writes.
-    this.config.set(tableResponse.map(GetTableResponseBody::getConfig).orElse(null));
```
mkuchenbecker, line 118: why did this drop?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: It was a style pass.

I made commit 57a91710 titled “Tighten read-bridge comments to why, not design essays.” The same message is on #674.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: from a style pass, no behavior change. I wanted comments to not look like a wall of design text and be hard to review. 

here is just a getter

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: if you want, I can apply this "style" pass onto 645 so it doesn't look like a diff. 

I didn't do that originally because I figured 645 is nearly done with review, so didn't want to touch it.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
cbb330: The set moved. because we don’t bind on every GET anymore.

#645 wrote config as soon as the table-load response arrived, then called Iceberg. Iceberg can skip the metadata loader when the location is unchanged, and it UUID-checks after the loader. So, binding first meant currentConfig() could be a new stamp while current() was still the old metadata. It would mismatch config and metadata. 

#668 binds only after a reload is actually accepted (config.set(fetched) after refreshFromMetadataLocation, and only if the loader ran). Skip-reload and a failed UUID check keep the old pair.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -113,29 +111,55 @@ public void doRefresh() {
...
+    super.refreshFromMetadataLocation(
+        tableLocation.orElse(null),
+        null,
+        20,
+        location -> {
+          TableMetadata bridged = loadMetadata(location, fetched);
+          loaded.set(true);
+          return bridged;
+        });
+    if (loaded.get()) {
```
abhisheknath2011: Concurrent thread could reset/reinitialize loaded property by the time code reaches here.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -113,29 +111,55 @@ public void doRefresh() {
...
+      config.set(fetched);
+    }
```
abhisheknath2011, line 134: Went through this code and looks good and should not break the existing behaviour.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -113,29 +111,55 @@ public void doRefresh() {
...
+   *
+   * <p>Decode first so a bad config never hits storage. {@link IllegalStateException} is wrapped as
+   * {@link Tasks.UnrecoverableException} so Iceberg's retry loop does not re-read the file. The
+   * wrap includes {@link #tableName()} so a Spark job over many tables can tell which one failed.
    */
-  protected TableMetadata loadMetadata(String metadataLocation) {
+  protected TableMetadata loadMetadata(String metadataLocation, Map<String, String> fetched) {
+    final ReadBridge bridge;
+    try {
+      bridge = ReadBridge.from(fetched);
```
abhisheknath2011, line 148: If config is null or empty can we skip this.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
```diff
@@ -113,29 +111,55 @@ public void doRefresh() {
...
+    } catch (IllegalStateException e) {
+      throw unrecoverableBridge(e);
+    }
     TableMetadata raw = TableMetadataParser.read(io(), metadataLocation);
-    return ReadBridge.apply(raw, currentConfig());
+    try {
+      return bridge.apply(raw);
```
abhisheknath2011, line 154: We can do bridge not null and then apply.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
cbb330: thanks, yep. No column-default stamp → from is a no-op, metadata is whatever Iceberg loaded. Same as today.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
cbb330: It is already skipped inside from. We let ReadBridge decide whether there is anything to apply because null/empty is only one of those cases. Putting the skip here would miss the others.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/OpenHouseTableOperations.java
cbb330: I'd rather keep this. above `from` never returns null. nothing to overlay is a no-op ReadBridge object (INERT), and apply on that just returns the `raw` metadata Iceberg already loaded.

if (bridge != null) would never skip unless we changed from to return null.

this also keeps all of the conditions in one place , similar to the above answer, in case there are other conditions that return the `raw` metadata.

abhisheknath2011 (approved):

PR #674, opened 2026-08-12
[default values] Move read-bridge column-default ramp into OpenHouse

cbb330:
## Summary

Depends on [#668](https://github.com/linkedin/openhouse/pull/668) (stacked on [#645](https://github.com/linkedin/openhouse/pull/645)).

Move **column-default rollout policy** into OpenHouse so a deployment only supplies the defaults themselves.

| Concern | Owner after this PR |
|---|---|
| Feature id, self-serve property, wire keys | OpenHouse (`ReadBridgeConfigResolver`) |
| Per-table ramp / kill | OpenHouse (`TableFeatureToggle` + table property) |
| “What are this table’s defaults?” | Deployment (`ColumnDefaultsSource`, e.g. LI ASL) |

**Ids (one token, capability-scoped — not bare `read-bridge`):**
- Feature id: `read-bridge.column-default`
- Self-serve property: `read-bridge.column-default.enabled`
- Config prefix: `openhouse.read-bridge.column-default.<fieldId>`

## Resolve flow

On get/create/update, `withConfig` calls `ReadBridgeConfigResolver.resolve(tableDto)`:

1. **No source** (`ColumnDefaultsSource.NONE`) → empty config, **no HTS call**
2. **Ramp off** (property `false`, or HTS not `ACTIVE`, or HTS error) → empty config; source not consulted
3. **Ramp on** → ask source → stamp `openhouse.read-bridge.column-default.<fieldId> = <single-value-json>`

Self-serve override (`isFeatureActivatedWithOverride`):
- property `true` / `false` → table wins, **HTS skipped**
- property absent → exact HTS lookup `(databaseId, tableId, featureId)`
- unparseable property → fail closed (off)

## Changes

- [x] Internal API Changes — `ApiConfig` wires resolver via `ObjectProvider<ColumnDefaultsSource>` (no `@ConditionalOnMissingBean` noop / no `@Primary`); `resolve(TableDto)` only
- [x] New Features — capability-specific ramp + self-serve `*.enabled` override; fail-open on toggle lookup errors
- [x] Refactoring — `ColumnDefaultsSource` is data-only; ramp is not the deployment’s job
- [x] Documentation — javadoc spells exact-match HTS (no glob `*` / `*` today) and when HTS is hit
[...]

PR #678, opened 2026-08-12
[default values] Handshake and drop read-bridge overlays on the server

cbb330:
## Summary

Depends on [#674](https://github.com/linkedin/openhouse/pull/674) (policy), stacked on [#668](https://github.com/linkedin/openhouse/pull/668) → [#645](https://github.com/linkedin/openhouse/pull/645). [#679](https://github.com/linkedin/openhouse/pull/679) (`ReadBridge.apply`) restacks on this PR.

Read-bridge overlays must not persist. This PR puts the drop on the **server** *before* apply lands, so there is no deployable cut where overlays are applied but not stripped.

Default-aware clients send `initial-default` on stamped field-ids (the handshake). The server gates Type 1 / Type 2, then strips those keys before Iceberg commit. Unstamped ids keep the writer's defaults.

Until #679, apply is still a no-op; PUTs only carry `initial-default` if the writer set it. The server is already safe: drop is a no-op when the handshake is absent, and Type 2 fail-closes unaware rewrites on ramped tables.

## How strip protection works

| Gate | When | Result |
|---|---|---|
| Type 1 | Ramped table; a previously stamped field-id is still in the schema but missing from incoming stamps | 400 `COLUMN_DEFAULT_REMOVED` |
| Type 2 | Ramped rewrite (`replace` / `overwrite` on main) while previous stamps are nonempty | 400 `COLUMN_DEFAULT_REWRITE` unless each remaining id's `initial-default` JSON equals the stamp |
| Unusable | Ramped write; source throw / unreadable schema / unreadable snapshots | 400 `COLUMN_DEFAULT_UNUSABLE` |
| Drop | After gates pass | Remove `initial-default` on stamped ids (same `findParents("id")` walk as the client) |

Unramped tables are a no-op. GET `resolve()` still fail-opens; only the write path is fail-closed. Awareness is `JsonNode.equals` against the stamped config, not key presence.

OSS never parses ASL. `ColumnDefaultsSource` is the seam (`NONE` in OSS); LinkedIn fills it. Type 1 compares stamped field-id sets from the resolver.

## Changes

- [x] New Features — `ReadBridgeStripProtection.prepare` on PUT table / snapshots
- [x] Internal API Changes — resolver write APIs `stampedColumnDefaults` / `isRampedForCommit`
- [x] Tests — unit Type 1/2/unusable, mock bean for `MockTablesApplication`, HTTP e2e that GET `initialDefault()` is null after overlay PUT

## Testing Done

- Local (Java 17): `./gradlew :services:tables:test --tests '*ReadBridgeStripProtectionTest' --tests '*ReadBridgeConfigResolverTest' --tests '*ReadBridgeColumnDefaultE2ETest'`

## Additional Information

### Stack

1. [#645](https://github.com/linkedin/openhouse/pull/645) — substrate (seam + encode + decode hook)
2. [#668](https://github.com/linkedin/openhouse/pull/668) — decode off Iceberg’s retry path
3. [#674](https://github.com/linkedin/openhouse/pull/674) — policy / ramp in OpenHouse
4. **This PR** — handshake on PUT, Type 1/2, drop before persist
5. [#679](https://github.com/linkedin/openhouse/pull/679) — `ReadBridge.apply` (merge after this PR)
6. [#681](https://github.com/linkedin/openhouse/pull/681) — Spark catalog itest (overlay)

### Rollback Plan

Revert this PR. After #679, clients send overlays; without the server drop they persist in Iceberg metadata.

### DIFF
--- a/services/common/src/main/java/com/linkedin/openhouse/common/exception/UnsupportedClientOperationException.java
+++ b/services/common/src/main/java/com/linkedin/openhouse/common/exception/UnsupportedClientOperationException.java
@@ -22,5 +22,7 @@ public class UnsupportedClientOperationException extends RuntimeException {
     GRANT_ON_LOCKED_TABLES,
     LOCKED_TABLE_OPERATION,
     RTAS_DISABLED,
+    COLUMN_DEFAULT_REWRITE,
+    COLUMN_DEFAULT_REMOVED,
   }
 }
--- a/services/common/src/main/java/com/linkedin/openhouse/common/exception/UnsupportedClientOperationException.java
+++ b/services/common/src/main/java/com/linkedin/openhouse/common/exception/UnsupportedClientOperationException.java
@@ -24,5 +24,6 @@ public class UnsupportedClientOperationException extends RuntimeException {
     RTAS_DISABLED,
     COLUMN_DEFAULT_REWRITE,
     COLUMN_DEFAULT_REMOVED,
+    COLUMN_DEFAULT_UNUSABLE,
   }
 }
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
@@ -1,7 +1,6 @@
 package com.linkedin.openhouse.tables.readbridge;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.linkedin.openhouse.common.exception.UnsupportedClientOperationException;
 import com.linkedin.openhouse.tables.model.TableDto;
 import com.linkedin.openhouse.tables.toggle.TableFeatureToggle;
 import java.util.Collections;
@@ -47,14 +46,15 @@ public class ReadBridgeConfigResolver {
    * ramped. Toggle or source failure throws — the write path fail-closes; {@link #resolve} does
    * not.
    *
-   * @throws UnsupportedClientOperationException if the source or ramp lookup cannot answer
+   * @throws ColumnDefaultException if the source or ramp lookup cannot answer
    */
-  public Map<Integer, String> stampedColumnDefaults(TableDto tableDto) {
+  public Map<Integer, String> stampedColumnDefaults(TableDto tableDto)
+      throws ColumnDefaultException {
     Objects.requireNonNull(tableDto, "tableDto");
     try {
       return columnDefaultsByFieldId(tableDto);
     } catch (RuntimeException e) {
-      throw unusable(tableDto, e);
+      throw ColumnDefaultException.unusable(tableDto, e);
     }
   }
 
@@ -62,14 +62,14 @@ public class ReadBridgeConfigResolver {
    * Write-path ramp. Toggle failure throws. {@code ColumnDefaultsSource.NONE} is never ramped, so
    * the toggle is not consulted.
    *
-   * @throws UnsupportedClientOperationException if the ramp lookup cannot answer
+   * @throws ColumnDefaultException if the ramp lookup cannot answer
    */
-  public boolean isRampedForCommit(TableDto tableDto) {
+  public boolean isRampedForCommit(TableDto tableDto) throws ColumnDefaultException {
     Objects.requireNonNull(tableDto, "tableDto");
     try {
       return isColumnDefaultRamped(tableDto);
     } catch (RuntimeException e) {
-      throw unusable(tableDto, e);
+      throw ColumnDefaultException.unusable(tableDto, e);
     }
   }
 
@@ -124,54 +124,5 @@ public class ReadBridgeConfigResolver {
       return false;
     }
     return featureToggle.isFeatureActivatedWithOverride(tableDto, COLUMN_DEFAULT_FEATURE_ID);
-  }
-
-  static UnsupportedClientOperationException unusable(TableDto tableDto, Throwable cause) {
-    return unusable(tableDto, null, cause);
-  }
-
-  static UnsupportedClientOperationException unusable(
-      TableDto incoming, TableDto existing, Throwable cause) {
-    return unusable(incoming, existing, causeMessage(cause), cause);
-  }
-
-  static UnsupportedClientOperationException unusable(
-      TableDto incoming, TableDto existing, String reason, Throwable cause) {
-    UnsupportedClientOperationException thrown =
-        new UnsupportedClientOperationException(
-            UnsupportedClientOperationException.Operation.COLUMN_DEFAULT_UNUSABLE,
-            String.format(
-                "COLUMN_DEFAULT_UNUSABLE: OpenHouse could not validate column defaults on %s.%s, so"
-                    + " the commit was rejected. Retry. If it persists, contact the OpenHouse team"
-                    + " with the Spark application logs and the table metadata path: %s. Cause: %s",
-                incoming.getDatabaseId(),
-                incoming.getTableId(),
-                metadataPath(incoming, existing),
-                reason));
-    if (cause != null) {
-      thrown.initCause(cause);
-    }
-    return thrown;
-  }
-
-  private static String causeMessage(Throwable cause) {
-    if (cause == null || cause.getMessage() == null) {
-      return cause == null ? "unknown" : cause.toString();
-    }
-    return cause.getMessage();
-  }
-
-  private static String metadataPath(TableDto incoming, TableDto existing) {
-    if (incoming != null
-        && incoming.getTableLocation() != null
-        && !incoming.getTableLocation().isEmpty()) {
-      return incoming.getTableLocation();
-    }
-    if (existing != null
-        && existing.getTableLocation() != null
-        && !existing.getTableLocation().isEmpty()) {
-      return existing.getTableLocation();
-    }
-    return "unavailable";
   }
 }
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
@@ -1,11 +1,13 @@
 package com.linkedin.openhouse.tables.readbridge;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.linkedin.openhouse.common.exception.UnsupportedClientOperationException;
 import com.linkedin.openhouse.tables.model.TableDto;
 import com.linkedin.openhouse.tables.toggle.TableFeatureToggle;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import lombok.extern.slf4j.Slf4j;
 
 /**
@@ -27,12 +29,14 @@ public class ReadBridgeConfigResolver {
 
   public ReadBridgeConfigResolver(
       ColumnDefaultsSource columnDefaultsSource, TableFeatureToggle featureToggle) {
-    this.columnDefaultsSource = columnDefaultsSource;
-    this.featureToggle = featureToggle;
+    this.columnDefaultsSource =
+        Objects.requireNonNull(columnDefaultsSource, "columnDefaultsSource");
+    this.featureToggle = Objects.requireNonNull(featureToggle, "featureToggle");
   }
 
   /** Merges independently gated capabilities; empty when nothing is bridged. */
   public Map<String, String> resolve(TableDto tableDto) {
+    Objects.requireNonNull(tableDto, "tableDto");
     Map<String, String> config = new HashMap<>();
     config.putAll(columnDefaultConfig(tableDto));
     return config;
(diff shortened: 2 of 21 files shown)


services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
```diff
@@ -0,0 +1,302 @@
...
+   * Reject Type 1 / Type 2 violations, then strip stamped {@code initial-default} from {@code
+   * incoming} so they are not persisted. Returns {@code incoming} unchanged when there is nothing
+   * to check or drop.
+   */
+  public TableDto prepare(TableDto existing, TableDto incoming) {
+    if (incoming == null) {
+      return incoming;
+    }
+    try {
+      Map<Integer, String> previous =
```
abhisheknath2011:
```suggestion
      Map<Integer, String> previousStamped =
```
Can we name this as previousStamped so that it does not create confusion. Off course stamped if not empty.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
```diff
@@ -0,0 +1,302 @@
...
+      return stripStampedDefaults(previous, incomingStamped, incoming);
+    } catch (UnsupportedClientOperationException e) {
+      throw e;
+    } catch (IllegalStateException e) {
+      // Source/toggle/parse failures are ISE and must 400. A bug (NPE) must stay 500.
+      throw unusable(incoming, existing, e);
+    }
+  }
+
+  private void rejectRemovedDefaults(
```
abhisheknath2011, line 83: Can we add some comment on the logic here?

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
```diff
@@ -0,0 +1,302 @@
...
+                    + " omitted it. %s Column defaults cannot be removed or changed.",
+                incoming.getDatabaseId(),
+                incoming.getTableId(),
+                fieldLabel(schema, fieldId),
+                SUPPORTED_CLIENTS));
+      }
+    }
+  }
+
+  private void rejectUnawareRewrite(Map<Integer, String> previous, TableDto incoming) {
```
abhisheknath2011: Comment for the logic here.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
```diff
@@ -0,0 +1,302 @@
...
+  public ReadBridgeStripProtection(ReadBridgeConfigResolver resolver) {
+    this.resolver = resolver;
+  }
+
+  /**
+   * Reject Type 1 / Type 2 violations, then strip stamped {@code initial-default} from {@code
+   * incoming} so they are not persisted. Returns {@code incoming} unchanged when there is nothing
+   * to check or drop.
+   */
+  public TableDto prepare(TableDto existing, TableDto incoming) {
```
abhisheknath2011: Could re-ram reintroduce persistence bug? Strip is gated on the ramp toggle ( stampedColumnDefaults  returns empty when off). Turn the toggle off on a stamped table and clients still send  initial-default  → overlays persist again. Should we decouple the drop from the ramp gate (stripping is safe)?

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
```diff
@@ -0,0 +1,302 @@
...
+    String schema = strip(incoming.getSchema(), stripIds);
+    List<String> intermediates = strip(incoming.getNewIntermediateSchemas(), stripIds);
+    if (Objects.equals(schema, incoming.getSchema())
+        && Objects.equals(intermediates, incoming.getNewIntermediateSchemas())) {
+      return incoming;
+    }
+    return incoming.toBuilder().schema(schema).newIntermediateSchemas(intermediates).build();
+  }
+
+  private boolean isRewrite(TableDto incoming) {
```
abhisheknath2011: Type 2 only checks main branch. Branch (WAP) rewrites by unaware clients could bypass the gate. Can we confirm scope or add a comment noting the gap?

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
```diff
@@ -0,0 +1,302 @@
...
+
+  private static Snapshot snapshot(String json) {
+    try {
+      return SnapshotParser.fromJson(json);
+    } catch (RuntimeException e) {
+      throw new IllegalStateException("unreadable snapshot", e);
+    }
+  }
+
+  private static String strip(String schemaJson, Set<Integer> fieldIds) {
```
abhisheknath2011: Can we add comment on the strip method with usage?

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
cbb330: Renamed to `previousStamped`.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
cbb330: Added a why-note on the method: still-present columns must keep their default. unstamped/unramped must skip.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
cbb330: Added a why-note: overwrite/replace must send a matching `initial-default` or risk dropping default values from data files. Handshake is trust. Appends are not rewrites and can be ignored.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
cbb330: Added a why-note on `strip`: drop `initial-default` only on stamped field-ids so the handshake never lands in metadata.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
cbb330: Agreed. Strip was accidentally gated because it reused the stamp map.

It now drops every `initial-default` on write, including ramp-off leftovers and unstamped writer defaults.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
cbb330: Good catch, confirmed: Type 2 is main-only so a historical overwrite in the full snapshot list is not treated as this commit. WAP / named-branch overwrite is therefore not gated.

This PR would get a bit hairy to adapt, so we will need a follow up PR to address the branch-based defaults

#669 adds commit deltas (`add-snapshot` / `set-snapshot-ref`). This makes the logic simpler to wire those into this path and handle the default values. Tracked in https://github.com/linkedin/openhouse/issues/693 and I will follow up with a fix after this PR

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
```diff
@@ -38,18 +38,52 @@ public Map<String, String> resolve(TableDto tableDto) {
...
     }
-    if (!isColumnDefaultRamped(tableDto)) {
+    if (!isColumnDefaultRamped(tableDto, failClosed)) {
       return Collections.emptyMap();
     }
     Map<Integer, JsonNode> columnDefaults;
     try {
       columnDefaults = columnDefaultsSource.defaults(tableDto);
     } catch (RuntimeException e) {
+      if (failClosed) {
```
mkuchenbecker: Why does this handle a sneaky throw with a sneaky throw? I would type the exception in the signature given you own the helper.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
```diff
@@ -60,21 +94,31 @@ private Map<String, String> columnDefaultConfig(TableDto tableDto) {
...
    */
-  private boolean isColumnDefaultRamped(TableDto tableDto) {
+  private boolean isColumnDefaultRamped(TableDto tableDto, boolean failClosed) {
+    if (columnDefaultsSource == ColumnDefaultsSource.NONE) {
+      return false;
+    }
     try {
       return featureToggle.isFeatureActivatedWithOverride(tableDto, COLUMN_DEFAULT_FEATURE_ID);
     } catch (RuntimeException e) {
+      if (failClosed) {
```
mkuchenbecker: Same comment as above.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
```diff
@@ -0,0 +1,317 @@
...
+ * overlay. On Iceberg schema JSON that is NestedField; {@code element-id} / {@code schema-id} are
+ * different keys.
+ */
+public class ReadBridgeStripProtection {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+  private static final String ID = "id";
+  private static final String NAME = "name";
+  private static final String INITIAL_DEFAULT = "initial-default";
+  private static final String SUPPORTED_CLIENTS =
```
mkuchenbecker: these are keys right?

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
```diff
@@ -0,0 +1,317 @@
...
+    try {
+      Map<Integer, String> previousStamped =
+          existing == null ? Collections.emptyMap() : resolver.stampedColumnDefaults(existing);
+      Map<Integer, String> incomingStamped = resolver.stampedColumnDefaults(incoming);
+      if (existing != null) {
+        rejectRemovedDefaults(previousStamped, incomingStamped, incoming);
+        rejectUnawareRewrite(previousStamped, incoming);
+      }
+      return stripInitialDefaults(incoming);
+    } catch (UnsupportedClientOperationException e) {
```
mkuchenbecker: where is that thrown from? is a more-specific catch than that entire logic chain availible?

services/tables/src/test/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtectionTest.java
```diff
@@ -0,0 +1,477 @@
...
+          + "{\"id\":2,\"required\":false,\"name\":\"country\",\"type\":\"string\","
+          + "\"initial-default\":\"x\"}]}";
+  private static final String SCHEMA_WITHOUT_DEFAULT =
+      "{\"type\":\"struct\",\"fields\":["
+          + "{\"id\":1,\"required\":false,\"name\":\"id\",\"type\":\"int\"},"
+          + "{\"id\":2,\"required\":false,\"name\":\"country\",\"type\":\"string\"}]}";
+  private static final String SCHEMA_WITHOUT_COUNTRY =
+      "{\"type\":\"struct\",\"fields\":["
+          + "{\"id\":1,\"required\":false,\"name\":\"id\",\"type\":\"int\"}]}";
+  private static final String NESTED_WITH_DEFAULT =
```
mkuchenbecker: nit Can we just use a multi-line literal vs a string with escapes? The raw json can aslo be written to a file for clarity.

services/common/src/main/java/com/linkedin/openhouse/common/exception/UnsupportedClientOperationException.java
mkuchenbecker, line 10: Why is this runtime exception? That is a sneaky throw.

mkuchenbecker: i.e. this actually seems like the wrong layer. UnsupportedCLient I htink should be handled at the boundary and this case would be illegal state.

mkuchenbecker (commented):
Overall makes sense. My biggest concern is the UnsupportedClientOperation being. sneaky throw; to me this coudl be resolved at the boundary. All uses of the error are in private methods with no need to worry about the type signature in this PR.

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
cbb330: fixed to use a typed exception everywhere within this PR where feasible

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
cbb330: fixed

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
cbb330: yes but line 47 is a exception message. opted to hardcode the message and put the keys in a static object

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtection.java
cbb330: yes, fixed

services/tables/src/test/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeStripProtectionTest.java
cbb330: pulled these out into test resource files with .json extension

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
```diff
@@ -27,61 +29,149 @@ public class ReadBridgeConfigResolver {
...
+        && incoming.getTableLocation() != null
+        && !incoming.getTableLocation().isEmpty()) {
+      return incoming.getTableLocation();
+    }
+    if (existing != null
+        && existing.getTableLocation() != null
+        && !existing.getTableLocation().isEmpty()) {
+      return existing.getTableLocation();
+    }
+    return "unavailable";
```
abhisheknath2011: Why this is hardcoded with value? Through empty?

services/common/src/main/java/com/linkedin/openhouse/common/exception/UnsupportedClientOperationException.java
cbb330: good point. i converted to use a new exception: ColumnDefaultException which is inherited from Exception. the existing unsupported client from main is not ideal , so we wont' refactor it here, but instead use a new one

services/tables/src/main/java/com/linkedin/openhouse/tables/readbridge/ReadBridgeConfigResolver.java
cbb330: this just exists for help in formatting an exception message. i don't like the style though of passing around strings for building an exception message. so i refactored to use string formatting in line instead of utility methods.

mkuchenbecker (approved):

abhisheknath2011 (approved):
Thanks @cbb330. Looks good.

PR #679, opened 2026-08-12
[default values] Implement ReadBridge.apply without an Iceberg fork

cbb330:
## Summary

Depends on [#678](https://github.com/linkedin/openhouse/pull/678) (server handshake + drop), stacked on [#674](https://github.com/linkedin/openhouse/pull/674) → [#668](https://github.com/linkedin/openhouse/pull/668) → [#645](https://github.com/linkedin/openhouse/pull/645).

Finish the **client** half of the column-default read bridge: turn server-stamped `config` into Iceberg `NestedField.initialDefault` at metadata load — **without forking Iceberg**.

#678 lands first so apply is never in tree without the drop. Overlays sit on `metadata.schema()`; the first write after a bridged load would persist them unless the server already strips them.

| Step | Where | What |
|---|---|---|
| Decode | `ReadBridge.from(config)` in `doRefresh` / before IO | Parse `openhouse.read-bridge.column-default.<fieldId>` (kept from #668, off the retry path) |
| Apply | `ReadBridge.apply(raw)` in `loadMetadata` | Overlay defaults onto **every** schema-id, then rebuild metadata |

## How apply works

Patch `initial-default` onto schema field objects by field-id in metadata JSON. No Iceberg fork.

**Failure policy (mirrors the server encoder):**

| Case | Behavior |
|---|---|
| No read-bridge keys / inert bridge | Return `raw` unchanged |
| Field-id absent from a schema | Capability gap → leave that field NULL |
| Default cannot bind to column type | Fail loud (`IllegalStateException`) |
| Malformed known config entry | Fail loud (decode; already unrecoverable for retries via #668) |
| Unknown `openhouse.read-bridge.*` key | Ignore (forward compatible) |

**Semantics notes:**
- Overlay hits **all** schema-ids (time-travel / tags / branches), not only current.
- Decode keeps validated JSON **strings** (shade-safe bind at apply time).
- Current stamp wins — no default history; missing ids on older schemas are gaps, not errors.

## Changes

- [x] New Features — `ReadBridge.apply` overlays `initialDefault`
- [x] Tests — `ReadBridgeTest`: single/multi-schema overlay, nested fields, gaps, fail-loud bind, inert path; `OpenHouseTableOperationsTest`: PUT still sends stamped `initial-default`

## Testing Done

- Local: `./gradlew :integrations:java:iceberg-1.2:openhouse-java-itest:test --tests '*ReadBridgeTest' --tests '*OpenHouseTableOperationsTest'`

## Additional Information

### Stack

1. [#645](https://github.com/linkedin/openhouse/pull/645) — substrate (seam + encode + decode hook)
2. [#668](https://github.com/linkedin/openhouse/pull/668) — decode off Iceberg’s retry path
3. [#674](https://github.com/linkedin/openhouse/pull/674) — policy / ramp in OpenHouse
4. [#678](https://github.com/linkedin/openhouse/pull/678) — handshake + drop on the server
5. **This PR** — `ReadBridge.apply`
6. [#681](https://github.com/linkedin/openhouse/pull/681) — Spark catalog itest (overlay)

### DIFF
--- a/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
+++ b/integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
@@ -3,12 +3,13 @@ package com.linkedin.openhouse.javaclient;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.StreamSupport;
 import org.apache.iceberg.TableMetadata;
 import org.apache.iceberg.TableMetadataParser;
 
@@ -37,62 +38,60 @@ final class ReadBridge {
   private static final String ID = "id";
   private static final String INITIAL_DEFAULT = "initial-default";
 
-  /** JSON strings, not JsonNodes — Jackson is relocated in the shaded client. */
-  private final Map<Integer, String> columnDefaults;
+  private final Map<Integer, JsonNode> columnDefaults;
 
-  private ReadBridge(Map<Integer, String> columnDefaults) {
+  private ReadBridge(Map<Integer, JsonNode> columnDefaults) {
     this.columnDefaults = columnDefaults;
   }
 
   /**
    * Decode stamped config. Returns {@link #INERT} when there is nothing to apply.
    *
-   * @throws IllegalStateException if a key this client owns is malformed
+   * @throws ReadBridgeException if a key this client owns is malformed
    */
-  static ReadBridge from(Map<String, String> config) {
-    Map<Integer, String> columnDefaults = decodeColumnDefaults(config);
+  static ReadBridge from(Map<String, String> config) throws ReadBridgeException {
+    Map<Integer, JsonNode> columnDefaults = decodeColumnDefaults(config);
     return columnDefaults.isEmpty() ? INERT : new ReadBridge(columnDefaults);
   }
 
   /** Overlay onto {@code raw}, or return it unchanged. */
-  TableMetadata apply(TableMetadata raw) {
+  TableMetadata apply(TableMetadata raw) throws ReadBridgeException {
     if (columnDefaults.isEmpty()) {
       return raw;
     }
     ObjectNode root = metadataJson(raw);
     boolean changed = false;
     for (JsonNode field : fieldObjects(root)) {
-      String defaultJson = columnDefaults.get(field.get(ID).asInt());
-      if (defaultJson != null) {
-        ((ObjectNode) field).set(INITIAL_DEFAULT, readTree(defaultJson));
+      JsonNode defaultValue = columnDefaults.get(field.get(ID).asInt());
+      if (defaultValue != null) {
+        ((ObjectNode) field).set(INITIAL_DEFAULT, defaultValue);
         changed = true;
       }
     }
     return changed ? fromMetadataJson(raw, root) : raw;
   }
 
-  Map<Integer, String> columnDefaults() {
+  Map<Integer, JsonNode> columnDefaults() {
     return columnDefaults;
   }
 
   /** Decode {@code column-default.<fieldId>} entries. Unknown keys are ignored. */
-  private static Map<Integer, String> decodeColumnDefaults(Map<String, String> config) {
+  private static Map<Integer, JsonNode> decodeColumnDefaults(Map<String, String> config)
+      throws ReadBridgeException {
     if (config == null) {
       return Collections.emptyMap();
     }
-    Map<Integer, String> byFieldId = new HashMap<>();
+    Map<Integer, JsonNode> byFieldId = new HashMap<>();
     for (Map.Entry<String, String> entry : config.entrySet()) {
       if (!entry.getKey().startsWith(COLUMN_DEFAULT_PREFIX)) {
         continue;
       }
       try {
         int fieldId = Integer.parseInt(entry.getKey().substring(COLUMN_DEFAULT_PREFIX.length()));
-        // Validate JSON; keep the original string so apply can bind without a relocated JsonNode.
-        MAPPER.readTree(entry.getValue());
-        byFieldId.put(fieldId, entry.getValue());
-      } catch (RuntimeException | JsonProcessingException e) {
+        byFieldId.put(fieldId, MAPPER.readTree(entry.getValue()));
+      } catch (NumberFormatException | JsonProcessingException e) {
         // Known keys are stamped as int field-id + JSON; anything else is a bug.
-        throw new IllegalStateException(
+        throw new ReadBridgeException(
             "read-bridge: unusable "
                 + COLUMN_DEFAULT_PREFIX
                 + " entry "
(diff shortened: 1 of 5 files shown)


integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -84,4 +104,48 @@ private static Map<Integer, JsonNode> columnDefaults(Map<String, String> config)
...
+    } catch (RuntimeException | JsonProcessingException e) {
+      throw new IllegalStateException("read-bridge: failed to rebuild table metadata schemas", e);
+    }
+  }
+
+  private static JsonNode readTree(String json) {
+    try {
+      return MAPPER.readTree(json);
+    } catch (JsonProcessingException e) {
+      throw new IllegalStateException("read-bridge: invalid json", e);
```
mkuchenbecker: any reason we are using sneaky throws vs typed?

cbb330: Confirmed^ 

Jackson's `JsonProcessingException` is checked, then we wrap it as `IllegalStateException` so from/apply stay unchecked.

then, `loadMetadata` catches that one type for both decode and apply and rethrows  as `Tasks.UnrecoverableException`. That's what keeps Iceberg from retrying the metadata read.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -15,20 +19,28 @@
...
-  private final Map<Integer, JsonNode> columnDefaults;
+  private static final String SCHEMAS = "schemas";
+  private static final String ID = "id";
+  private static final String INITIAL_DEFAULT = "initial-default";
+
+  /** JSON strings, not JsonNodes — Jackson is relocated in the shaded client. */
+  private final Map<Integer, String> columnDefaults;
 
-  private ReadBridge(Map<Integer, JsonNode> columnDefaults) {
+  private ReadBridge(Map<Integer, String> columnDefaults) {
```
mkuchenbecker: seems like a type regression to use string vs jsonnode

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -81,4 +104,48 @@ private static Map<Integer, JsonNode> columnDefaults(Map<String, String> config)
...
+          "read-bridge: metadata JSON missing required '" + SCHEMAS + "' array");
+    }
+    List<JsonNode> fields = new ArrayList<>();
+    for (JsonNode schema : schemas) {
+      List<JsonNode> found = schema.findParents(ID);
+      if (found != null) {
+        fields.addAll(found);
+      }
+    }
+    return fields;
```
mkuchenbecker: Are we not able to construct a new json node here?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -81,4 +104,48 @@ private static Map<Integer, JsonNode> columnDefaults(Map<String, String> config)
...
+
+  /** Schema field objects are the JSON nodes that carry {@code id}. */
+  private static List<JsonNode> fieldObjects(ObjectNode metadata) {
+    JsonNode schemas = metadata.get(SCHEMAS);
+    if (schemas == null || !schemas.isArray()) {
+      throw new IllegalStateException(
+          "read-bridge: metadata JSON missing required '" + SCHEMAS + "' array");
+    }
+    List<JsonNode> fields = new ArrayList<>();
+    for (JsonNode schema : schemas) {
```
mkuchenbecker: .Map vs for  loop :D

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
mkuchenbecker: Can we type it anyway? I like exceptions in the signature (or modeled as a sum result type)

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
abhisheknath2011: Is there a possibility of using JsonNode so that we can deal with json apis directly?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -47,26 +59,37 @@ TableMetadata apply(TableMetadata raw) {
...
     for (Map.Entry<String, String> entry : config.entrySet()) {
       if (!entry.getKey().startsWith(COLUMN_DEFAULT_PREFIX)) {
         continue;
       }
       try {
         int fieldId = Integer.parseInt(entry.getKey().substring(COLUMN_DEFAULT_PREFIX.length()));
-        byFieldId.put(fieldId, MAPPER.readTree(entry.getValue()));
+        // Validate JSON; keep the original string so apply can bind without a relocated JsonNode.
+        MAPPER.readTree(entry.getValue());
+        byFieldId.put(fieldId, entry.getValue());
```
abhisheknath2011: Are applying the default values as string? How do we ensure that if the String value is converted to appropriate field type (Integer, string etc) while applying the default values ?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: fixed, using jsonnode instead.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: fixed , now we throw a typed ReadBridgeException and catch it at the higher layer to ensure it throws as iceberg's UnrecoverableException in order to prevent retries.

cbb330: try {
  return bridge.apply(raw);
} catch (ReadBridgeException e) {
  throw unrecoverableBridge(e);
}

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: yes we can, I added it. the downside is its a bit mechanical. 

Now, it walks schema field objects and patches those. No extra ArrayNode.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: fixed

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: fixed, using jsonnode now

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: Config must be a string so that it can return on the wire to the client. 

then on the client, TableMetadataParser.fromJson binds it to the column type; unbindable throws CANNOT_BIND.

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
```diff
@@ -35,27 +55,51 @@ private ReadBridge(Map<Integer, JsonNode> columnDefaults) {
...
+    try {
+      json = MAPPER.writeValueAsString(root);
+    } catch (JsonProcessingException e) {
+      throw ReadBridgeException.unusableMetadata(
+          "read-bridge: failed to serialize patched metadata", e);
+    }
+    try {
+      return TableMetadataParser.fromJson(raw.metadataFileLocation(), json);
+    } catch (RuntimeException e) {
+      throw ReadBridgeException.cannotBind(e);
```
mkuchenbecker, line 92: RedBridgeException is not a checked exception?

Why does TAbleMetadataParser return a runtime at-all? Is there a more specific exception to catch?

mkuchenbecker (approved):

integrations/java/iceberg-1.2/openhouse-java-runtime/src/main/java/com/linkedin/openhouse/javaclient/ReadBridge.java
cbb330: ReadBridgeException does extend Exception; apply still catches RuntimeException from TableMetadataParser.fromJson / toJson and wraps it as CANNOT_BIND / UNUSABLE_METADATA

abhisheknath2011 (approved):

PR #690, opened 2026-08-21
Iceberg REST Phase 1 [2/3]: Runtime facade and unit coverage

cbb330:
## Summary

Part 2 of 3 for a read-only Iceberg REST Catalog facade.

This PR implements the generated contract as a thin Spring MVC adapter. Protocol translation lives in `IcebergRestApiHandler`; existing OpenHouse table behavior is reused, not rewritten. The facade is disabled by default.

## Changes

- [x] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [x] Refactoring
- [x] Documentation
- [x] Tests

### Client-facing API Changes

When `cluster.tables.iceberg-rest.enabled=true`:

- `GET /v1/config` returns prefix `iceberg` and advertises only the implemented endpoints.
- `GET /v1/iceberg/namespaces/{namespace}/tables` lists tables with opaque pagination.
- `GET /v1/iceberg/namespaces/{namespace}/tables/{table}` loads Iceberg table metadata.
- `HEAD /v1/iceberg/namespaces/{namespace}/tables/{table}` checks existence.

Existing OpenHouse `/v1/databases/...` APIs are unchanged. Unsupported Iceberg REST operations are not generated or advertised.

### Internal API Changes

- `IcebergRestCatalogController` implements the generated interfaces and delegates to `IcebergRestApiHandler`.
- `OpenHouseIcebergRestApiHandler` reuses `TablesApiHandler` for authorization/validation and `OpenHouseInternalCatalog` for Iceberg `LoadTableResponse`.
- `TablesService` is unchanged. The earlier principal-aware list overload is removed; listing now uses the existing paginated handler path.
- Iceberg JSON is serialized by a write-only converter that does not affect OpenHouse JSON.
- Controller-scoped advice returns Iceberg `ErrorResponse` envelopes and sanitizes 403/500 messages.

### New Features

- Feature flag `cluster.tables.iceberg-rest.enabled` (default `false`), also on `ClusterProperties`.
- Local docker-compose enables the flag for smoke testing.
- Capabilities advertisement, prefix validation, single-level namespace decoding, opaque page tokens, and explicit rejection of `snapshots=refs`.

### Refactoring

- Moves protocol logic out of the controller so the generated interface remains a transport adapter.

### Documentation

- Adds `docs/iceberg-rest-catalog.md` covering enablement, architecture, limitations, observability, and spec maintenance.

### Tests

- Controller tests cover typed responses, sanitized 403s, and delegation.
- Handler tests cover capabilities, pagination tokens, prefix/page-token validation, snapshot projection rejection, and reuse of the existing read handler.
- Feature-flag tests confirm the controller is absent by default and present when enabled.

## Testing Done

Validated locally with JDK 17:

```bash
./gradlew :services:tables:check
```

That includes the new controller, handler, and feature-flag tests. Client round-trips and the PyIceberg CI smoke test are in #691.

This PR is stacked on #689, so the GitHub files tab shows only the runtime/facade delta.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

Merge in order:

1. #689 — Read-only contract and codegen
2. **#690 — Runtime facade and unit coverage** ← you are here
3. #691 — Client and integration validation


cbb330:
Follow-up for the 1.11 contract bump in #689: `loadTable` now accepts the Iceberg 1.11 `referenced-by` query parameter and ignores it in Phase 1. Docs updated accordingly.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseIcebergRestApiHandler.java
```diff
@@ -0,0 +1,172 @@
...
+
+  private static Namespace decodeSingleLevelNamespace(String encodedNamespace) {
+    Namespace namespace = RESTUtil.decodeNamespace(encodedNamespace);
+    if (namespace.isEmpty() || namespace.levels().length != 1) {
+      throw new NoSuchNamespaceException("Only single-level namespaces are supported");
+    }
+    return namespace;
+  }
+
+  private static PageCursor decodePageToken(String pageToken, Integer requestedPageSize) {
```
mkuchenbecker, line 129: **Fix before merge — empty `pageToken` is rejected as malformed, breaking `listTables` for every current Java client.** The spec's PageToken schema says clients "may initiate the first paginated request by sending an empty query parameter `pageToken`", and every Iceberg Java client from 1.6.0 through 1.11.0 does exactly that: `RESTSessionCatalog.listTables` sends `pageToken=` unconditionally on its first request. Spring binds that to the empty string, this method takes the non-null branch, base64-decodes "" to an empty array, splits to one part instead of three, and throws, which the error mapper renders as HTTP 400. Result: `catalog.listTables()` — and with it Spark's `SHOW TABLES` and Trino's metadata listing — fails outright on every current client, while `loadTable` still works, so the failure looks arbitrary to operators. Fix: treat a blank token as the first page — `if (pageToken == null || pageToken.isEmpty())`.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseIcebergRestApiHandler.java
```diff
@@ -0,0 +1,172 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.data.domain.Page;
+import org.springframework.stereotype.Component;
+
+/** Default Iceberg REST adapter backed by existing OpenHouse API handlers and catalog behavior. */
+@Component
+@ConditionalOnProperty(value = "cluster.tables.iceberg-rest.enabled", havingValue = "true")
+public class OpenHouseIcebergRestApiHandler implements IcebergRestApiHandler {
+
+  static final int DEFAULT_PAGE_SIZE = 100;
```
mkuchenbecker, line 36: **Fix before merge — the server paginates when the client never asked, silently truncating at 100.** The spec's PageToken schema: "Servers that support pagination must return all results in a single response with the value of `next-page-token` set to `null` if the query parameter `pageToken` is not set in the request." A request with no `pageToken` gets `DEFAULT_PAGE_SIZE = 100` here instead. The clients that never send a token are real and current: PyIceberg (all released versions) and Iceberg Java ≤1.5.x issue one unpaginated GET and read what comes back — a 350-table namespace returns 100 tables with no error, and downstream jobs quietly process a third of the catalog. This is the most dangerous finding in the stack precisely because nothing fails. Fix: when no `pageToken` is present, return the full listing with `next-page-token: null`; paginate only when the client opts in via `pageToken` or `pageSize`.

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/IcebergRestExceptionHandler.java
```diff
@@ -0,0 +1,64 @@
...
+  public ResponseEntity<ErrorResponse> handleNoSuchTable(NoSuchTableException e) {
+    return errorResponse(404, e.getMessage(), NoSuchTableException.class.getSimpleName());
+  }
+
+  @ExceptionHandler(NoSuchNamespaceException.class)
+  public ResponseEntity<ErrorResponse> handleNoSuchNamespace(NoSuchNamespaceException e) {
+    return errorResponse(404, e.getMessage(), NoSuchNamespaceException.class.getSimpleName());
+  }
+
+  @ExceptionHandler({RequestValidationFailureException.class, IllegalArgumentException.class})
```
mkuchenbecker, line 34: **Fix before merge — any interior `IllegalArgumentException` becomes "HTTP 400: your fault", including server misconfiguration.** This maps every `IllegalArgumentException` — a generic type thrown throughout OpenHouse's interior, not just by the facade's input checks — to 400 with the exception message as the body. A reachable path: `loadTable` reaches `FileIOManager.getFileIO` (through `OpenHouseInternalCatalog.newTableOps`), which throws `IllegalArgumentException("S3 is not configured")` when the cluster lacks that storage bean — the client is blamed for a server fault, never retries (Iceberg clients don't retry 400) even though the fault clears during a rolling deploy, and the cluster's storage configuration leaks to callers. The facade's own boundary checks also signal through this sneaky channel (validatePrefix L117, decodePageToken L138/143/147, validatePageSize L154, plus `UnsupportedOperationException` at L86 and the 501 mapping at L44): bare runtime types get protocol meaning only from a distant catch that also catches everyone else's. Fix: one facade-owned bad-request exception thrown at those validation sites, with 400 keyed on it alone (or on `RequestValidationFailureException`); unrecognized interior exceptions fall to 500. Same change for the 501 mapping.

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/IcebergRestExceptionHandler.java
```diff
@@ -0,0 +1,64 @@
...
+  public ResponseEntity<ErrorResponse> handleForbidden(AccessDeniedException e) {
+    return errorResponse(403, "Access denied", ForbiddenException.class.getSimpleName());
+  }
+
+  @ExceptionHandler(UnsupportedOperationException.class)
+  public ResponseEntity<ErrorResponse> handleNotImplemented(UnsupportedOperationException e) {
+    return errorResponse(501, e.getMessage(), UnsupportedOperationException.class.getSimpleName());
+  }
+
+  @ExceptionHandler(Exception.class)
```
mkuchenbecker, line 49: **Fix before merge — a malformed query parameter returns 500 where the spec declares 400.** `GET .../tables?pageSize=abc` makes Spring throw `MethodArgumentTypeMismatchException` before any handler code runs; with no specific mapping it lands in this catch-all, logs at ERROR with a stack trace, and returns 500. Clients treat 500 as transient and retry a request they should fix; operators chasing the 500s find no server fault. Fix (~5 lines): map `MethodArgumentTypeMismatchException` and `MissingServletRequestParameterException` to 400.

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/IcebergRestExceptionHandler.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.bind.annotation.RestControllerAdvice;
+
+/** Scoped exception mapper for Iceberg REST endpoints. */
+@Order(Ordered.HIGHEST_PRECEDENCE)
+@RestControllerAdvice(assignableTypes = IcebergRestCatalogController.class)
+@ConditionalOnProperty(value = "cluster.tables.iceberg-rest.enabled", havingValue = "true")
+@Slf4j
+public class IcebergRestExceptionHandler {
```
mkuchenbecker, line 22: **Fix before enabling — this mapper covers 5 of the ~14 failure outcomes the tables core produces.** OpenHouse's native `OpenHouseExceptionHandler` renders the core's failures across eight statuses (404, 409, 422, and 503 among them); this advice renders five and collapses the rest to 500 via the `Exception` catch-all. Concretely: when OPA — the authorization service consulted on every table read — is unreachable, the core throws `AuthorizationServiceException`; the native API answers 503 (retryable), this surface answers 500. Every core outcome added later silently joins the 500 bucket. Fix: add handlers for the classes the native mapper already enumerates, plus a parity test asserting the two mappers cover the same set, so a new core outcome fails the build instead of degrading silently.

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/IcebergRestExceptionHandler.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
+import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.bind.annotation.RestControllerAdvice;
+
+/** Scoped exception mapper for Iceberg REST endpoints. */
+@Order(Ordered.HIGHEST_PRECEDENCE)
+@RestControllerAdvice(assignableTypes = IcebergRestCatalogController.class)
```
mkuchenbecker, line 19: **Fix before enabling — error translation is scoped to the controller class, so unrouted Iceberg paths answer in OpenHouse's error format with a stack trace.** `assignableTypes` applies only once a request reaches this controller. A request to an Iceberg-spec path the facade doesn't implement — e.g. `GET /v1/iceberg/namespaces/{ns}` (namespace load) — matches no route and is rendered by the global handler in OpenHouse's JSON shape, abbreviated Java stack trace included. This is not exotic: Iceberg 1.5.x clients don't read the `/v1/config` `endpoints` list (that gating shipped in 1.7), so `USE db` on a 1.5.x client hits exactly this in normal operation. Fix: a path-scoped error renderer for the Iceberg URL prefix, so everything under the Iceberg surface speaks Iceberg's error format.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseIcebergRestApiHandler.java
```diff
@@ -0,0 +1,172 @@
...
+    if (snapshots != null && !"all".equals(snapshots)) {
+      throw new UnsupportedOperationException(
+          "The snapshots=refs projection is not supported by this catalog");
+    }
+    // Iceberg 1.11 loadTable may send referenced-by for view-load chains; Phase 1 ignores it.
+
+    Namespace icebergNamespace = decodeSingleLevelNamespace(namespace);
+    String databaseId = icebergNamespace.level(0);
+    try {
+      tablesApiHandler.getTable(databaseId, table, extractAuthenticatedUserPrincipal());
```
mkuchenbecker, line 94: **Fix before enabling — the facade calls the native API's protocol adapter instead of `TablesService`, so native-API policy silently becomes Iceberg contract.** Concrete leak today: `TablesApiHandler.getTable` validates identifiers against `^[a-zA-Z0-9_]+$` and throws a validation failure, which this surface renders as 400. Iceberg places no restriction on names, and its clients probe existence by treating 404 as "no" — so Spark's ``DROP TABLE IF EXISTS db.`my-table` `` gets 400 instead of 404 and the `IF EXISTS` guard fails at its one job. Also, the authorization check and the returned metadata come from two separate calls on two different paths, with nothing keeping them consistent when the core's read policy gains a rule. Fix: depend on `TablesService.getTable` (which already applies authorization) and build the response from its result; note the audit aspect currently points at `TablesApiHandler`, so move the audit hook to the service layer in the same change or Iceberg reads vanish from the audit log. Interim two-line version: here and in `tableExists`, catch `RequestValidationFailureException` and rethrow `NoSuchTableException` — a name OpenHouse cannot store is a table that does not exist.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseIcebergRestApiHandler.java
```diff
@@ -0,0 +1,172 @@
...
+
+  @Override
+  public CatalogConfig getConfig(String warehouse) {
+    return new CatalogConfig(
+            Collections.singletonMap("prefix", ICEBERG_REST_PREFIX), Collections.emptyMap())
+        .endpoints(IcebergRestOpenHouseSupport.SUPPORTED_ENDPOINTS);
+  }
+
+  @Override
+  public ListTablesResponse listTables(
```
mkuchenbecker, line 57: **Fix before enabling — listing a namespace that doesn't exist returns 200 + empty; the spec declares 404 (`NoSuchNamespaceException`).** No existence check runs: the backing `OpenHouseInternalCatalog.listTables` runs `findAllByDatabaseId`, which returns an empty page for a never-created database rather than throwing. A typo'd `SHOW TABLES IN prod_bd` reads as "empty database", and existence probes built on listing always answer yes. The suite masks this: `listTablesEmptyNamespace` exercises an emptied-but-existing namespace, which passes identically. The fix is a decision plus a test on either branch: conform (resolve the database via the existing databases API, throw `NoSuchNamespaceException`, which already maps to 404, and add the contract test) or document the deviation and pin the `200 []` with a test. Checking via the database API is preferred over an "empty first page means 404" rule, which would answer 404 for a real-but-empty namespace.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseIcebergRestApiHandler.java
```diff
@@ -0,0 +1,172 @@
...
+  public LoadTableResponse loadTable(
+      String prefix,
+      String namespace,
+      String table,
+      String accessDelegation,
+      String ifNoneMatch,
+      String snapshots,
+      String referencedBy) {
+    validatePrefix(prefix);
+    if (snapshots != null && !"all".equals(snapshots)) {
```
mkuchenbecker, line 85: **Fix before enabling — `snapshots=refs` fails every table load for clients configured the documented way.** `refs` is a narrowing projection: a server returning all snapshots has returned a valid superset, so no implementation is needed — yet this guard 501s it, and any operator who sets `snapshot-loading-mode=refs` (the documented way to make loads cheap on long-history tables) fails on every load, with no fallback. 501 is also not among the responses the spec declares for loadTable. And the same `!"all".equals(...)` branch reports a typo (`snapshots=xyz`) as "not implemented", where the spec's enum `[all, refs]` makes it a 400. Fix: accept `refs` by ignoring it; return 400 for values outside the enum; add the controller-level test for the malformed value (that partition is currently untested).

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/IcebergRestSerdeConfig.java
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+
+/**
+ * Registers the {@link IcebergRestHttpMessageConverter} so that Spring MVC can serialize typed
+ * Iceberg REST responses returned by {@link IcebergRestCatalogController}.
+ */
+@Configuration
+@ConditionalOnProperty(value = "cluster.tables.iceberg-rest.enabled", havingValue = "true")
```
mkuchenbecker, line 14: **Fix before enabling — the feature switch is four independent copies of this property, and partial activation emits unparseable 200s.** The controller, the error mapper, this config, and a fourth bean each carry their own `@ConditionalOnProperty` for `cluster.tables.iceberg-rest.enabled`, and the `ClusterProperties` field added for the same switch is read by nothing. If a future edit leaves the controller on and this config off, Spring's default Jackson mapper serializes the Iceberg responses in camelCase — every response becomes HTTP 200 with a body no Iceberg client can parse, and nothing fails server-side. Separately, `converters.add(0, ...)` registers the converter application-wide at highest priority, where only its `supports()` check keeps it off native-API responses — narrow it to the concrete Iceberg response types. Fix: one `@Configuration` class gated once, declaring or importing all Iceberg REST beans; delete or wire the dead field.

services/tables/src/main/java/com/linkedin/openhouse/tables/controller/IcebergRestSerde.java
```diff
@@ -0,0 +1,33 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import org.apache.iceberg.rest.RESTSerializers;
+
+/** Serde helper for Iceberg REST payloads that require kebab-case and Iceberg serializers. */
+final class IcebergRestSerde {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper(new JsonFactory());
```
mkuchenbecker, line 15: **Fix before enabling — the biggest payload's wire format is an accident of reflection.** `LoadTableResult` is type-mapped out of codegen onto Iceberg 1.5.2's `LoadTableResponse`, and 1.5.2 ships no serializer for that class — so this reflection-over-private-fields mapper with a global kebab-case strategy produces the wire format: `metadata-location` is spelled right only because the Java field happens to be named `metadataLocation`. Two consequences: a spec change to this payload is invisible to the build (the model was mapped away), and bumping the server's Iceberg dependency to ≥1.7 silently switches serialization to Iceberg's real parser, changing the wire format as a side effect of a dependency bump. Fix: a round-trip test that serializes through this mapper and parses back with Iceberg's own `LoadTableResponseParser`, pinning the format to Iceberg's definition.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseIcebergRestApiHandler.java
```diff
@@ -0,0 +1,172 @@
...
+    return new ListTablesResponse().identifiers(identifiers).nextPageToken(nextPageToken);
+  }
+
+  @Override
+  public LoadTableResponse loadTable(
+      String prefix,
+      String namespace,
+      String table,
+      String accessDelegation,
+      String ifNoneMatch,
```
mkuchenbecker, line 81: **Fix before enabling — the documented "no ETag/304" deviation has no pin.** The spec defines `304 Not Modified` keyed on `If-None-Match`; the parameter is accepted here and never read, and docs/iceberg-rest-catalog.md records the deviation ("Conditional ETag responses are not currently emitted"). Deliberate and documented — but no test locks it, so a change that starts emitting 304s (or a caching layer injecting them) alters client-visible behavior with nothing tripping. Fix: one pin test — a load-table request carrying `If-None-Match` still returns 200 — labeled as a pin so its failure reads "behavior changed", not "bug".

docs/iceberg-rest-catalog.md
```diff
@@ -0,0 +1,65 @@
...
+
+The OpenAPI-generated interfaces own the HTTP contract. `IcebergRestCatalogController` is a thin
+Spring MVC adapter, and `IcebergRestApiHandler` translates the Iceberg protocol to existing
+`TablesApiHandler` and `OpenHouseInternalCatalog` behavior. The facade does not add business rules
+or change the existing OpenHouse endpoints.
+
+Iceberg response types use a narrowly scoped Spring `HttpMessageConverter`. Errors are translated
+by controller-scoped advice into the standard Iceberg error envelope.
+
+## Compatibility and limitations
```
mkuchenbecker, line 30: **Follow-up — namespace endpoints are absent, so `SHOW SCHEMAS` is silently empty and `USE db` throws.** Modern Java clients return an empty list from `listNamespaces` without error when the endpoint isn't advertised, and `USE db` throws `UnsupportedOperationException` — while reading the same database's tables works, a split operators will report as a bug. OpenHouse databases map one-to-one onto single-level Iceberg namespaces and `GET /v1/databases` already exists, so list/load-namespace is mechanical — but it is scope expansion, hence a follow-up. Cheap now: add a line here stating that namespace listing returns empty rather than erroring, so the behavior is not mistaken for an empty catalog.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseIcebergRestApiHandler.java
```diff
@@ -0,0 +1,172 @@
...
+        openHouseInternalCatalog, TableIdentifier.of(icebergNamespace, table));
+  }
+
+  @Override
+  public void tableExists(String prefix, String namespace, String table) {
+    validatePrefix(prefix);
+    Namespace icebergNamespace = decodeSingleLevelNamespace(namespace);
+    String databaseId = icebergNamespace.level(0);
+    try {
+      tablesApiHandler.getTable(databaseId, table, extractAuthenticatedUserPrincipal());
```
mkuchenbecker, line 109: **Follow-up — a missing login silently becomes the principal `"undefined"`.** The handler reads the security context ambiently, and the shared helper converts an absent authentication into the literal string `"undefined"`, which then flows into authorization as if it were a user. Any future invocation off a request thread (async wrapper, warm-up task) authorizes as `"undefined"` instead of failing, and nothing in the method signatures reveals the dependency. Fix: pass the acting principal as a parameter from the controller, matching the native controllers' convention.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/IcebergRestApiHandler.java
```diff
@@ -0,0 +1,27 @@
...
+  CatalogConfig getConfig(String warehouse);
+
+  ListTablesResponse listTables(
+      String prefix, String namespace, String pageToken, Integer pageSize);
+
+  LoadTableResponse loadTable(
+      String prefix,
+      String namespace,
+      String table,
+      String accessDelegation,
```
mkuchenbecker, line 21: **Follow-up — unused wire strings cross the seam, and validation has no single owner.** `accessDelegation`, `ifNoneMatch`, and `referencedBy` are accepted by this interface and never used by the implementation; each method re-validates the prefix; and the page-size range check sits inside `decodePageToken`'s catch, so a valid token carrying an out-of-range page size reports "Invalid Iceberg REST page token" — the token was fine. Fix: parse once at the controller edge into a validated value object, drop the unused parameters, and move the range check out of the catch so its message survives.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseIcebergRestApiHandler.java
```diff
@@ -0,0 +1,172 @@
...
+  private final OpenHouseInternalCatalog openHouseInternalCatalog;
+
+  public OpenHouseIcebergRestApiHandler(
+      TablesApiHandler tablesApiHandler, OpenHouseInternalCatalog openHouseInternalCatalog) {
+    this.tablesApiHandler = tablesApiHandler;
+    this.openHouseInternalCatalog = openHouseInternalCatalog;
+  }
+
+  @Override
+  public CatalogConfig getConfig(String warehouse) {
```
mkuchenbecker, line 50: **Follow-up — the `warehouse` argument is discarded.** A client pointed at the wrong OpenHouse cluster with an explicit `warehouse=` connects successfully and reads the wrong catalog. Spec-legal (the field is a hint), but cheap to reject when the name doesn't match this cluster's identity, turning a silent wrong-catalog read into an immediate 400.

mkuchenbecker (commented):
Part of a coordinated review of the #689/#690/#691 stack. It was generated based on the review skills at https://github.com/mkuchenbecker/code-review-skills — linked so the criteria being judged are inspectable: arch-review (module topology, contracts, and the error channel), testing-review (testing as falsification), plus an Iceberg REST spec conformance pass.

Verdict for this PR: much of the facade is right — the error envelope shape and the 404 `type` strings are exactly what Iceberg's client-side `ErrorHandlers` switch on; the `/v1/config` `endpoints` strings byte-match the Java client's endpoint gating; putting `prefix` in `overrides` is the robust choice. But the facade does not work against any stock Iceberg Java client newer than 1.5.2, and two behaviors produce silently wrong answers rather than errors. Four fix-before-merge items are inline: an empty `pageToken` answered with 400 on every 1.6+ client; forced pagination silently truncating unpaginated clients at 100 tables; interior `IllegalArgumentException` rendered to clients as "400: your fault" (including server misconfiguration); and malformed query parameters answered with 500. Eight fix-before-enabling items and six follow-ups are also inline. "Fix before enabling" means: nothing blocks merge — the feature flag is off and nothing calls these endpoints — but each item will bite the first real deployment and is cheapest to fix while there are no callers.

18 inline comments: 4 fix-before-merge, 8 fix-before-enabling, 6 follow-ups.

---
_Generated by [Claude Code](https://claude.ai/code)_

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseIcebergRestApiHandler.java
```diff
@@ -0,0 +1,172 @@
...
+  private static int validatePageSize(Integer requestedPageSize) {
+    int pageSize = requestedPageSize == null ? DEFAULT_PAGE_SIZE : requestedPageSize;
+    if (pageSize < 1 || pageSize > MAX_PAGE_SIZE) {
+      throw new IllegalArgumentException(
+          String.format("page-size must be between 1 and %s", MAX_PAGE_SIZE));
+    }
+    return pageSize;
+  }
+
+  private static String encodePageToken(int page, int pageSize) {
```
mkuchenbecker, line 160: **Follow-up — offset-encoded page tokens skip or duplicate rows under concurrent writes.** The token encodes `v1:<page-index>:<page-size>` over a `tableId`-sorted query — an offset. A table created or dropped mid-iteration shifts every later row across page boundaries, so a crawler misses or repeats identifiers. Since the sort key is already `tableId`, a keyset token (encode the last-seen `tableId`, filter `tableId > :last`) is stable and no harder — and the token is opaque, so no client notices the change.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/handler/impl/OpenHouseIcebergRestApiHandler.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }
+
+  private static void validatePrefix(String prefix) {
+    if (!ICEBERG_REST_PREFIX.equals(prefix)) {
+      throw new IllegalArgumentException("Unsupported Iceberg REST prefix");
+    }
+  }
+
+  private static Namespace decodeSingleLevelNamespace(String encodedNamespace) {
```
mkuchenbecker, line 121: **Follow-up — the single-level-namespace limitation is enforced here but untested.** This rejection (empty or multi-level namespace answered with 404) implements the documented "only single-level namespaces" limitation, and no test on any endpoint exercises it. One controller-level test with a two-level namespace (`%1F`-separated) closes it.

PR #691, opened 2026-08-21
Iceberg REST Phase 1 [3/3]: Client and integration validation

cbb330:
## Summary

Part 3 of 3 for a read-only Iceberg REST Catalog facade.

This PR validates the facade with stock Iceberg clients. It does not change runtime behavior. Java `RESTCatalog` round-trips, a PyIceberg smoke test, and CI wiring live here; unit coverage for the facade is in #690.

## Changes

- [ ] Internal API Changes
- [ ] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

### Tests

`IcebergRestCatalogRoundTripTest` creates tables through the existing OpenHouse API and reads them through Iceberg's `RESTCatalog`:

- Config advertises only the three implemented table operations.
- List tables: namespace isolation, empty namespaces, and opaque `pageToken` / `pageSize` pagination.
- Load table: schema, field IDs, required/optional fields, partition spec, sort order, location, and properties.
- HEAD existence for present, missing, and wrong-namespace tables.
- Missing tables and namespaces raise Iceberg `NoSuchTableException`.
- Invalid prefixes return 400; `snapshots=refs` returns 501.
- Existing OpenHouse `GET /v1/databases/{db}/tables/{table}` still works.
- Table loads retain existing read audit attribution and increment `http.server.requests`.

CI:

- `integrations/python/dataloader/scripts/iceberg_rest_catalog_smoke.py` creates a table via OpenHouse and reads it with PyIceberg `RestCatalog`.
- `.github/workflows/build-run-tests.yml` runs `:services:tables:check` and that smoke test against docker-compose.

This PR is stacked on #690, so the GitHub files tab shows only the integration-test and CI delta.

## Testing Done

Validated locally with JDK 17:

```bash
./gradlew :services:tables:check
python3 -m py_compile integrations/python/dataloader/scripts/iceberg_rest_catalog_smoke.py
```

The Java round-trip suite passes, including pagination, capabilities, existing-API regression, audit, and metrics checks. The Docker/PyIceberg smoke test is exercised by CI, not re-run locally for this update.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

Merge in order:

1. #689 — Read-only contract and codegen
2. #690 — Runtime facade and unit coverage
3. **#691 — Client and integration validation** ← you are here

### DIFF
--- a/.github/workflows/build-run-tests.yml
+++ b/.github/workflows/build-run-tests.yml
@@ -21,7 +21,7 @@ jobs:
         uses: gradle/actions/setup-gradle@v5
 
       - name: Build with Gradle
-        run: ./gradlew clean build
+        run: ./gradlew clean build :services:tables:check
 
       - name: Start Docker Containers
         run: docker compose -f infra/recipes/docker-compose/oh-only/docker-compose.yml up -d --build
@@ -33,6 +33,15 @@ jobs:
         uses: actions/setup-python@v6
         with:
           python-version: '3.12'
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+
+      - name: Run Iceberg REST Catalog Smoke Test (PyIceberg)
+        working-directory: integrations/python/dataloader
+        run: uv run python scripts/iceberg_rest_catalog_smoke.py ../../../tables-test-fixtures/tables-test-fixtures-iceberg-1.2/src/main/resources/dummy.token
 
       - name: Install dependencies
         run: pip install -r scripts/python/requirements.txt
--- a/services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/IcebergRestCatalogRoundTripTest.java
+++ b/services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/IcebergRestCatalogRoundTripTest.java
@@ -1,0 +1,494 @@
+package com.linkedin.openhouse.tables.e2e.h2;
+
+import static com.linkedin.openhouse.common.api.validator.ValidatorConstants.INITIAL_TABLE_VERSION;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import com.google.gson.Gson;
+import com.jayway.jsonpath.JsonPath;
+import com.linkedin.openhouse.common.audit.AuditHandler;
+import com.linkedin.openhouse.common.security.DummyTokenInterceptor;
+import com.linkedin.openhouse.common.test.cluster.PropertyOverrideContextInitializer;
+import com.linkedin.openhouse.tables.api.spec.v0.request.CreateUpdateTableRequestBody;
+import com.linkedin.openhouse.tables.api.spec.v0.request.components.TimePartitionSpec;
+import com.linkedin.openhouse.tables.audit.model.OperationType;
+import com.linkedin.openhouse.tables.audit.model.TableAuditEvent;
+import com.linkedin.openhouse.tables.mock.properties.AuthorizationPropertiesInitializer;
+import io.micrometer.core.instrument.MeterRegistry;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.apache.iceberg.PartitionSpec;
+import org.apache.iceberg.Schema;
+import org.apache.iceberg.Table;
+import org.apache.iceberg.catalog.Namespace;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.exceptions.NoSuchTableException;
+import org.apache.iceberg.rest.RESTCatalog;
+import org.apache.iceberg.types.Types;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.mockito.ArgumentCaptor;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.web.server.LocalServerPort;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.HttpServerErrorException;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Comprehensive read-only round-trip integration tests for the Iceberg REST Catalog endpoints.
+ *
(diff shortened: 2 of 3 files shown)


.github/workflows/build-run-tests.yml
```diff
@@ -34,11 +34,20 @@ jobs:
...
           python-version: '3.12'
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+
+      - name: Run Iceberg REST Catalog Smoke Test (PyIceberg)
+        working-directory: integrations/python/dataloader
+        run: uv run python scripts/iceberg_rest_catalog_smoke.py ../../../tables-test-fixtures/tables-test-fixtures-iceberg-1.2/src/main/resources/dummy.token
```
mkuchenbecker, line 44: This is local

mkuchenbecker (commented):
Part of a coordinated review of the #689/#690/#691 stack. It was generated based on the review skills at https://github.com/mkuchenbecker/code-review-skills — linked so the criteria being judged are inspectable; this PR was judged primarily by testing-review, whose axiom is that a test is a question that can prove a claim wrong: the suite is evaluated by deriving the tests that should exist from the documented contract and diffing them against the tests that do.

Verdict for this PR: the read-fidelity tests (schema, partition spec, sort order, error envelope) are competent and worth keeping as-is. The suite's central gap: it validates with clients too old to exercise the behaviors that break current clients (inline, fix before merge), and one test is a pass that cannot fail for its stated reason. Also investigated: the standard-suite question this file's own javadoc raises. Iceberg's REST Compatibility Kit runs out-of-process with its own bundled 1.11 client, but it extends `CatalogTests` (97 test methods) whose fixtures create namespaces and tables through the catalog under test, and it has no read-only mode — essentially zero of its tests pass against a four-endpoint read-only facade. The cheapest import is the Polaris pattern this file's javadoc already names — extend `CatalogTests<RESTCatalog>` with non-runnable tests `@Disabled` with reasons — gated on a test-scoped Iceberg ≥1.7 client (`CatalogTests` first appears in 1.7; the 1.5.2 server runtime is untouched). The disabled reasons then double as the endpoint roadmap.

4 inline comments: 2 fix-before-merge, 2 follow-ups.

---
_Generated by [Claude Code](https://claude.ai/code)_

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/IcebergRestCatalogRoundTripTest.java
```diff
@@ -0,0 +1,494 @@
...
+ * <p>Follows the same validation pattern as Apache Polaris's {@code
+ * PolarisRestCatalogIntegrationBase} (which extends Iceberg's {@code CatalogTests<RESTCatalog>}).
+ * Since OpenHouse exposes a read-only Iceberg REST surface, we cannot extend {@code CatalogTests}
+ * directly (it requires write operations for test setup). Instead, tables are created via the
+ * OpenHouse API and all read operations are validated through a standard Iceberg {@link
+ * RESTCatalog} client.
+ *
+ * <p>TODO: When Iceberg is upgraded to 1.7+, this test should extend {@code
+ * CatalogTests<RESTCatalog>} with write tests disabled via {@code @Override} stubs, matching the
+ * Polaris pattern. {@code CatalogTests} was introduced in Iceberg 1.7 and is not available in our
```
mkuchenbecker, line 63: **Fix before merge — this suite validates with clients too old to hit the bugs it exists to catch.** The e2e client is Iceberg 1.5.2 (as this comment notes): no listing pagination loop, no `endpoints` awareness — so it cannot send the empty `pageToken` that returns 400 to every 1.6+ client, nor observe the forced-pagination truncation; the hand-built pagination test below always supplies a non-empty token; and the CI PyIceberg is a personal-fork git pin, attesting fork compatibility rather than what users `pip install`. Fix without any client upgrade — three plain-HTTP tests: (1) `?pageToken=` (empty) returns 200 with the full listing; (2) a request with no `pageToken` returns all rows with `next-page-token: null`; (3) truncation at real scale — create `DEFAULT_PAGE_SIZE + 1` (101) tables, list with no token, assert all 101 identifiers arrive (today no fixture exceeds 2 tables, so the truncation path is unreachable by the suite at any client version). And point the smoke test at a released PyIceberg.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/IcebergRestCatalogRoundTripTest.java
```diff
@@ -0,0 +1,494 @@
...
+  }
+
+  @Test
+  void loadTableSortOrder() {
+    Table table = restCatalog.loadTable(TableIdentifier.of(DB1, T1));
+    // Default sort order is unsorted
+    assertThat(table.sortOrder().isUnsorted()).isTrue();
+  }
+
+  @Test
```
mkuchenbecker, line 312: **Fix before merge — this test cannot fail for its stated reason.** Named for the current-snapshot contract, it asserts only that `schema()` is non-null (the comment concedes the snapshot may be null), and the schema is already asserted in more detail by four other tests — a vacuous pass advertising snapshot coverage that does not exist. Fix: make it a labeled pin of the real behavior — `currentSnapshot()` is null for a freshly created OpenHouse table — so the day the facade starts returning snapshots, the pin flips and someone looks. Or delete it as redundant.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/IcebergRestCatalogRoundTripTest.java
```diff
@@ -0,0 +1,494 @@
...
+  // Load table — metadata
+  // ---------------------------------------------------------------------------
+
+  @Test
+  void loadTableLocation() {
+    Table table = restCatalog.loadTable(TableIdentifier.of(DB1, T1));
+    assertThat(table.location()).isNotEmpty();
+  }
+
+  @Test
```
mkuchenbecker, line 298: **Follow-up — this asserts a value that can never be null.** Iceberg's `BaseTable.properties()` never returns null, so the assertion holds for any table, unconditionally. Strengthen it to assert a specific OpenHouse-injected property key is present (a real claim about load fidelity), or delete it.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/IcebergRestCatalogRoundTripTest.java
```diff
@@ -0,0 +1,494 @@
...
+  // ---------------------------------------------------------------------------
+
+  @Test
+  void configAdvertisesOnlySupportedEndpoints() {
+    ResponseEntity<String> response =
+        restTemplate.exchange(
+            baseUrl() + "/v1/config", HttpMethod.GET, authorizedRequest(), String.class);
+
+    List<String> endpoints = JsonPath.read(response.getBody(), "$.endpoints");
+    assertThat(endpoints)
```
mkuchenbecker, line 159: **Follow-up — the two suites encode contradictory contracts for endpoint order, and both hand-copy the list.** This `containsExactly` asserts an order the spec does not define for `CatalogConfig.endpoints`, while the Python smoke script compares an unordered set — and the literal list here is a drifting copy of the generated `IcebergRestOpenHouseSupport.SUPPORTED_ENDPOINTS`. Assert against the generated constant, order-independently, keeping at most one literal copy somewhere as the intentional change-detector.

PR #696, opened 2026-08-26
Add entityType discriminator and table-scoped HTS queries

ruolin59:
> Continues #683, which was opened from a fork before I had write access here. That PR carries the review history; this one is the same 17 commits on an upstream branch, rebased onto current `main`.

## Summary

This is the first PR toward supporting Iceberg views in OpenHouse. Views will share the
`(databaseId, objectId)` key space with tables, so this change adds the discriminator that tells
them apart and makes the existing table queries filter on it.

The new `entity_type` column runs from MySQL/H2 through HTS and the generated client. `VIEW` means
view; `TABLE` and a legacy `NULL` both mean table. The column is nullable and not backfilled, so
existing rows and existing table writes are unaffected. Nothing writes `VIEW` yet, so this is inert
at runtime.

Inside HTS the discriminator is the `EntityType` enum. It stays a `String` on the wire, because
`UserTable` generates the OpenAPI spec and `:client:hts`, and an enum there would make a future
entity type a breaking change for already-deployed clients. `StorageType` sets the same precedent.
A JPA `AttributeConverter` resolves a `NULL` column to `TABLE` on read, so `entity_type` is nullable
only inside MySQL and total everywhere in Java.

## JDBC methods

Entity type is chosen by calling a different method rather than by passing an argument.

| Scope | Methods |
|---|---|
| Neutral | `findBy…`, `existsBy…`, `deleteBy…`, `findById`, `existsById`, `deleteById`, `renameTableId` |
| Databases | `findAllDistinctDatabaseIds` ×2 |
| Both types | `findAllByFilters` ×2, `findAllByDatabaseIdAndTableIdLikeAllIgnoreCase` ×2 |
| Tables only *(new)* | `findTableBy…`, `findAllTablesByFilters` ×2, `findAllTablesByDatabaseIdAndTableIdLikeAllIgnoreCase` ×2 |

Nothing existing was renamed or changed behaviourally. The shared JPQL moved into constants
(`COMMON_FILTER_CLAUSES`, `PATTERN_KEY_CLAUSES`, `TABLE_ROW_PREDICATE`) so each table-scoped method
composes it, and the expansions were compared to confirm the general queries are byte-identical to
before. View methods follow the same shape and land with the code that calls them.

Three things shaped this:

- Methods returning database names stay unfiltered. Filtering them would hide a database that
  contains only views, even though it exists and is addressable.
- Point reads and mutations on the shared key stay neutral, since the PUT, delete, and restore paths
  need to see a row of any type to detect a collision.
- The pattern methods kept dedicated table-scoped versions instead of folding into
  `findAllByFilters`, which matches `tableId` exactly. Merging a `LIKE` parameter would make `_`
  behave as a wildcard, and identifiers here often contain underscores.

`findAllByDatabaseIdIgnoreCase` ×2 were dropped. The paged one was never called, and paged
`listTables` already went through `findAllByFilters`, so `findAllTablesByFilters` covers both.

## Callers need no type logic

HTS returns the correct rows, so none of its callers check entity type. A view 404s from the table point
read, surfaces as `HouseTableNotFoundException`, and reads as absent — which makes `doRefresh`,
`dropTable`, `findTableRefById`, and rename-source all correct as written. `dropTable` matters here
because it bypasses `loadTable` to survive corrupted metadata, so a guard on the refresh path would
have missed it.

Name occupancy — "what is at this key?" — needs to see rows of any type, and its only callers are
`CREATE TABLE` and the rename-destination check. That lands with the view work.

## The tables service stays out of it

An earlier revision put an `entityType` field on the internal `HouseTable` pointer. It had no
consumer, and it was fed from an `openhouse.entityType` property that nothing writes, so it was
always null — which then forced a null check into `HouseTableMapper.stripOhNamespace`, a method
MapStruct applies to every String property on the mapper.

Both are gone. `stripOhNamespace` is byte-identical to `main` again, and the whole `iceberg/` diff
is one `@Mapping(target = "entityType", ignore = true)`. The discriminator is owned by HTS; the
tables service has no knowledge of it.

## Tests

Existing tests cover every changed call site in `UserTablesServiceImpl` — `testGetUserTables`,
`testUserTableQuery`, `testGetUserTablesWithTablePattern`, `testGetUserTablesWithSearchFilter`,
`testUserTableGet`, `testListDatabases`. All still pass unmodified: the diff on that test class is
**247 insertions, 0 deletions**, and no existing assertion anywhere in this PR was deleted or
relaxed. Since the table behaviour was meant to be unchanged, those tests are the regression proof.

New tests were written before the implementation. They cover both-types vs tables-only results
across the filter and pattern families, page counts with views interleaved, the point read treating
a view as absent while the neutral read still returns it, and unrecognised discriminators failing
closed. Page assertions check content, size, total elements, and total pages, so filtering a
returned page instead of the query would fail them. Case handling is asserted in Java, since H2 in
`MODE=MySQL` is case-sensitive and production MySQL is not.

| Module | Tests | Failures |
|---|---|---|
| `services:housetables` | 185 | 0 |
| `iceberg:openhouse:internalcatalog` | 82 | 0 |
| `services:tables` | 475 | 0 |
| `iceberg:openhouse:htscatalog` | 20 | 0 |
| `tables-test-fixtures_2.12` (Iceberg 1.2) | 8 | 0 |
| `tables-test-fixtures-iceberg-1.5_2.12` | 8 | 0 |

Both fixture variants compile and the 1.2 fixture's tests run, since `HouseTableRepository` is
inherited by Spring Data proxies in published fixture code.

## Rollout

`schema.sql` uses `CREATE TABLE IF NOT EXISTS`, so production needs
`ALTER TABLE user_table_row ADD COLUMN entity_type VARCHAR(128) DEFAULT NULL` **before deploying this pr**. No backfill needed. Deploy HTS before the tables service, since the filtering lives in HTS.

That DDL is also recorded under `services/housetables/ddl/`, as a baseline snapshot plus the
`ALTER`. The files are inert — outside `src/main/resources`, so Spring cannot execute them and
Gradle does not package them — and exist only so the sequence of schema changes is captured in the
repository. The baseline is derived from `schema.sql` and is marked pending verification against
production `SHOW CREATE TABLE`. Migration tooling was evaluated and deferred; Flyway is deprecated
internally in favour of Pretzel, tracked in BDP-108649.


mkuchenbecker:
Why are we stacked on this pr vs https://github.com/linkedin/openhouse/pull/683

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
```diff
@@ -39,11 +39,37 @@ Optional<UserTableRow> findByDatabaseIdIgnoreCaseAndTableIdIgnoreCase(
...
+   * table point read in the tables service. The neutral {@link
+   * #findByDatabaseIdIgnoreCaseAndTableIdIgnoreCase} above stays unfiltered because the writers
+   * must see a row of any type to detect a collision at a shared key.
+   */
+  @Query(
+      "SELECT u FROM UserTableRow u WHERE "
+          + "lower(u.databaseId) = lower(:databaseId) AND "
+          + "lower(u.tableId) = lower(:tableId) AND "
+          + TABLE_ROW_PREDICATE)
+  Optional<UserTableRow> findTableByDatabaseIdIgnoreCaseAndTableIdIgnoreCase(
```
abhisheknath2011, line 67: We need something similar to get view as well right? This query has table row filter, so will return only table.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -302,7 +302,8 @@ private Page<UserTableDto> listTables(UserTable userTable, int page, int size, S
     return METRICS_REPORTER.executeWithStats(
         () ->
             htsJdbcRepository
-                .findAllByFilters(userTable.getDatabaseId(), null, null, null, null, null, pageable)
+                .findAllTablesByFilters(
```
abhisheknath2011, line 305: So most of the methods are renamed because as they are solely for tables. So in that case what are the methods for view that we are supporting in the initial version. I see on common filter for both tables and views (returns list of views). Are there any other methods supported?

ruolin59: this pr was originally intended to be "just add the entity_type column and make sure tables paths don't access it", so there's no view access methods here. The view access methods are in the next pr here: https://github.com/linkedin/openhouse/pull/697

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
ruolin59: see https://github.com/linkedin/openhouse/pull/696#discussion_r3882306455
the views-related queries are in the other pr

aastha25, line 67: maybe a comment for the follow up PR, but in general, there should be some query patterns for findAllTablesAndViewsByDatabaseId(AndTableId)LikeAllIgnoreCase. User initiated queries like - list tables in a db should ideally, and have historically, returned all tables and views existing in a db. Why do we only want to get only tables / only views at this layer and let caller initiate 2 calls to HTS? instead we could return all tables + all views and let caller handle usecases in a targetted manner.

aastha25:
lgtm, in scope of first adding a nullable column 'entity_type' to the mysql table `user_table_row` and making the compatible changes with the table API endpoints.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/mapper/HouseTableMapper.java
```diff
@@ -41,7 +41,11 @@ public HouseTable toHouseTable(TableMetadata tableMetadata, FileIO fileIO) {
   @Mappings({@Mapping(target = "tableLocation", source = "userTable.metadataLocation")})
   public abstract HouseTable toHouseTable(UserTable userTable);
 
-  @Mappings({@Mapping(target = "metadataLocation", source = "houseTable.tableLocation")})
+  // HouseTable has no entityType; the discriminator is owned by HTS, not the tables service.
```
abhisheknath2011: Can we update the comment here? HouseTables itself is HTS.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/mapper/HouseTableMapper.java
```diff
@@ -41,7 +41,11 @@ public HouseTable toHouseTable(TableMetadata tableMetadata, FileIO fileIO) {
...
+  @Mappings({
+    @Mapping(target = "metadataLocation", source = "houseTable.tableLocation"),
+    @Mapping(target = "entityType", ignore = true)
```
abhisheknath2011, line 48: Why ignore is true here? Entity type is needed for both tables and views with the new changes?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/model/EntityTypeConverter.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @Override
+  public String convertToDatabaseColumn(EntityType entityType) {
+    return entityType == null ? null : entityType.name();
+  }
+
+  @Override
+  public EntityType convertToEntityAttribute(String columnValue) {
+    if (columnValue == null) {
+      return EntityType.TABLE;
```
abhisheknath2011, line 28: Can we add a comment here as null is treated as TABLE?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
```diff
@@ -90,6 +100,60 @@ Iterable<UserTableRow> findAllByFilters(
       String storageType,
       Long creationTime);
 
+  @Query(
+      "SELECT u FROM UserTableRow u WHERE " + PATTERN_KEY_CLAUSES + " AND " + TABLE_ROW_PREDICATE)
+  Iterable<UserTableRow> findAllTablesByDatabaseIdAndTableIdLikeAllIgnoreCase(
```
abhisheknath2011, line 105: Some of the method can be common for tables and views like this one. This is table name pattern based filter so this could be applicable for views as well. This could be consolidated in the next PR.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableHtsJdbcRepository.java
```diff
@@ -90,6 +100,60 @@ Iterable<UserTableRow> findAllByFilters(
...
+      @Param("storageType") String storageType,
+      @Param("creationTime") Long creationTime,
+      Pageable pageable);
+
+  @Query(
+      "select DISTINCT u from UserTableRow u where "
+          + COMMON_FILTER_CLAUSES
+          + " AND "
+          + TABLE_ROW_PREDICATE)
+  Iterable<UserTableRow> findAllTablesByFilters(
```
abhisheknath2011, line 149: Can we also list down what all methods would be supported on views? Based on that could come up with methods for tables or views or both?

abhisheknath2011 (commented):
Thanks @ruolin59. Added some comments and some follow up comments for the next PR - https://github.com/linkedin/openhouse/pull/697.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/mapper/HouseTableMapper.java
ruolin59: updated, but actually HouseTableMapper resides inside the tableService rather than HTS

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/mapper/HouseTableMapper.java
ruolin59: see above. Because this is still inside the Table Service and not in HTS, there is no entityType being set yet, as there is still no implementation in Table Service. Once the TableService is implemented properly end-to-end to ensure that entityType is properly set, I can flip this to `ignore=false`

abhisheknath2011 (dismissed):

aastha25 (approved):

abhisheknath2011 (approved):

PR #697, opened 2026-08-26
HTS API: Give views a typed lifecycle in HTS

ruolin59:
## What this does

This adds everything a view needs to be created, read, listed and dropped through HTS, and closes the gap it left open: type was selected by method on the read paths, but key mutations were still neutral.

**Base is `views-entity-type-discriminator` which added the `entity_type` column to handle the new "VIEW" type.** HTS must ship before any tables-service code that calls the new paths.

~600 lines of production code across 12 files; the rest of the diff is tests.

## Endpoints

All five view routes are new. Table routes are pre-existing, but PUT gains ingress stamping and DELETE and rename become table-scoped.

| Capability | Table | View |
|---|---|---|
| Point read | `GET /hts/tables` | `GET /hts/views` |
| Query | `GET /hts/tables/query` | `GET /hts/views/query` |
| Paged query | `GET /v1/hts/tables/query` | `GET /v1/hts/views/query` |
| Write | `PUT /hts/tables` | `PUT /hts/views` |
| Hard delete | `DELETE /hts/tables` | `DELETE /hts/views` |
| Soft delete | `DELETE /v1/hts/tables?isSoftDelete=` | **absent** — views are never soft-deleted |
| Rename | `PATCH /hts/tables/rename` | **absent** — view rename is unsupported |

Plus `GET /hts/entities`, a neutral point read answering what occupies a key, of either type, for collision detection that must see both.

## Repository surface

15 unchanged, 2 modified, 1 removed, 12 added — 29 methods.

Type is selected by the method or route you call; no public caller supplies a type selector. `renameTableId` stamps the discriminator from a string constant in the query itself, so no caller supplies it. That constant is not derived from `EntityType`: `EntityType.TABLE.name()` is a method call and therefore illegal in an annotation value, so renaming an enum constant would not update the query.

## Design notes

**Key mutations are type-scoped.** One conditional statement carries the type predicate, so a wrong-type delete or rename affects zero rows and returns 404 with the row intact — rather than read-then-check-then-delete, a TOCTOU window on an irreversible operation.

**Inherited key-addressed deletes are sealed.** `deleteById`, `delete(T)`, `deleteAllById` and `deleteAll(Iterable)` throw, so the only way to remove a row is a method that names the type. No-arg `deleteAll()` stays: it addresses no key, so it cannot confuse one type for another.

[...]

PR #698, opened 2026-08-26
E2E: Cover House Tables against a deployed MySQL

ruolin59:
## Summary

Add deployment-level House Tables E2E coverage against MySQL initialized from the current recorded DDL. This PR now targets `main`; it no longer depends on the closed #697 branch. Production service code and DDL are unchanged from upstream.

## Changes

- [ ] Internal API Changes
- [x] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [x] Documentation
- [x] Tests

- Add an `oh-only-mysql` recipe that applies `services/housetables/ddl/*.sql` in filename order on a fresh volume. HTS waits for an authenticated TCP schema query to succeed, avoiding the reproduced first-start race with MySQL initialization.
- Add 25 HTTP/SQL E2E cases covering typed table/view lifecycles, conflicts, legacy NULL discriminators, restore stamping, corrupt-value diagnostics, and mixed-case identifiers. View listings use the current `/v1/hts/views/query` API, not the removed legacy route.
- Check the updated live-table column order, identifier/metadata widths, nullable version/discriminator, recorded table collations, and `idx_user_table_upper_db_table` functional index. This detects stale volumes or accidental use of the service's different bootstrap schema.
- Run both the existing tables integration suite and the new HTS suite against MySQL in CI. Readiness polling has bounded HTTP requests, validates expected statuses, and prints diagnostics on timeout. `--require-database` makes unavailable database coverage fail instead of skip.
- Document local setup, strict versus optional database mode, and the fresh-volume requirement.
- Add `python3 scripts/python/run_local_e2e.py` as the README-documented, one-command local entrypoint. It reuses the Compose recipe, builds both services, installs test dependencies in a temporary virtual environment, waits for readiness, and runs both suites against a fresh isolated database. Random loopback ports avoid conflicts with existing stacks; its own containers and volumes are removed afterward. The existing tables test accepts an optional `OPENHOUSE_TABLES_HOST`, preserving localhost:8000 by default.

## Testing Done

Built the current upstream services with Java 17:

```bash
./gradlew :services:housetables:bootJar :services:tables:bootJar -x CopyGitHooksTask
```

Started the recipe as Compose project `oh698`, with a local-only override exposing MySQL on `127.0.0.1:13306` and giving OPA a unique container name. The existing MySQL container on port 3306 was left untouched. Recreated the deployment with a fresh anonymous MySQL volume to verify the DDL sequence and health-gated startup, then ran:

```bash
HTS_DB_PORT=13306 python3 scripts/python/hts_integration_test.py http://localhost:8001 --require-database
# 25 passed, 0 skipped

python3 scripts/python/integration_test.py tables-test-fixtures/tables-test-fixtures-iceberg-1.2/src/main/resources/dummy.token
# All tests passed successfully
```

Both suites passed on the warm deployment and again after cold initialization on MySQL 8.4.11. With the database port deliberately unreachable, strict mode exited nonzero before running HTTP cases; optional mode reported **21 passed, 4 skipped**.

Also ran the new entrypoint end to end:

```bash
python3 scripts/python/run_local_e2e.py
# 25 passed, 0 skipped
# All tests passed successfully
# Both E2E suites passed; test containers and database removed.
```

This separate deployment used dynamically assigned ports while the existing stacks remained running. Verified scoped container cleanup after both a failed startup-readiness attempt and the successful run.

`./gradlew spotlessCheck -x CopyGitHooksTask` passed. `CopyGitHooksTask` was excluded because it assumes `.git` is a directory and fails in a Git worktree; the commit/push formatting check was run directly with that task excluded. Compose configuration validation and `git diff --check` also passed.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

CI now exercises the MySQL recipe instead of `oh-only`; the H2 recipe remains unchanged but is not exercised by this CI job. Corrupt-discriminator expectations are pinned to the recorded `utf8mb4_0900_ai_ci` NO PAD collation, including the explicit 404-vs-500 typed/neutral disagreement for malformed values.

The recorded live and soft-deleted table schemas are production-verified upstream. `job_row` and `table_toggle_rule` remain bootstrap-derived approximations, as documented in the baseline. Docker initialization runs only on an empty data volume; it does not migrate an existing volume.

The branch is rebased onto `main` with only the E2E changes. Index metadata is checked, but query-plan/index-selection assertions are intentionally out of scope.

### DIFF
--- a/.github/workflows/build-run-tests.yml
+++ b/.github/workflows/build-run-tests.yml
@@ -23,11 +23,28 @@ jobs:
       - name: Build with Gradle
         run: ./gradlew clean build
 
+      # The MySQL-backed variant of oh-only. It costs one extra container and lets the
+      # House Tables tests below reach the database directly, which is the only way to
+      # cover a legacy NULL entity_type and to read the column back after a restore.
       - name: Start Docker Containers
-        run: docker compose -f infra/recipes/docker-compose/oh-only/docker-compose.yml up -d --build
+        run: docker compose -f infra/recipes/docker-compose/oh-only-mysql/docker-compose.yml up -d --build
 
-      - name: Wait for Docker Containers to start
-        run: sleep 30
+      - name: Wait for Docker Containers to be ready
+        run: |
+          echo "Waiting for openhouse-tables and openhouse-housetables..."
+          for i in $(seq 1 60); do
+            tables_status=$(curl -s -o /dev/null -w "%{http_code}" http://localhost:8000/v1/databases 2>&1 || echo "000")
+            hts_status=$(curl -s -o /dev/null -w "%{http_code}" "http://localhost:8001/hts/tables/query" 2>&1 || echo "000")
+            echo "Poll $i: tables=$tables_status hts=$hts_status"
+            if [ "$tables_status" != "000" ] && [ "$hts_status" = "200" ]; then
+              echo "Services ready after $((i * 5))s"
+              exit 0
+            fi
+            sleep 5
+          done
+          echo "Timed out waiting for services after 300s"
+          docker compose -f infra/recipes/docker-compose/oh-only-mysql/docker-compose.yml ps
+          exit 1
 
       - name: Set up Python
         uses: actions/setup-python@v6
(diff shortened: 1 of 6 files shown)


ruolin59:
Superseded by #769, which preserves the deployment tests, adds production-DDL index regression coverage, and is rebased onto the prerequisite fix in #763.

PR #703, opened 2026-08-31
HTS API alternative implementation: Give views a typed lifecycle in HTS (alternative to #697)

ruolin59:
## What this is

An **alternative implementation of BDP-108627**, the same ticket as **#697**. It is meant to be compared against that PR, not stacked on it. Both branch from `rufan/views-entity-type-discriminator`.

The premise: keep #697's HTTP contract and its SQL layer as they are, and rebuild the part between them — the CRUD wiring and error handling. One deliberate exception, called out below: views are paginated-only, so the non-paginated `GET /hts/views/query` that #697 mirrored from tables is dropped.

| | #697 | this PR |
|---|---|---|
| HTTP API | — | **identical except views are paginated-only** — the one non-paginated view route is removed by design (see below); every other route, verb, param, body and status is unchanged |
| SQL / JPQL | — | **identical** apart from removing the two now-dead non-paginated view finders; every constant, table finder and paginated view finder is verbatim |
| Controller ↔ repository wiring | as landed | rebuilt |

## What is deliberately unchanged

**The HTTP surface.** `GET /hts/entities` and every pre-existing table route keep their paths, verbs, parameters, defaults, bodies and status codes. The view routes match #697's with one exception: there is no non-paginated `GET /hts/views/query` (see the next section).

**The SQL layer.** Every JPQL constant (`COMMON_FILTER_CLAUSES`, `PATTERN_KEY_CLAUSES`, `TABLE_ROW_PREDICATE`, `VIEW_ROW_PREDICATE`, `STAMP_TABLE_TYPE`), every query and `countQuery`, the typed conditional deletes returning an affected-row count, and the type-scoped rename are carried over verbatim. `TABLE_ROW_PREDICATE` keeps its `IS NULL` arm, so legacy rows still read as tables.

## JDBC repository surface (`UserTableHtsJdbcRepository`)

An accounting of the MySQL-backed repository methods, grouped by the entity type each targets. "New" and "reshaped" are relative to the base branch `rufan/views-entity-type-discriminator`, which already carried the table-typed **reads** (the frozen tables path) but only **neutral writes**.

**View-scoped — every method is new in this PR** (the repository had no view methods before):

| Method | Returns | Purpose |
|---|---|---|
| `findViewByDatabaseIdIgnoreCaseAndTableIdIgnoreCase(db, table)` | `Optional<UserTableRow>` | point read under `VIEW_ROW_PREDICATE` |
| `findAllViewsByFilters(db, table, version, metadataLocation, storageType, creationTime, Pageable)` | `Page<UserTableRow>` | paginated multi-attribute filter |
| `findAllViewsByDatabaseIdAndTableIdLikeAllIgnoreCase(db, pattern, Pageable)` | `Page<UserTableRow>` | paginated name-pattern list |
| `deleteViewByDatabaseIdIgnoreCaseAndTableIdIgnoreCase(db, table)` | `int` | conditional delete under `VIEW_ROW_PREDICATE`, affected-row count |
| `deleteViewById(key)` | `int` | default delegating to the delete above |

Plus the constants `VIEW` and `VIEW_ROW_PREDICATE`. There is deliberately **no** non-paginated view finder, no `findViewById` default, and no view rename: views are paginated-only and are not renameable.

**Table-scoped:**

- `findTableByDatabaseIdIgnoreCaseAndTableIdIgnoreCase → Optional`, `findAllTablesByDatabaseIdAndTableIdLikeAllIgnoreCase` (`Iterable` and `Page`), `findAllTablesByFilters` (`Iterable` and `Page`) — *inherited from base*, unchanged.
- `deleteTableByDatabaseIdIgnoreCaseAndTableIdIgnoreCase → int` and `deleteTableById(key) → int` — **new**: replace the base's single neutral `void delete…`, now scoped by `TABLE_ROW_PREDICATE` and returning an affected-row count so the service maps missing and wrong-type alike to 404.
- `renameTableId(…) → int` — **reshaped**: the base's neutral `void renameTableId` gains `TABLE_ROW_PREDICATE` scoping and `STAMP_TABLE_TYPE` (rewriting a legacy null or non-canonical spelling as the row moves), and returns a row count. Table-only because views are not renameable.

**Both (type-neutral):**

- `findByDatabaseIdIgnoreCaseAndTableIdIgnoreCase → Optional` and `existsByDatabaseIdIgnoreCaseAndTableIdIgnoreCase → boolean` — *inherited*; the writers read through these so they can see a row of any type and detect a collision at a shared key, which is why they stay unfiltered.
- `findAllByDatabaseIdAndTableIdLikeAllIgnoreCase` (`Iterable` and `Page`), `findAllByFilters` (`Iterable` and `Page`), `findAllDistinctDatabaseIds()` and its paginated form — *inherited*.
- `findById` / `existsById` defaults — *inherited*.
- `deleteById`, `delete`, `deleteAllById`, `deleteAll(Iterable)`, `deleteAll()` — **sealed in this PR** to throw `UnsupportedOperationException`, forcing callers onto the typed deletes; the base's neutral `deleteBy…` derived delete is removed.

## What is rebuilt, and why

**A view write cannot persist a table.** `putUserTable` and `putUserView` each supply their own `EntityType` to one private persistence primitive, so the named method — not the payload — is the authority. In #697, `putView` delegates to `putUserTable` and relies on the controller having stamped the type; a direct Java caller could persist the wrong one.

**"Not found" cannot be manufactured from a failure.** The new point reads return `Optional`, and only the handler turns empty into a 404. A dependency or hydration failure travels as an exception and surfaces as a 500. This matters most for the neutral read, where the ticket calls out "the name is free" as the dangerous default.

**The service owns its view-query vocabulary.** View queries cross the boundary as an owned `UserViewQuery`, built once in the handler after validation, so the transport `UserTable`'s nullability and its inert `entityType` stop at the handler. The invalid "table pattern without a database" state cannot be constructed.

**Corruption is translated at the repository boundary, structurally.** An `@Around` aspect (`UserTableRepositoryTranslationAspect`) wraps every `UserTableHtsJdbcRepository` method: when a corrupt discriminator surfaces during hydration it is rethrown as an unwrapped `CorruptEntityTypeException`, and every other `DataAccessException` is rethrown exactly as it arrived, so non-corrupt failures behave as before. Because it is an aspect over the whole repository rather than a wrap at each call site, a read added later is covered without anyone remembering to wrap it. The interception point is load-bearing and can fail silently — Spring Data's own proxy holds `PersistenceExceptionTranslationInterceptor`, and the aspect must land in the outer proxy to see the already-translated `DataAccessException` — so a structural test pins it (`RepositoryTranslationAspectTest`) and a Spring Data upgrade that relocated the interceptor fails loudly rather than downgrading every corrupt read to a generic 500.

Because corruption is unwrapped before it leaves the repository, the shared advice gains **one handler and no ORM vocabulary**. #697 instead teaches `services/common` to import and unwrap `JpaSystemException` and `InvalidDataAccessApiUsageException`.

`CorruptEntityTypeException` does not extend `IllegalArgumentException`, so stored-data corruption stays distinguishable from client-input error.

**Views are paginated-only.** #697 gave views a non-paginated `GET /hts/views/query` mirroring `GET /hts/tables/query`. This PR drops it: views can be listed only through the paginated `GET /v1/hts/views/query`. The non-paginated table query is legacy — it still backs the tables service's `/v0` and `/v1` list-all, and its empty-filter form returns database names rather than tables — and a trace confirmed neither the tables service nor the internal catalog references any view query path. Mirroring that shape would inherit tech debt into a greenfield surface with no caller, so on this new path we took the more principled option, which also removes the unbounded-query concern for views entirely. Tables are untouched: `GET /hts/tables/query` stays, because it has live callers.

**Smaller ones.** PUT ingress validation is an injectable, unit-testable component that throws `RequestValidationFailureException` directly, rather than a private static helper in the controller. No-arg `deleteAll()` is sealed alongside the four key-addressed deletes, with teardown moved to a test fixture. View metrics are owned by housetables; the pre-existing `HTS_`/`REPO_` constants stay in `services/common` because `services/tables` consumes 19 of them.

## Footprint on shared code

```
CorruptEntityTypeException.java   +19   (new)
OpenHouseExceptionHandler.java    +22   (one @Hidden handler, 0 deletions)
```

`ServiceAuditAspect` and every other file in `services/common` are byte-identical to base. ORM imports added to `services/common`: **0** (#697 adds 3). The translation mechanism itself — a small `CorruptEntityTypeTranslation` utility and the aspect that applies it — lives in `services/housetables`, the module that already owns JPA.

## Deliberate asymmetries

Two remain, both in signatures rather than behaviour: `getUserTable` signals absence by exception while the new point reads return `Optional`, and table queries take the transport `UserTable` while view queries take an owned type.

Both were measured rather than assumed. Unifying the absence convention costs 16 pre-existing test edits and risks a silent wire change — the current 404 carries `cause: null` because it wraps a message-less `NoSuchElementException`, and no test pins that field. Unifying the query vocabulary means rewriting 11 private query helpers and 3 branch predicates that encode the frozen empty-filter behaviour. Both are worth doing on their own terms, with their own review; neither has view-specific content.

## Not adopted

- **A generic corruption 500 body.** Would make it the only differently-shaped 500 in the service.
- **Removing the `.stacktrace(...)` sites from the shared handler.** These do not leak: `ServiceAuditAspect` already strips the stack trace from the client body and keeps it only in the internal audit event, so deleting the sites would degrade the audit trail without changing any client response. The populate-then-strip indirection is real, but it is a `services/common` contract shared by tables, jobs and housetables — a whole-platform refactor, not a views concern. The views-specific error-handling improvement Mike's review asked for was instead made through the corruption path (`CorruptEntityTypeException` + the repository aspect), which gives a corrupt read a proper diagnostic 500 rather than raw ORM vocabulary.
- **The soft-delete version TOCTOU.** Pre-existing; needs a version-guarded delete and its own interleaving test, so tracked separately rather than folded into a views PR.

The production collation, an open question across both PR stacks, is now confirmed `utf8mb4_0900_ai_ci` — accent-insensitive and NO PAD. This needs no code change: it means a corrupt `'TÁBLE'` matches the type predicate and fails hydration with a diagnostic 500, and a corrupt `'TABLE '` is invisible to the typed routes. Both require a direct database write, since the API writes only canonical enum names, so neither is a reachable state.

## Testing

| Module | Tests | Failures |
|---|---|---|
| `services:housetables` | 404 | 0 |
| `services:common` | 12 | 0 |
| `services:tables` | 526 | 0 |
| `services:jobs` | 52 | 0 |

`tables` (526) and `jobs` (52) match their base counts after this branch was rebased onto the current base. `:client:hts` regenerates cleanly with **five** new operations — `getEntity`, `getUserView`, `getPaginatedUserViews`, `putUserView`, `deleteView` (no non-paginated `getUserViews`). The generated OpenAPI response sets match #697's for every route that still exists, including a documented `500` on the three GET routes that can reach a corrupt row (`/hts/entities`, `/hts/views`, `/v1/hts/views/query`).

The change was also driven through #698's deployment e2e suite against a real MySQL 8.4 container: **23 passed, 0 skipped**. That suite was written against #697's API; the only edit needed was to route its view-listing through the paginated endpoint, since the non-paginated one is intentionally gone — every other assertion passed unchanged, which is independent evidence the rest of the contract is indistinguishable to a client. The confirmed production collation (`utf8mb4_0900_ai_ci`) is exercised by that run.

Pre-existing tests were changed only where the frozen contract or an adopted decision forced it, and each was strengthened rather than weakened; the three teardown edits come from sealing `deleteAll()`. No assertion was deleted or relaxed.

## A note on a pre-existing flake

`UserTablesServiceTest.testUserTablePurge` failed twice under concurrent multi-module load and passed 6/6 in isolation. `SoftDeletedUserTablesMapper` derives `deletedAtMs` from `Instant.now()`, and that field is part of the soft-deleted composite key; the test soft-deletes the same table twice and expects two rows, so two calls in the same millisecond upsert into one. The mapper, the key and that test are untouched here, and #697 carries identical code.

## Review process

Design-first: an implementation plan reviewed independently over two rounds, tests written before the production logic and reviewed against it, then an implementation review, followed by several rounds of simplification that removed roughly 1,400 lines of machinery in favour of the smallest thing that holds the contract. Load-bearing mechanics were verified empirically rather than assumed — most importantly the aspect's interception point, proven against Spring 5.3.25 by dumping the nested proxy chains: the aspect sits in a separate outer proxy from Spring Data's exception-translation interceptor, so no `@Order` value could move it, and a negative control (aspect disabled) confirmed the corrupt-row tests genuinely depend on it.


services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/validator/EntityTypeIngressValidator.java
```diff
@@ -0,0 +1,39 @@
...
+import com.linkedin.openhouse.housetables.model.EntityType;
+import org.springframework.stereotype.Component;
+
+/**
+ * Resolves the entity type of a PUT payload at ingress, ahead of every other validation. The wire
+ * field stays nullable for rolling compatibility: a payload may agree with its route or stay
+ * silent, never override it.
+ */
+@Component
+public class EntityTypeIngressValidator {
```
abhisheknath2011: Ingress also means API gateway layer. In which layer this validator is invoked?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -41,13 +43,18 @@ public class UserHouseTablesController {
...
+  private static final String HTS_VIEWS_GENERAL_ENDPOINT = "/hts/views";
+  private static final String HTS_VIEWS_QUERY_ENDPOINT_V1 = "/v1/hts/views/query";
 
   @Autowired private UserTableHtsApiHandler tableHtsApiHandler;
 
   @Autowired private SoftDeletedUserTableHtsApiHandler softDeletedTablesHtsApiHandler;
 
   @Autowired private UserTablesMapper userTablesMapper;
 
+  @Autowired private EntityTypeIngressValidator entityTypeIngressValidator;
```
abhisheknath2011: As this used at controller layer, can just keep the validator name as just EntityTypeValidator?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/api/validator/EntityTypeIngressValidator.java
ruolin59: this should be the ingress for HouseTables service, aka HTS

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
ruolin59: yeah i also think it's a bit confusing which file's for which service, I'll look into a better name for this

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/CorruptEntityTypeTranslation.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Set;
+import java.util.concurrent.Callable;
+import org.springframework.dao.DataAccessException;
+
+/**
+ * Containment for the one ORM detail housetables cannot ignore: {@code EntityTypeConverter} throws
+ * on a corrupt discriminator, and Spring's persistence exception translation buries that inside a
+ * {@link DataAccessException} whose message is the wrapper's rather than the diagnostic.
+ */
+public final class CorruptEntityTypeTranslation {
```
abhisheknath2011, line 16: Seems like translation is happing as part of JPA layer exception. Can we make this code generic so that this can be applied to other exception as well if needed?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/CorruptEntityTypeTranslation.java
```diff
@@ -0,0 +1,58 @@
...
+
+  /** Bounds the cause walk, so a cyclic chain terminates instead of spinning. */
+  private static final int CAUSE_CHAIN_MAX_DEPTH = 20;
```
abhisheknath2011, line 24: How did arrive at this max depth? Should we increase the depth so that it covers important stacktrace?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/CorruptEntityTypeTranslation.java
ruolin59: Kept it specific on purpose. There's exactly one exception that needs unwrapping today (`CorruptEntityTypeException`), and this util is actually what a much larger read-repository abstraction earlier in this PR got collapsed down to. The reusable part — the cause walk in `findCorruptEntityTypeCause` — is already a generic algorithm; only the target type is fixed.

What I'd rather not generalize yet is the policy in `translating()` (unwrap corruption, rethrow everything else as the same instance): a second exception would likely want a different policy — its own HTTP status, maybe no unwrap at all — which is better designed against a real second case than guessed now. If one shows up, generalizing the finder to `findCause(Throwable, Class<T>)` is a ~2-line change.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/CorruptEntityTypeTranslation.java
ruolin59: Good catch that this reads ambiguously. `CAUSE_CHAIN_MAX_DEPTH` bounds the cause chain (`getCause()` hops between nested exceptions), not stack-trace frames — so it has no effect on how much of the stack trace is captured (that's preserved via exception chaining regardless).

It isn't really load-bearing for termination either: the identity visited-set already guarantees we stop (finitely many distinct causes), so the cap is just a secondary bound on a pathologically long acyclic chain. The real corruption cause sits ~2-3 hops down (`DataAccessException` → optional JPA wrapper → `CorruptEntityTypeException`), so 20 is already generous — no need to increase. Reworded the comment in 25a5b310 to make the cause-chain-vs-stack-trace distinction explicit.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableRepositoryTranslationAspect.java
```diff
@@ -0,0 +1,47 @@
...
+ * wraps it replaced did. That nesting is structural, not an ordering: no {@code @Order} value can
+ * move this advice inside Spring Data's chain, which is why none is declared.
+ * RepositoryTranslationAspectTest pins both halves.
+ *
+ * <p>Advising writes too is harmless: {@code translating} rethrows every non-corruption {@code
+ * DataAccessException} as the very same instance, so the conflict mapping is untouched.
+ */
+@Aspect
+@Component
+public class UserTableRepositoryTranslationAspect {
```
abhisheknath2011, line 28: Why do we need aspect for translation? Trying to understand the context behind this.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/model/UserViewQuery.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * The vocabulary the view query service accepts, built in the handler so no query-shaped {@code
+ * UserTable} reaches the service and the inert {@code entityType} is discarded there.
+ *
+ * <p>Only the three reachable states are constructible: there is no factory for a table pattern
+ * without a database, which the validator already rejects.
+ */
+@EqualsAndHashCode
+@ToString
+public final class UserViewQuery {
```
abhisheknath2011, line 16: Why do we need a separate class for mapping/validating query params specifically for tableId and databaseId?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
```diff
@@ -95,15 +135,33 @@ public Page<UserTableDto> getAllUserTables(
...
+    // names the occupant rather than the requested type.
+    if (existingUserTableRow.isPresent()
+        && existingUserTableRow.get().getEntityType() != entityType) {
+      throw new AlreadyExistsException(
+          existingUserTableRow.get().getEntityType().name(),
+          userTable.getDatabaseId() + "." + userTable.getTableId());
+    }
 
+    // Overwritten before mapping, so no transport spelling reaches the enum boundary.
+    UserTable ownedEntity = userTable.toBuilder().entityType(entityType.name()).build();
```
abhisheknath2011:
```suggestion
    UserTable updatedEntity = userTable.toBuilder().entityType(entityType.name()).build();
```

services/housetables/src/test/java/com/linkedin/openhouse/housetables/e2e/fixture/UserTableRawSeeder.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.stereotype.Component;
+
+/**
+ * Plants rows whose {@code entity_type} column is SQL NULL, which only rows predating the
+ * discriminator may be. JPA cannot express them any more: {@code EntityTypeConverter} rejects a
+ * null write so that the legacy population the {@code IS NULL} predicate arm carries cannot grow.
+ */
+@Component
+public class UserTableRawSeeder {
```
abhisheknath2011, line 15: Where this is used? Why JPA can't handle this usecase. Also why do we need this as well the new create will have either entity type as `table` or `view`.

services/housetables/src/test/java/com/linkedin/openhouse/housetables/e2e/fixture/UserTableStoreCleaner.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+@Component
+public class UserTableStoreCleaner {
+
+  @PersistenceContext private EntityManager entityManager;
+
+  /** Removes every row regardless of discriminator, so a corrupt row cannot survive a teardown. */
+  @Transactional
+  public void clear() {
+    entityManager.createQuery("DELETE FROM UserTableRow").executeUpdate();
```
abhisheknath2011, line 21: Why do we need this? There is no filter condition.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableRepositoryTranslationAspect.java
ruolin59: This is addressing two #697 blocking comments that got deferred there: [converter throws on corruption during hydration but nothing translates it before it leaves housetables](https://github.com/linkedin/openhouse/pull/697#discussion_r3876601775), and [common advice having to import JpaSystemException etc](https://github.com/linkedin/openhouse/pull/697#discussion_r3876601781). The ask was to translate at the housetables persistence boundary. The aspect is that boundary. Shared advice (what #697 does) makes common import ORM types, and a per-call-site try/catch is fragile since hydration is lazy (a findAll might not throw til it's iterated, outside the try). The aspect wraps every repo method instead, so every read incl future ones gets translated at the edge and common stays ORM-free.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/model/UserViewQuery.java
ruolin59: This is the fix Mike asked for on #697. He named UserViewQuery in [these](https://github.com/linkedin/openhouse/pull/697#discussion_r3876601708) [two](https://github.com/linkedin/openhouse/pull/697#discussion_r3876601786) comments. Point is to keep the transport UserTable (nullable everything + an inert entityType) out of the service's query contract; UserViewQuery only lets the 3 real states exist and rejects pattern-without-db at construction. #697 deferred it to stay consistent with the base table-query methods. Here views are a new path so we took the principled route and left tables alone.

services/housetables/src/main/java/com/linkedin/openhouse/housetables/services/UserTablesServiceImpl.java
ruolin59: Went with stampedEntity instead of updatedEntity: this method is an upsert so "updated" reads a bit ambiguous, and "stamped" matches the existing stamp-the-type language here (the ingress validator + STAMP_TABLE_TYPE). Renamed in 5eb646fc.

services/housetables/src/test/java/com/linkedin/openhouse/housetables/e2e/fixture/UserTableRawSeeder.java
ruolin59: This is only used for testing (the seedLegacyRow split Mike asked for [on #697](https://github.com/linkedin/openhouse/pull/697#discussion_r3876601722)). It plants a legacy NULL entity_type row the app can't create anymore since null writes are rejected, so tests can verify those rows still read back as TABLE.

services/housetables/src/test/java/com/linkedin/openhouse/housetables/e2e/fixture/UserTableStoreCleaner.java
ruolin59: This is only used for testing. It clears the whole store between tests, which the repo no longer exposes now that deleteAll() is sealed.

aastha25 (approved):

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
```diff
@@ -257,6 +268,137 @@ public ResponseEntity<Void> renameTable(
...
+      summary = "Delete a User View",
+      description =
+          "Delete a User House Table view entry identified by databaseID and tableId. Views are "
+              + "always hard deleted; there is no soft-deleted view store.",
+      tags = {"UserTable"})
+  @ApiResponses(
+      value = {
+        @ApiResponse(responseCode = "204", description = "User View DELETE: NO_CONTENT"),
+        @ApiResponse(responseCode = "400", description = "User View DELETE: BAD_REQUEST"),
+        @ApiResponse(responseCode = "404", description = "User View DELETE: TBL_DB_NOT_FOUND")
```
abhisheknath2011, line 389: Do we need to have 500 for corrupt entity here?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableRepositoryTranslationAspect.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Around(
+      "execution(* com.linkedin.openhouse.housetables.repository.impl.jdbc.UserTableHtsJdbcRepository.*(..))")
+  public Object translateCorruption(ProceedingJoinPoint joinPoint) {
```
abhisheknath2011, line 32: Did we verify if the aspect actually works?

services/housetables/src/main/java/com/linkedin/openhouse/housetables/controller/UserHouseTablesController.java
ruolin59: no need, we aren't checking for the specific entitytype here, only doing a query of "delete ... where ... entity_type = VIEW", so corrupt entities are ignored and a 404 will return

services/housetables/src/main/java/com/linkedin/openhouse/housetables/repository/impl/jdbc/UserTableRepositoryTranslationAspect.java
abhisheknath2011: This is critical because the whole corruption→500 contract hinges on the aspect and this can silently fail as well.

ruolin59: Yeah, `RepositoryTranslationAspectTest` covers exactly this. `testAspectRunsOutsidePersistenceExceptionTranslationSoItSeesTheTranslatedFailure` checks it two ways: it walks the real repo proxy chain and asserts the aspect sits in an outer proxy than `PersistenceExceptionTranslationInterceptor`, then plants a real `UNKNOWN` row via raw JDBC and asserts the real read comes back as an unwrapped `CorruptEntityTypeException`. `testCorruptRowStillCarriesTheDiagnosticWithNoCallSiteWrapAnywhere` does the same through the service with no call-site wrap, and `testCorruptDiscriminatorResponseCarriesColumnDiagnostic` in `HtsControllerTest` drives it over real HTTP (GET /hts/entities on a real corrupt row, 500 + diagnostic).

On the silent-fail worry: that structural assertion is the guard. If a Spring Data upgrade moved the translator so the aspect no longer sat outside it, that test fails loudly instead of every corrupt read quietly degrading to a generic 500.

abhisheknath2011 (approved):
Thanks @ruolin59 for addressing the comments!

PR #711, opened 2026-09-02
[RTAS] Use table privileges for replacement

mkuchenbecker:
## Summary

RTAS used a direct `tableCreator` string comparison instead of the standard table authorization path. Route both RTAS phases through `UPDATE_TABLE_METADATA` so the configured authorization handler determines whether the requester is the table administrator.

UPDATE_TABLE_METADATA is only granted to the table admin / creator. 

## Changes

- [ ] Internal API Changes
- [x] Bug Fixes
- [ ] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

Both staged replacement and replacement commit now use `checkTableWritePathPrivileges` with `UPDATE_TABLE_METADATA`. This reuses the same authorization path as other table writes and leaves principal interpretation to the authorization implementation.

The replica check now uses null-safe enum equality so legacy tables without an explicit `tableType` continue through the primary-table authorization path.

Added coverage for staged replacement, replacement commit, authorized and unauthorized metadata updates, and legacy tables without `tableType`.

## Testing Done

Ran the focused RTAS authorization tests and the complete `IcebergSnapshotsServiceTest` class. The tests and `spotlessJavaCheck` completed successfully.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/services/TablesServiceImpl.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/services/TablesServiceImpl.java
@@ -115,8 +115,8 @@ public class TablesServiceImpl implements TablesService {
 
     // Special case handling
     if (tableDto.isPresent() && createUpdateTableRequestBody.isStageReplace()) {
-      // Check if table creator has the privilege to replace the table.
-      authorizationUtils.checkReplaceTablePrivilege(tableDto.get(), tableCreatorUpdater);
+      authorizationUtils.checkTableWritePathPrivileges(
+          tableDto.get(), tableCreatorUpdater, Privileges.UPDATE_TABLE_METADATA);
     } else if (tableDto.isPresent()) {
       if (failOnExist) {
         throw new AlreadyExistsException("Table", String.format("%s.%s", databaseId, tableId));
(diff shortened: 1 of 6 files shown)


mkuchenbecker:
bad impl, use UPDATE_TABLE_METADATA permisison

services/tables/src/main/java/com/linkedin/openhouse/tables/services/TablesServiceImpl.java
```diff
@@ -115,8 +115,8 @@ public Pair<TableDto, Boolean> putTable(
 
     // Special case handling
     if (tableDto.isPresent() && createUpdateTableRequestBody.isStageReplace()) {
-      // Check if table creator has the privilege to replace the table.
-      authorizationUtils.checkReplaceTablePrivilege(tableDto.get(), tableCreatorUpdater);
+      authorizationUtils.checkTableWritePathPrivileges(
```
maluchari, line 118: Can we codify this in our contract stating only table owners can replace their own tables?

maluchari: may be as a part of separate PR - Should we consider how locking would happen along with RTAS? If a table is locked should we allow vs disallow

maluchari (approved):

jiang95-dev (approved):

services/tables/src/main/java/com/linkedin/openhouse/tables/services/TablesServiceImpl.java
mkuchenbecker: We already block when the table is locked.

PR #712, opened 2026-09-03
Optimizer: Add commit-time TableStatsPublisher seam for Table Optimizer stats

abhisheknath2011:
## Summary
First (inert) step of the Table Optimizer "Stats Bridge": add a commit-time seam so per-commit table stats can later be published to the optimizer. `OpenHouseInternalTableOperations.doCommit()` now invokes a pluggable `TableStatsPublisher` right after a successful commit, with the just-committed `TableMetadata` in hand. The default implementation is a no-op, so behavior is unchanged until a concrete publisher is configured.

**New Features / Internal API**
- `TableStatsPublisher` — interface for publishing stats derived from a successful commit. Contract: best-effort, non- blocking, must not throw.
- `NoOpTableStatsPublisher` — default `@Component` implementation; keeps the hook inert in OSS/dev.
- `OpenHouseInternalTableOperations.doCommit()` — after `commitStatus = SUCCESS`, calls `publishOnCommit(tableIdentifier, committedMetadata)` inside a `try/catch(Throwable)` that only logs, so a
  publisher failure can never affect the commit.
- `OpenHouseInternalCatalog.newTableOps()` — injects the publisher bean into the table-ops constructor.

**Refactoring**
- Replaced `@AllArgsConstructor` on `OpenHouseInternalTableOperations` with two explicit constructors. The existing 7-arg constructor is preserved and delegates to `NoOpTableStatsPublisher`, so all existing callers/tests compile unchanged; a new 8-arg constructor accepts the publisher. **Backward compatible.**

**Context**
A concrete publisher (async call to the optimizer stats API, feature-gated) lands in follow-up PRs (stats extraction + LinkedIn internal publisher). This PR intentionally contains only the seam so the commit path is provably unaffected.

**Follow up PR:**
https://github.com/linkedin/openhouse/pull/713 

pull request in 2-3 lines.

## Changes

- [x] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [x] Refactoring
- [ ] Documentation
- [x] Tests

## Testing Done
./gradlew build 

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

### DIFF
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
@@ -68,6 +68,8 @@ public class OpenHouseInternalCatalog extends BaseMetastoreCatalog {
 
   @Autowired TableMetadataCache tableMetadataCache;
 
+  @Autowired TableStatsPublisher tableStatsPublisher;
+
   @Override
   protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
     FileIO fileIO = resolveFileIO(tableIdentifier);
@@ -80,7 +82,8 @@ public class OpenHouseInternalCatalog extends BaseMetastoreCatalog {
         tableIdentifier,
         metricsReporter,
         fileIOManager,
-        tableMetadataCache);
+        tableMetadataCache,
+        tableStatsPublisher);
   }
 
   @Override
--- a/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
+++ b/iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
@@ -42,7 +42,6 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
-import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.iceberg.BaseMetastoreTableOperations;
@@ -70,7 +69,6 @@ import org.apache.iceberg.expressions.Term;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.relocated.com.google.common.base.Objects;
 
-@AllArgsConstructor
 @Slf4j
 public class OpenHouseInternalTableOperations extends BaseMetastoreTableOperations {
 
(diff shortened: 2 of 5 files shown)


cbb330:
I see the event is
- serverside
- built from tablemetadata

Given that, there is these two events have similar behavior: OpenHouseTableAuditEvent/OpenHouseServiceAuditEvent. 

These are also commit time, server side, built from request/table metadata.  Can they be deduplicated/extended for this purpose? 

e.g. OpenHouseTableAuditEvent currently is emitted post-commit, which allows it to utilize some of the context from the request. post-commit as the seam could work for table optimizer as well, and we would prefer deduplicated events if the seam is equal in behavior

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalTableOperations.java
```diff
@@ -421,6 +463,17 @@ updatedMtDataRef, io().newOutputFile(newMetadataLocation)),
         updateMetadataFieldForTable(metadata, newMetadataLocation);
       }
       commitStatus = CommitStatus.SUCCESS;
+      // Best-effort, fire-and-forget publish of commit-time stats to the Table Optimizer. This
+      // must never affect the commit: the publisher contract forbids throwing, but we still guard
+      // against Throwable so a misbehaving implementation cannot fail a successful commit.
+      try {
+        tableStatsPublisher.publishOnCommit(tableIdentifier, updatedMtDataRef);
```
abhisheknath2011, line 470: This is published only on successful commits. This is moved to the finally block on the PR: https://github.com/linkedin/openhouse/pull/713

abhisheknath2011:
> I see the event is
> 
> * serverside
> * built from tablemetadata
> 
> Given that, there is these two events have similar behavior: OpenHouseTableAuditEvent/OpenHouseServiceAuditEvent.
> 
> These are also commit time, server side, built from request/table metadata. Can they be deduplicated/extended for this purpose?
> 
> e.g. OpenHouseTableAuditEvent currently is emitted post-commit, which allows it to utilize some of the context from the request. post-commit as the seam could work for table optimizer as well, and we would prefer deduplicated events if the seam is equal in behavior

Yes as part of optimizer we are capturing commit event stats as async post commit (fire and forget) to store the events in mysql. Internally optimizer API is invoked to store in mysql. Yes we have server side events but we don't have table specific stats needed by Optimizer. Although we can enrich the existing server side event with required data we need a separate consumer pipeline to process these events and store in mysql (current stats and historical state). In future NRT stats could bridge the gap, but we are not there yet. So capturing async post commit event bridges this gap without adding additional load on tables service (as the optimizer stats collection API is part of optimizer service which is deployed separately) and this solution could potentially be used in the long term if needed.

Only successful commits are published, so no duplication. The publish part is moved to the finally block in this PR: https://github.com/linkedin/openhouse/pull/713

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/OpenHouseInternalCatalog.java
```diff
@@ -80,7 +82,8 @@ protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
         tableIdentifier,
         metricsReporter,
         fileIOManager,
-        tableMetadataCache);
+        tableMetadataCache,
+        tableStatsPublisher);
```
mkuchenbecker, line 86: Rather than a stats publisher can this take in a collection of N PostCommit operations where this is the only one? 

The main thought is best-effort postcommit operation is a generic feature, and we plug-in stats or whatever. Postcommit can enforce bounded guarentees such as tight TTL and the observability while the actual buisiness logic remains agnostic.

iceberg/openhouse/internalcatalog/src/main/java/com/linkedin/openhouse/internal/catalog/TableStatsPublisher.java
```diff
@@ -0,0 +1,28 @@
...
+public interface TableStatsPublisher {
+
+  /**
+   * Publish stats derived from a successful commit. Best-effort; implementations must not throw and
+   * must not block the commit thread.
+   *
+   * @param tableIdentifier the committed table
+   * @param committedMetadata the table metadata as committed (includes the current snapshot)
+   */
+  void publishOnCommit(TableIdentifier tableIdentifier, TableMetadata committedMetadata);
```
mkuchenbecker, line 27: OnCommit isn't useful descriptor. 

Should this just be a runnable interface? I am thinking that We might end up with a functional interface or similar Publisher.Publish or publisher.Run.

mkuchenbecker (commented):
I would model it as generic postcommit runs N postcommit operations with bounded guarantees such as max TTL, spawned in the background, etc. Then we know that e.g. a postcommit op at most consumes one thread for 2 seconds, and results / errors consistently consumed. 


A caveat is we need to police the load / threads run and be careful on what we put here and what we guarentee because its by-definition best effort.

PR #726, opened 2026-09-10
Introduce generic lock reasons and action-type declarations

yeilho:
## Summary

Introduce generic lock-reason metadata and an optional request action-type declaration. Existing lock enforcement and authorization remain unchanged.

## Changes

- [x] New Features
- [x] Tests

- Add `LEGACY` and `SYSTEM_ONLY` lock classifications. Omitted or null reasons resolve to `LEGACY` only when `locked=true`; inactive missing/null reasons remain null, and explicit reasons are preserved.
- Add the optional `action-type` catalog property and `X-OpenHouse-Action-Type` header in both supported Java catalog variants. The catalog accepts only `SYSTEM`, case-insensitively, and emits it in uppercase. An absent property omits the header; other supplied values, including `USER`, are rejected.
- Preserve compatibility across persisted policies, API serialization, and generated clients without an unconditional client-side reason default.

## Testing Done

- [x] Added and updated tests for model compatibility, generated-client defaults, real client/server round trips, and both Java catalog variants.
- [x] Local code review completed by GPT-6 Astra and Claude Opus 5.

45 focused generated-SDK/HTTP and Java catalog cases passed on this revision, including rejection of `USER` and mixed-case variants. Scoped formatting checks passed. Full-stack regression coverage is recorded in #740.

# Additional Information

The action declaration does not grant privileges or bypass locks. Lifecycle, enforcement, audit, maintenance-job adoption, and any backfill remain separate follow-up work.

### DIFF
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/CreateUpdateLockRequestBody.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/CreateUpdateLockRequestBody.java
@@ -1,6 +1,7 @@
 package com.linkedin.openhouse.tables.api.spec.v0.request;
 
 import com.google.gson.GsonBuilder;
+import com.linkedin.openhouse.tables.api.spec.v0.request.components.LockReason;
 import io.swagger.v3.oas.annotations.media.Schema;
 import javax.validation.constraints.NotEmpty;
 import javax.validation.constraints.NotNull;
@@ -24,6 +25,12 @@ public class CreateUpdateLockRequestBody {
 
   @Schema(description = "reason for creating/updating the lock on table")
   String message;
+
+  @Schema(description = "Structured lock reason. Omit to preserve legacy lock behavior.")
+  LockReason reason;
+
+  @Schema(description = "Expected table UUID (generation). Required for a reasoned lock.")
+  String expectedTableUUID;
 
   @Schema(
       description = "lock creation epoch time measured in UTC milliseconds for a table",
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockReason.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockReason.java
@@ -1,0 +1,6 @@
+package com.linkedin.openhouse.tables.api.spec.v0.request.components;
+
+/** Structured lock reasons. An omitted reason denotes a legacy lock. */
+public enum LockReason {
+  TIER3_AUTO_CLEANUP
+}
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockState.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockState.java
@@ -22,6 +22,15 @@ public class LockState {
   @Builder.Default
   String message = "Default";
 
+  @Schema(description = "Structured lock reason. Null denotes a legacy lock.")
+  LockReason reason;
+
+  @Schema(description = "Authenticated principal that created the reasoned lock.")
+  String lockOwner;
+
+  @Schema(description = "Table UUID (generation) on which the reasoned lock was created.")
+  String tableUUID;
+
   @Schema(
       description = "lock creation epoch time measured in UTC milliseconds for a table",
       example = "1651002318265")
--- a/services/tables/src/main/java/com/linkedin/openhouse/tables/services/TablesServiceImpl.java
+++ b/services/tables/src/main/java/com/linkedin/openhouse/tables/services/TablesServiceImpl.java
@@ -14,6 +14,7 @@ import com.linkedin.openhouse.internal.catalog.model.SoftDeletedTablePrimaryKey;
 import com.linkedin.openhouse.tables.api.spec.v0.request.CreateUpdateLockRequestBody;
 import com.linkedin.openhouse.tables.api.spec.v0.request.CreateUpdateTableRequestBody;
 import com.linkedin.openhouse.tables.api.spec.v0.request.UpdateAclPoliciesRequestBody;
+import com.linkedin.openhouse.tables.api.spec.v0.request.components.LockReason;
 import com.linkedin.openhouse.tables.api.spec.v0.request.components.LockState;
 import com.linkedin.openhouse.tables.api.spec.v0.request.components.Policies;
 import com.linkedin.openhouse.tables.api.spec.v0.response.components.AclPolicy;
@@ -31,7 +32,9 @@ import com.linkedin.openhouse.tables.utils.TableUUIDGenerator;
 import io.opentelemetry.instrumentation.annotations.WithSpan;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
+import org.apache.commons.lang.StringUtils;
 import org.apache.iceberg.exceptions.BadRequestException;
 import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.exceptions.CommitStateUnknownException;
@@ -112,6 +115,8 @@ public class TablesServiceImpl implements TablesService {
     Optional<TableDto> tableDto =
         openHouseInternalRepository.findById(
             TableDtoPrimaryKey.builder().databaseId(databaseId).tableId(tableId).build());
+
+    LockPolicyValidator.validateUnchanged(tableDto.orElse(null), createUpdateTableRequestBody);
 
     // Special case handling
     if (tableDto.isPresent() && createUpdateTableRequestBody.isStageReplace()) {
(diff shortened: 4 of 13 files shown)


services/tables/src/main/java/com/linkedin/openhouse/tables/services/TablesServiceImpl.java
mkuchenbecker, line 134: Another area to update.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockReason.java
```diff
@@ -0,0 +1,6 @@
+package com.linkedin.openhouse.tables.api.spec.v0.request.components;
+
+/** Structured lock reasons. An omitted reason denotes a legacy lock. */
+public enum LockReason {
+  TIER3_AUTO_CLEANUP
```
mkuchenbecker: This isn't forward compatible. Can you add default or legacy?

mkuchenbecker: correlary is we should do a backfill as a follow up for reason.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/TablesControllerTest.java
```diff
@@ -1629,6 +1629,28 @@ public void createSucceedsForLockPolicyOnTable() throws Exception {
     RequestAndValidateHelper.deleteTableAndValidateResponse(mvc, GET_TABLE_RESPONSE_BODY);
   }
 
+  @Test
```
mkuchenbecker, line 1633: there is no test that exercises client and server, correct?

integrations/java/iceberg-1.2/openhouse-java-runtime/src/iceberg-version-specific/java/com/linkedin/openhouse/javaclient/OpenHouseCatalog.java
```diff
@@ -104,6 +105,9 @@ public class OpenHouseCatalog extends BaseMetastoreCatalog
 
   public static final String CLIENT_VERSION = "client-version";
 
+  /** Opt-in request declaration; it does not grant privileges or bypass locks. */
+  public static final String SYSTEM_ACTION = "system-action";
```
abhisheknath2011: Seems like system-action is passed and initialized as part of Catalog initialization. How this field would be used?

yeilho: A maintenance job configures its dedicated catalog with  system-action=true which will bypass lock restriction to allow reads and writes operations. I will have follow-up PRs that will do this enforcement.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockReason.java
yeilho: Yes to the first comment. Got it for the second one.

services/tables/src/main/java/com/linkedin/openhouse/tables/services/TablesServiceImpl.java
yeilho: My plan was to update this as a part of upcoming enforcement changes.

services/tables/src/test/java/com/linkedin/openhouse/tables/e2e/h2/TablesControllerTest.java
yeilho: That's correct. I will add more tests that covers the trip between client and server with reason and flag without any enforcement.

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
```diff
@@ -27,6 +27,7 @@ public abstract class WebClientFactory {
   private static final String SESSION_ID = "session-id";
 
   public static final String HTTP_HEADER_CLIENT_NAME = "X-Client-Name";
+  public static final String HTTP_HEADER_SYSTEM_ACTION = "X-OpenHouse-System-Action";
```
mkuchenbecker: action type = system action?

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockReason.java
```diff
@@ -0,0 +1,7 @@
+package com.linkedin.openhouse.tables.api.spec.v0.request.components;
+
+/** Structured lock reasons. Omitted or null reasons are interpreted as LEGACY. */
+public enum LockReason {
+  LEGACY,
+  TIER3_AUTO_CLEANUP
```
mkuchenbecker: this is a business use case; can we extend this in Li-openhouse?

>TIER3_AUTO_CLEANUP
Litmus test is descripbe the functionality without mentioning cleanup or tier.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockState.java
```diff
@@ -33,4 +40,8 @@ public class LockState {
       example = "3")
   @Builder.Default
   int expirationInDays = 0;
+
+  public LockReason getReason() {
+    return reason == null ? LockReason.LEGACY : reason;
```
mkuchenbecker: This means tables that are not locked would have a lock reason, no?

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/CreateUpdateLockRequestBody.java
```diff
@@ -36,6 +44,10 @@ public class CreateUpdateLockRequestBody {
       example = "3")
   int expirationInDays = 0;
 
+  public LockReason getReason() {
```
mkuchenbecker, line 47: similar comment here, this will default lock reason as legacy where not specified.

mkuchenbecker (commented):
Main thing is separate "tier 3" from openhouse OSS. Otherwise looks fine. I would likely extend lock reasons or define this as a generaic lock type that we apply under tier 3 conditions.

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
yeilho: I initially intended this as a boolean ( true / false ), but we could use an action-type enum and rename the header to `X-OpenHouse-Action-Type` so the value can be `SYSTEM` or `USER`. Is that what you had in mind?

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockReason.java
yeilho: Wil do.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/components/LockState.java
yeilho: Right now, unlocked tables should have `lockState=null` so it has no reason. On the other hand, locked=false currently defaults to `LEGACY`. I’ll restrict the `LEGACY` default to `locked=true`.

services/tables/src/main/java/com/linkedin/openhouse/tables/api/spec/v0/request/CreateUpdateLockRequestBody.java
yeilho: Ack. Will take care of it.

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
yeilho: I went ahead and implemented to take X-OpenHouse-Action-Type as proposed in my earlier comment.

mkuchenbecker (dismissed):

client/secureclient/src/main/java/com/linkedin/openhouse/client/ssl/WebClientFactory.java
```diff
@@ -27,6 +27,9 @@ public abstract class WebClientFactory {
   private static final String SESSION_ID = "session-id";
 
   public static final String HTTP_HEADER_CLIENT_NAME = "X-Client-Name";
+  public static final String HTTP_HEADER_ACTION_TYPE = "X-OpenHouse-Action-Type";
+  public static final String ACTION_TYPE_SYSTEM = "SYSTEM";
+  public static final String ACTION_TYPE_USER = "USER";
```
maluchari: Why is this actually required? Will ACTION_TYPE_SYSTEM not suffice for now?

maluchari (dismissed):
LGTM. One qn regarding the USER action type which seems to be not required at this point.

yeilho:
> LGTM. One qn regarding the USER action type which seems to be not required at this point.

Yes,  USER  isn’t needed for the current use case. I’ll remove it.

abhisheknath2011 (dismissed):

maluchari (approved):
Thanks for addressing the comments

PR #732, opened 2026-09-16
Add optional time zone to retention policy retention job

mkuchenbecker:
## Stack

Part of a stacked change (listed base first):

1. **#732: retention time zone, backend core** (this PR)
2. #735: retention time zone, Spark 3.5 SQL configuration surface

#733 (zoned retention DML integration tests) is merged into this PR.

---

## Summary

Retention can now measure a table's age in a time zone you choose, instead of always using UTC. A policy that keeps 30 days in `America/Los_Angeles` counts those 30 days in Pacific local time. This matters around midnight: a run measured in UTC can delete a day that is still today where you are.

You never lose data early. A policy that keeps N periods keeps the last N whole periods plus the part of the current period that has already elapsed, so the kept window is between N and N+1 periods. Checked at 12:01, a one-day policy keeps yesterday and all of today so far, about 36 hours; the same policy on an hourly table keeps the last 24 whole hours plus the current partial hour. A policy with no time zone keeps behaving exactly as it does today.

This PR is the OpenHouse backend that computes retention in the chosen zone. The SQL to set the zone is added in #735, and wiring it into the LinkedIn retention app comes after that.

### Problem

Retention measures a table's age only in UTC. A team that thinks in local time, and partitions its data by local date, loses the current local day too early whenever a run happens after UTC midnight but before local midnight.

### Requirements

The time zone is given to us as an input; why a team wants local-time retention is out of scope.

| Tier | Requirement |
|---|---|
| Must | A policy may name a time zone: an IANA id such as `America/Los_Angeles`, or a fixed offset such as `+05:30`. |
| Must | Retention measures the age in that zone, using the zone's offset on the day of the run. |
| Must | Retention keeps deleting whole partitions only; naming a time zone never makes it rewrite individual rows. |
| Must | When the delete has to round to a partition, it rounds toward keeping more data, never less. |
| Must | A policy with no time zone keeps today's delete behavior. |
| Should | The delete is exact where the table's partitioning allows it. |
| Won't | A delete exact to the second for tables partitioned by a real timestamp column. |
| Won't | A different zone per partition, or new granularities. |
| Out of scope | Why a team wants local-time retention. |

### What retention keeps

Retention deletes data older than a cutoff and keeps everything at or after it. The cutoff itself is kept, and the first deleted row is the one just before it.

Retention finds the cutoff in three steps: take the current time in the chosen zone, round it down to the start of the current period (the current hour, day, month, or year, per the policy's granularity), then step back by the policy's count. Rounding down keeps the current partial period, and stepping back keeps the whole periods before it. That is why a one-day policy keeps between one and two days depending on the time of day.

How exact the cutoff can be depends on how the table is partitioned:

| Partition column | The delete |
|---|---|
| A string column holding a formatted date, such as `yyyy-MM-dd` | Is exact: retention compares the column against the same formatted date. |
| A real timestamp column | Rounds the cutoff down to a whole partition, keeping up to one extra partition and never deleting a partition early. |

The offset retention uses is the zone's offset on the day of the cutoff, so an IANA zone follows daylight saving and a fixed offset does not.

### How retention stays a whole-partition delete

Retention deletes by dropping whole partitions, which is fast because it never reads or rewrites row data. A time zone shifts only the single cutoff value into the chosen zone; the stored data is untouched. (Converting every row's timestamp into the zone would force each run to read and rewrite the whole table.)

For a table partitioned by a real timestamp column, the cutoff also has to land on a partition edge, because partitions are stored in UTC. Retention rounds the cutoff down to the nearest whole UTC partition. (Deleting at the exact local instant would fall inside a partition and force that one partition to be rewritten.) Both keep every run to whole-partition drops.

### Worked example

A daily table keeps 30 days in `America/Los_Angeles`, and a run happens at `2024-02-01T02:00Z`, which is still January 31 in Los Angeles. The cutoff is local midnight on January 1, which is `2024-01-01T08:00Z`, rounded down to the start of the UTC day, `2024-01-01T00:00Z`. The `2024-01-01` partition is kept. Measured in UTC the same run would read as February 1 and would delete that partition.

## Changes

- [x] Internal API Changes
- [ ] Bug Fixes
- [x] New Features
- [ ] Code Style
- [ ] Refactoring
- [ ] Documentation
- [x] Tests

New Features: a retention policy may name a time zone, and the retention job measures the table's age in that zone. Internal API Changes: `SparkJobUtil.createDeleteStatement` and `createDeleteFilter`, `Operations.runRetention`, `RetentionSparkApp` (`--timeZone`), `RetentionConfig`, `TablesClient`, and `TableRetentionTask` gain a `timeZone` parameter or field. Tests: added zone-aware unit and integration tests; the existing UTC tests are unchanged.

## Testing Done

Java 17 tests pass. `SparkJobUtilTest` covers the timestamp-column and string-column cases, the Iceberg backup filter, and daylight-saving transitions; `RetentionPolicySpecValidatorTest` covers the new time-zone validation; `AppsTest` and `TableRetentionTaskTest` cover the job wiring; and the embedded-Spark `OperationsTest` cases (merged from #733) run zoned retention end to end and confirm the delete stays a whole-partition drop. A policy with no time zone deletes the same data as before, and the existing UTC tests pass unchanged.

# Additional Information

- [ ] Breaking Changes
- [ ] Deprecations

This feature is delivered in layers: this PR is the OpenHouse backend core, followed by the SQL `WITH TIMEZONE` grammar in the Spark extensions (#735), then the li-openhouse orchestration (the LinkedIn retention app, Central Policy Store mapping, and emitted stats).

🤖 Generated with [GitHub Copilot CLI](https://docs.github.com/copilot/github-copilot-cli)


apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
```diff
@@ -56,33 +59,65 @@ private SparkJobUtil() {}
...
+      // clock) via zonedStringBoundary so this SQL delete and the Iceberg backup filter in
+      // createDeleteFilter derive the same label even across daylight-saving transitions.
+      String condition;
+      if (hasTimeZoneOverride) {
+        ZonedDateTime effectiveNow = now.withZoneSameInstant(ZoneId.of(timeZone));
+        condition =
+            String.format(
+                "%s < '%s'",
+                columnName, zonedStringBoundary(effectiveNow, granularity, count, columnPattern));
+      } else {
```
mkuchenbecker: your have overloaded two behaviors with this. It reads terrible to have else conditions and even if conditions are suspect. 

Audit the code you modified and remove else conditions added if possible. If conditions should be evaluated but not removed. Branchless execution or minimal branching is best so we evaluate at the right layer.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
```diff
@@ -56,33 +59,65 @@ private SparkJobUtil() {}
...
-          "Table: {}. Column pattern: {}, columnName {}, granularity {}s, " + "retention query: {}",
+          "Table: {}. Column pattern: {}, columnName {}, granularity {}s, timeZone {}, retention query: {}",
           fqtn,
           columnPattern,
           columnName,
           granularity,
+          timeZone,
           query);
       return query;
+    } else if (hasTimeZoneOverride) {
```
mkuchenbecker: similar comment here, else is suspect. It seems like we have two query types embeddded here. Its unclear why the query changes and not the value supplied to the query.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
```diff
@@ -101,17 +136,90 @@ public static String createDeleteStatement(
...
+      ZonedDateTime cutoffDate = now.minus(timeUnitGranularity.getDuration().multipliedBy(count));
+      if (!StringUtils.isBlank(columnPattern)) {
+        String formattedCutoffDate = DateTimeFormatter.ofPattern(columnPattern).format(cutoffDate);
+        return Expressions.lessThan(columnName, formattedCutoffDate);
+      } else {
+        long formattedCutoffDate =
+            cutoffDate.truncatedTo(timeUnitGranularity).toEpochSecond() * MICROS_PER_SECOND;
+        return Expressions.lessThan(columnName, formattedCutoffDate);
+      }
+    }
```
mkuchenbecker: Wow, cyclomatic complexity is high here. Else is completely unneeded in the nested if, but in-general flatten this to states vs nested conditions.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
```diff
@@ -101,17 +136,90 @@ public static String createDeleteStatement(
...
-      long formattedCutoffDate =
-          cutoffDate.truncatedTo(timeUnitGranularity).toEpochSecond() * 1000 * 1000; // microsecond
-      return Expressions.lessThan(columnName, formattedCutoffDate);
+      long micros = snappedNativeBoundaryUtcMicros(effectiveNow, granularity, count);
+      return Expressions.lessThan(columnName, micros);
+    }
+  }
+
+  /**
+   * Compute the retention boundary for a string-partitioned column as a formatted wall-clock label.
```
mkuchenbecker: boundary is used liberally in the PRs and its distacting invented term. Its a retention range that is inclusive or exclusive. "boundary" is made-up.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
```diff
@@ -101,17 +136,90 @@ public static String createDeleteStatement(
...
+  }
+
+  /**
+   * Compute the retention boundary for a native timestamp column: the start of the current period
+   * in the given zone, moved back by {@code count} periods, then snapped down to the UTC partition
+   * edge so the resulting delete covers whole partitions (a metadata-only partition drop). The
+   * returned value is the boundary as a UTC wall-clock timestamp; retention deletes rows strictly
+   * before it and keeps rows at or after it.
+   */
+  private static LocalDateTime snappedNativeBoundaryUtc(
```
mkuchenbecker: "snapping" uis an invented concept and you are imperativley doing something I don't understand vs declaratively determining a result.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
```diff
@@ -101,17 +136,90 @@ public static String createDeleteStatement(
...
+    return DateTimeFormatter.ofPattern(columnPattern).format(boundary);
+  }
+
+  /**
+   * Compute the native-timestamp retention boundary as absolute microseconds since the UTC epoch,
+   * so the executed SQL delete and the Iceberg backup filter compare against the identical instant
+   * regardless of the Spark session time zone.
+   */
+  private static long snappedNativeBoundaryUtcMicros(
+      ZonedDateTime nowInZone, String granularity, int count) {
```
mkuchenbecker: remove this function entirely.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
mkuchenbecker: remove this function entirely.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
```diff
@@ -101,17 +136,90 @@ public static String createDeleteStatement(
...
+  private static LocalDateTime snappedNativeBoundaryUtc(
+      ZonedDateTime nowInZone, String granularity, int count) {
+    ChronoUnit unit = convertGranularityToChrono(granularity.toUpperCase());
+    ZonedDateTime periodStart = truncateToGranularity(nowInZone, unit);
+    ZonedDateTime boundaryInZone = periodStart.minus(count, unit);
+    ZonedDateTime boundaryUtc = boundaryInZone.withZoneSameInstant(ZoneOffset.UTC);
+    return truncateToGranularity(boundaryUtc, unit).toLocalDateTime();
+  }
+
+  private static ZonedDateTime truncateToGranularity(ZonedDateTime dateTime, ChronoUnit unit) {
```
mkuchenbecker: remove this function.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
```diff
@@ -101,17 +136,90 @@ public static String createDeleteStatement(
...
+      case HOURS:
+        return dateTime.truncatedTo(ChronoUnit.HOURS);
+      case MONTHS:
+        return dateTime.toLocalDate().withDayOfMonth(1).atStartOfDay(dateTime.getZone());
+      case YEARS:
+        return dateTime.toLocalDate().withDayOfYear(1).atStartOfDay(dateTime.getZone());
+      case DAYS:
+        return dateTime.truncatedTo(ChronoUnit.DAYS);
+      default:
+        throw new IllegalArgumentException(
```
mkuchenbecker: unchecked exception, how was this not caught be review?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
```diff
@@ -101,17 +136,90 @@ public static String createDeleteStatement(
...
+  /**
+   * Compute the retention boundary for a native timestamp column: the start of the current period
+   * in the given zone, moved back by {@code count} periods, then snapped down to the UTC partition
+   * edge so the resulting delete covers whole partitions (a metadata-only partition drop). The
+   * returned value is the boundary as a UTC wall-clock timestamp; retention deletes rows strictly
+   * before it and keeps rows at or after it.
+   */
+  private static LocalDateTime snappedNativeBoundaryUtc(
+      ZonedDateTime nowInZone, String granularity, int count) {
+    ChronoUnit unit = convertGranularityToChrono(granularity.toUpperCase());
```
mkuchenbecker: these parameters make no sense

mkuchenbecker (commented):
Clean this up. Ensure that the feature in INCLUSIVE. I.e. if its 1201, I have one day retention, and I run retention if the column should retain 24 hours of data based on an hourly partitioned table and 36 hours if its a daily retention.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
```diff
@@ -19,100 +22,103 @@
...
-    Case1: "2024-01-01"
-      query:  to_date(substring(datePartition, 0, CHAR_LENGTH('yyyy-MM-dd')), 'yyyy-MM-dd') <
-              date_trunc('DAY', current_timestamp() - INTERVAL 30 DAYs)"
-      result: record will be deleted
-    Case2: "2024-01.01"
-      query:  to_date(substring(datePartition, 0, CHAR_LENGTH('yyyy-MM-dd')), 'yyyy-MM-dd') <
-              date_trunc('DAY', current_timestamp() - INTERVAL 3 DAYs)"
-      result: records will be filtered from deletion
-  */
-  private static final String RETENTION_CONDITION_WITH_PATTERN_TEMPLATE =
```
mkuchenbecker, line 56: why did we remove these templates?

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
```diff
@@ -19,100 +22,103 @@
...
+
+  /**
+   * The oldest label to keep for a string-partitioned column: the retention-zone wall clock moved
+   * back {@code count} periods and formatted with the column's pattern, carrying the zone offset so
+   * patterns that include an offset field format correctly. Rows whose label sorts before this are
+   * deleted; rows at or after it are kept. {@link #createDeleteStatement} and {@link
+   * #createDeleteFilter} both call this, so the executed delete and the Iceberg backup filter agree
+   * even across daylight-saving transitions.
+   */
+  private static String retentionRangeStartLabel(
```
mkuchenbecker: remove any 1-2 line helpers. they do not add clarity they obscure code.

apps/spark/src/main/java/com/linkedin/openhouse/jobs/util/SparkJobUtil.java
```diff
@@ -19,100 +22,103 @@
...
+      ZonedDateTime asOf, String granularity, int count) {
+    ChronoUnit period = convertGranularityToChrono(granularity.toUpperCase());
+    ZonedDateTime rangeStart = startOfPeriod(asOf, period).minus(count, period);
+    ZonedDateTime utcPartitionEdge =
+        startOfPeriod(rangeStart.withZoneSameInstant(ZoneOffset.UTC), period);
+    return utcPartitionEdge.toEpochSecond() * MICROS_PER_SECOND;
+  }
+
+  /** Start of the period containing {@code time}, in {@code time}'s own zone. */
+  private static ZonedDateTime startOfPeriod(ZonedDateTime time, ChronoUnit period) {
```
mkuchenbecker: helpers needing helpers is a smell. Just make a bigger helper.

PR #737, opened 2026-09-17
Guard SYSTEM_ONLY lock lifecycle and expose status

yeilho:
## Summary

Add guarded `SYSTEM_ONLY` lock creation/unlock and metadata-only lock status, building on #726's generic lock contract.

## Changes

Record the acting principal and table generation, keep matching retries idempotent, and require reason/owner/generation guards for unlock. Preserve legacy unlock behavior, explicit recovery for pre-owner locks, active-only `LEGACY` defaults, and lock metadata across ordinary table/snapshot writes.

## Testing Done

- [x] Local code review completed by GPT-6 Astra and Claude Opus 5.
- [x] Added and updated tests for the changes made.

175 targeted lifecycle, policy, controller, DTO, and generated-SDK/HTTP cases passed on this increment.

# Additional Information

Stack 1/4; depends on #726. Targets `main`, so the draft includes its unmerged predecessor. [Review only this increment](https://github.com/yeilho/openhouse/compare/6a2bd3845ecb95bc1dbc6593ac72ac2d980907fa...b03768a2373bd4a8773ee8a6985cf617bb0d541d).

Read/write enforcement and automated deletion are separate.

Generated with [GitHub Copilot CLI](https://docs.github.com/copilot/github-copilot-cli).

PR #738, opened 2026-09-17
Enforce SYSTEM_ONLY locks with action-type declarations

yeilho:
## Summary

Enforce `SYSTEM_ONLY` locks using `X-OpenHouse-Action-Type` without granting additional privileges.

## Changes

Require a `SYSTEM` declaration for otherwise-authorized data reads and writes under an active `SYSTEM_ONLY` lock. An absent declaration returns HTTP 423; other supplied values, including `USER`, unknown, blank, padded, and old boolean declarations, return HTTP 400 when this access check is reached. Preserve existing ACL and legacy-lock checks, protect lock metadata, and surface HTTP 423 with generic authorized-unlock guidance rather than hiding the denial as a missing table.

## Testing Done

- [x] Local code review completed by GPT-6 Astra and Claude Opus 5.
- [x] Added and updated tests for the changes made.

436 service, Java, and Spark cases passed across both runtime variants, covering declaration parsing and authorization order, SDK/HTTP metadata and snapshot writes, catalog load/refresh/tableExists, denial propagation without commit retries, and legacy 400/404 behavior.

# Additional Information

Stack 2/4; depends on #737. Targets `main`, so the draft includes unmerged predecessors. [Review only this increment](https://github.com/yeilho/openhouse/compare/b03768a2373bd4a8773ee8a6985cf617bb0d541d...eb1689a0f4c0a9cb5e6b17c62d72e4cf3567d314).

`SYSTEM` is case-insensitive. Validation remains at the existing post-authorization active-`SYSTEM_ONLY` access check, not global request middleware. Automated deletion, SQL unlock, and deployment are separate.

Generated with [GitHub Copilot CLI](https://docs.github.com/copilot/github-copilot-cli).

### DIFF
--- a/client/tableclient/src/test/java/com/linkedin/openhouse/tables/client/api/TableApiUnlockCompatibilityTest.java
+++ b/client/tableclient/src/test/java/com/linkedin/openhouse/tables/client/api/TableApiUnlockCompatibilityTest.java
@@ -1,0 +1,61 @@
+package com.linkedin.openhouse.tables.client.api;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import com.linkedin.openhouse.tables.client.invoker.ApiClient;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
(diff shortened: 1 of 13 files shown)


yeilho:
Folded into #737, which now carries the SYSTEM_ONLY lock lifecycle, enforcement, audit field and Java client fix in one PR. Closing in favor of #737.

